PR #49, opened 2017-12-04
AutoLayout Helper

3lvis:
Adds AutoLayout sugar.

Now instead of:
```swift
let loginView = LoginView()
loginView.translatesAutoresizingMaskIntoConstraints = false
addSubview(loginView)

NSLayoutConstraint.activate([
    loginView.topAnchor.constraint(equalTo: topAnchor),
    loginView.bottomAnchor.constraint(equalTo: bottomAnchor),
    loginView.leadingAnchor.constraint(equalTo: leadingAnchor),
    loginView.trailingAnchor.constraint(equalTo: trailingAnchor),
])
```

We could do:

```swift
// Changes translatesAutoresizingMaskIntoConstraints to false 
// and constraints top, bottom, leading and trailing to its superview.

let loginView = LoginView()
addSubview(loginView)
loginView.fillInSuperView()
```


Troika/Extensions/UIView+Troika.swift
```diff
@@ -0,0 +1,12 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import UIKit
+
+public extension UIView {
+    public convenience init(withAutoLayout autoLayout: Bool) {
```
axldyb: This is convenient. Keep in mind that several of the components is designed with a custom initializer and won't be able to use this (PreviewGridView, MarketGridView, Button, Ribbon, TextField, ToastView).

Troika/Extensions/NSLayoutConstraint+Troika.swift
```diff
@@ -0,0 +1,51 @@
...
+public extension UIView {
+
+    /// Layouts the current view to fit it's superview.
+    ///
+    /// - Parameters:
+    ///   - insets: The inset for fitting the superview.
+    ///   - isActive: A boolean on whether the constraint is active or not.
+    /// - Returns: The added constraints.
+    @discardableResult
+    func fillInSuperview(insets: EdgeInsets = .zero, isActive: Bool = true) -> [NSLayoutConstraint] {
```
axldyb: I like this naming better that the previous one :)

Troika/Extensions/NSLayoutConstraint+Troika.swift
```diff
@@ -0,0 +1,51 @@
...
+        guard let superview = self.superview else {
+            return [NSLayoutConstraint]()
+        }
+
+        translatesAutoresizingMaskIntoConstraints = false
+
+        var constraints = [NSLayoutConstraint]()
+        constraints.append(topAnchor.constraint(equalTo: superview.topAnchor, constant: insets.top))
+        constraints.append(leadingAnchor.constraint(equalTo: superview.leadingAnchor, constant: insets.leading))
+        constraints.append(bottomAnchor.constraint(equalTo: superview.bottomAnchor, constant: insets.bottom))
```
axldyb: In order to have a view with padding to each side here we will have to pass a negative value for bottom and trailing. `UIEdgeInsets(top: 8, left: 8, bottom: -8, right: -8)` This is not in line with how you expect `UIEdgeInsets` to work.
Documentation says:
> // specify amount to inset (positive) for each of the edges. values can be negative to 'outset'

3lvis:
PR comments addressed. Check Layout+Helpers.swift

Troika/Extensions/Layout.swift
```diff
@@ -0,0 +1,35 @@
...
+        guard let superview = self.superview else {
+            return [NSLayoutConstraint]()
+        }
+
+        translatesAutoresizingMaskIntoConstraints = false
+
+        var constraints = [NSLayoutConstraint]()
+        constraints.append(topAnchor.constraint(equalTo: superview.topAnchor, constant: insets.top))
+        constraints.append(leadingAnchor.constraint(equalTo: superview.leadingAnchor, constant: insets.leading))
+        constraints.append(bottomAnchor.constraint(equalTo: superview.bottomAnchor, constant: insets.bottom))
```
axldyb: Now that `Insets` always has positive values this will cause the setting of the constraint constant here to have the view bleed outside the superview. The trailing and bottom constraint constant must be negative to have an inset from the superview edges.

3lvis: Duh! You’re right. I’ll fix it.

3lvis:
I'll try again in a different branch, some nasty conflicts going on.

PR #51, opened 2017-12-06
Button margins to avoid truncation

ulrikus:
# What?
Changes the `titleEdgeInsets` to a `contentEdgeInsets` so that the size of the button itself changes and not the title content bounds. Adds `margins` to the _button style_ to that for the button styles without borders will be aligned with text etc.

# Why?
Improvement on the button class to make it more usable.

# Screenshots
skjermbilde 2017-12-06 kl 09 11 49

skjermbilde 2017-12-06 kl 09 07 32

### DIFF
--- a/Troika/Components/Button/Button+Style.swift
+++ b/Troika/Components/Button/Button+Style.swift
@@ -91,5 +91,12 @@ public extension Button {
             default: return .sardine
             }
         }
+
+        var margins: UIEdgeInsets {
+            switch self {
+            case .link: return UIEdgeInsets(top: .smallSpacing, left: 0, bottom: .smallSpacing, right: 0)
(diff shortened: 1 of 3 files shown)


Troika/Components/Button/Button+Style.swift
```diff
@@ -91,5 +91,12 @@ public extension Button {
             default: return .sardine
             }
         }
+
+        var margins: UIEdgeInsets {
+            switch self {
+            case .link: return UIEdgeInsets(top: .smallSpacing, left: 0, bottom: .smallSpacing, right: 0)
```
3lvis: Why does the link doesn’t have the same margins?

ulrikus: Because then it will become difficult to align it with text other places because of its intrinsic margins. You would have to offset it with the side margin to align it. The margins are not visible anyways. Those were my thoughts at least.

3lvis:
I think I get it :). Would be nice next time to get a snapshot of both.

PR #54, opened 2017-12-07
Add support for Dynamic Type

3lvis:
artboard

### DIFF
--- a/Troika/Extensions/UIFont+Troika.swift
+++ b/Troika/Extensions/UIFont+Troika.swift
@@ -7,37 +7,62 @@ import UIKit
 public extension UIFont {
 
     public static var title1: UIFont {
-        return UIFont(name: "FINNTypeWebStrippet-Medium", size: 36.0)!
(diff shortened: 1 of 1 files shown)


Troika/Extensions/UIFont+Troika.swift
```diff
@@ -7,37 +7,60 @@ import UIKit
 public extension UIFont {
 
     public static var title1: UIFont {
-        return UIFont(name: "FINNTypeWebStrippet-Medium", size: 36.0)!
```
ulrikus: Why are the font sizes changed?

Troika/Extensions/UIFont+Troika.swift
```diff
@@ -7,37 +7,60 @@ import UIKit
...
+        let font = UIFont(name: "FINNTypeWebStrippet-Medium", size: 28)!
```
ulrikus: Why are the font sizes changed?

3lvis:
Should work, otherwise we should tweak it.

PR #55, opened 2017-12-08
Change button styles

ulrikus:
# What?
Renaming the `.flat` button style to `.callToAction` and adds a different `.flat` style which follows the way it is explained in _finninvers_. All the styles are depicted below with their disabled version.

(the _Left button_ and _Right button_ are still just to show how the button looks when no `widthAnchor` is set. They are not separate styles)

# Why?
I misunderstood the explanation on _finnivers_ at first, but now we have all the desired buttons.

# Screenshot
simulator screen shot - iphone 8 plus - 2017-12-08 at 09 49 47

### DIFF
--- a/Troika/Components/Button/Button+Style.swift
+++ b/Troika/Components/Button/Button+Style.swift
@@ -10,6 +10,7 @@ public extension Button {
         case `default`
         case callToAction
         case destructive
+        case flat
         case link
 
         var font: UIFont {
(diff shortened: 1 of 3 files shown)


ulrikus:
@osanoj, regarding your concerns from Slack conversation.

As far as I can see the Troika changes just touches these areas:
- `SearchResultsActionSectionHeaderView` (label and button)
- `SearchResultsLoggedOutViewController` (label and button)
- `SearchResultsTableViewCell` (label)
- `SearchResultsEmptyViewController` (label)
- `SearchResultsSectionHeaderView` (label)
- `SearchResultsSimpleSectionHeaderView` (label)
- `SearchResultsTableViewCell` (label)

The filter cells are not affected I think. Not quite sure what to look for either, but things stay in place and nothing crashes so I think we're fine. :)

3lvis:
giphy 12

PR #62, opened 2018-01-17
Use generic extension to dequeue a cell in PreviewGridView

richardtop:
Hi,

I've found a generic extension for dequeuing a cell, however, I also found a case when it wasn't used.

As a side note, you could declare extension like this:
```Swift
  func dequeueCell<T: ReusableCell>(for indexPath: IndexPath) -> T {
    return self.dequeueReusableCell(withIdentifier: T.reuseIdentifier(), for: indexPath) as! T
  }
```
And dequeue cell without passing `CellClass` at all, it will be inferred:
```Swift
let cell: MenuCell = tableView.dequeueCell(for: indexPath)
```
or, maybe you have some reasons for not using this style?

Anyway, I like the idea of codifying a design system like this.

### DIFF
--- a/Sources/Components/PreviewGrid/PreviewGridView/PreviewGridView.swift
+++ b/Sources/Components/PreviewGrid/PreviewGridView/PreviewGridView.swift
@@ -119,7 +119,7 @@ extension PreviewGridView: UICollectionViewDataSource {
     }
 
     public func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
-        let cell = collectionView.dequeueReusableCell(withReuseIdentifier: String(describing: PreviewGridCell.self), for: indexPath) as! PreviewGridCell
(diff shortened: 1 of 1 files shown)


3lvis:
Hi @richardtop,

Nice suggestion, I think the reason we chose one option over the other one is most likely just code style. I'll share your version with the team.

richardtop:
@3lvis conceptually both approaches are very similar. The only difference in the case of a generic method, it is impossible to dynamically pick particular cell class at runtime.

3lvis:
@richardtop Hi! I remember why we went for the current approach, the answer is consistency. Check the following comparison.

```swift
// Registering
tableView.register(UITableViewCell.self)

// Dequeueing
let cell = tableView.dequeue(UITableViewCell.self, for: indexPath)
```

## vs

```swift
// Registering
tableView.register(UITableViewCell.self)

// Dequeueing
let cell: UITableViewCell = tableView.dequeueCell(for: indexPath)
```

PR #63, opened 2018-01-25
Small styling changes

ingridvf:
What?
1. text-field bottom border from 2 to 1
2. changed from title4 to body on empty view text block

Why?
1. Makes the fields take less focus, and looks smoother.
2. title4 should not e used for text blocks, we use body styling for that.


ag12:
:clap::skin-tone-3:

3lvis:
Kicking ass, Ingrid!

PR #65, opened 2018-01-30
changed corner radius

ingridvf:
What?
changed corner-radius from 4 and 3 to 8 for buttons and ribbons.

Why?
More rounded corners is more friendly, and matches the guidelines in zeplin.

buttons-after
buttons-before
ribbon-after
ribbon-before


3lvis:
Finally!

PR #66, opened 2018-01-30
List of views for startup

ulrikus:
# What?
Shows a table view of all the different playground views on startup. Currently you can't go back to change view again. You would have to restart the app to do so.

May need/change: 
- _Back_ button (navigationController?)
- It could be a `CollectionViewController` with some cool cells (this was faster to make)

# Why?
So you don't have to choose which playground you want to show before you run the app.


3lvis:
Nice improvement, Ulrik!

ulrikus:
It's a bit hard coded some places, but YOLO!

PR #67, opened 2018-01-30
Added RegisterView

ag12:
**What?**
Added register view. 
simulator screen shot - iphone x - 2018-01-30 at 12 54 27


ulrikus:
Maybe have like the typical confirm password? Then we can check if they are the same before we enable the Register button. Or is it too much?

3lvis:
I would say too much, if the user screws up there's always the "Forgot password" feature.

ag12:
I have not seen that one in for a while, we will have the new design soon.

PR #68, opened 2018-01-31
Added dismiss view controller, and some clean up

ag12:
**What?**
autosearch

*Double tap on the presented view controller to dismiss it. 
*Removed some unnecessary constraints 
*Created Playground.swift (Easier to add new Playground pages for designers) 
*Some cleanup

### DIFF
--- a/Sources/Components/Button/Playground/ButtonPlayground.swift
+++ b/Sources/Components/Button/Playground/ButtonPlayground.swift
@@ -95,12 +95,10 @@ public class ButtonPlayground: UIView {
             destructiveButton.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -.largeSpacing),
 
             flatButton.topAnchor.constraint(equalTo: destructiveButton.bottomAnchor, constant: .mediumLargeSpacing),
-            flatButton.leadingAnchor.constraint(equalTo: leadingAnchor, constant: .largeSpacing),
-            flatButton.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -.largeSpacing),
+            flatButton.centerXAnchor.constraint(equalTo: centerXAnchor),
 
             linkButton.topAnchor.constraint(equalTo: flatButton.bottomAnchor, constant: .mediumLargeSpacing),
-            linkButton.leadingAnchor.constraint(equalTo: leadingAnchor, constant: .largeSpacing),
-            linkButton.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -.largeSpacing),
+            linkButton.centerXAnchor.constraint(equalTo: centerXAnchor),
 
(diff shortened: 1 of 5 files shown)


3lvis:
:rocket:

Sources/Components/Button/Playground/ButtonPlayground.swift
```diff
@@ -95,12 +95,10 @@ public class ButtonPlayground: UIView {
             destructiveButton.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -.largeSpacing),
 
             flatButton.topAnchor.constraint(equalTo: destructiveButton.bottomAnchor, constant: .mediumLargeSpacing),
-            flatButton.leadingAnchor.constraint(equalTo: leadingAnchor, constant: .largeSpacing),
-            flatButton.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -.largeSpacing),
+            flatButton.centerXAnchor.constraint(equalTo: centerXAnchor),
```
ulrikus: This will cause the button to bleed outside the edge of the screen if the button title is too large. I think the `trailingAnchor` and `leadingAnchor` are important constraints.

ulrikus: Event though it will most likely never be that large I think it's good practice.

3lvis: Good point. Might be good to have an example that shows that in case we don’t have any.

ag12: That's a good point, but what I saw was that the button was bigger than what it actually looked like. You could select it even when you pressed outside of the button. 
Should I add those constraints back? @ulrikus

ulrikus: Maybe change the "_equalTo_" to "_lessThanOrEqualTo_" to that it snaps to the size of the button. I think that will do it. You may need to have the centerXAnchor also, I'm not sure.

ag12: Now: screen shot 2018-01-31 at 09 56 07

Before: 
screen shot 2018-01-31 at 09 57 22

ag12:
```
            flatButton.topAnchor.constraint(equalTo: destructiveButton.bottomAnchor, constant: .mediumLargeSpacing),
            flatButton.leadingAnchor.constraint(lessThanOrEqualTo: leadingAnchor, constant: .largeSpacing),
            flatButton.trailingAnchor.constraint(lessThanOrEqualTo: trailingAnchor),
            flatButton.centerXAnchor.constraint(equalTo: centerXAnchor),
```
screen shot 2018-01-31 at 10 00 24

ag12: It makes sense in flat buttons has a background color.

PR #70, opened 2018-02-01
Enable customized login view

ag12:
**What?**
Enabled .enablesReturnKeyAutomatically for TextField. 
Added update method for EmailTextField (When switching between Login and Register view)


ag12:
🙈

PR #73, opened 2018-02-05
Design user entry v2.x.x

ag12:
**What?**
Added privacy policy botton and `func loginView(_ loginView: LoginView, didSelectPrivacyPolicyButton button: Button)` from login view

Removed customer service button and `func registerView(_ registerView: RegisterView, didSelectCustomerServiceButton button: Button)` from register view. 

Added `public var isValid: Bool` on `TextField`
Small refactoring. 

screen shot 2018-02-05 at 16 11 20

### DIFF
--- a/Sources/Components/TextField/TextField.swift
+++ b/Sources/Components/TextField/TextField.swift
@@ -95,6 +95,7 @@ public class TextField: UIView {
     }()
 
     public let inputType: InputType
+
     public var placeholderText: String = "" {
         didSet {
             typeLabel.text = placeholderText
@@ -108,6 +109,19 @@ public class TextField: UIView {
     }
 
     public weak var delegate: TextFieldDelegate?
+
+    public var isValid: Bool {
+        guard let text = textField.text else {
+            return false
+        }
+
+        switch inputType {
+        case .password:
+            return isValidEmail(text)
+        case .email:
+            return isValidEmail(text)
+        }
+    }
 
     // MARK: - Setup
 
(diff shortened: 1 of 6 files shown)


Sources/Components/TextField/TextField.swift
```diff
@@ -206,11 +220,17 @@ public class TextField: UIView {
...
+
     fileprivate func isValidEmail(_ emailAdress: String) -> Bool {
-        let emailRegEx = "[A-Z0-9a-z._%+-]+@[A-Za-z0-9.-]+\\.[A-Za-z]{2,64}"
+        return evaluate("[A-Z0-9a-z._%+-]+@[A-Za-z0-9.-]+\\.[A-Za-z]{2,64}", with: emailAdress)
+    }
 
-        let emailTest = NSPredicate(format: "SELF MATCHES %@", emailRegEx)
-        return emailTest.evaluate(with: emailAdress)
+    fileprivate func isValidPassword(_ password: String) -> Bool {
+        return evaluate("{3,64}", with: password)
```
3lvis: Where did you get this from?

ag12: Yes! I forgot to ask if there’s a specification for password validation in (SPiD). Definitely not 64 char long.

ag12: It seams to be at least 6 characters long.

3lvis: I don’t think assuming any password validation other than not-empty is a good idea.

3lvis: I think assuming any password validation other than non-empty is a bad idea.

ag12: ah.. of course!!! Worst case it can lock existing users out. 
But I think this option could be optional/custom validation in future.

PR #82, opened 2018-02-15
Add function that calculates the size for a BroadcastView given a c…

henrikholmsen:
…onstrained width


axldyb:
Yeah, the market page. The grid view custom layout doesn't support auto layout.

3lvis:
I see :)

Issue #80, opened 2018-02-12
Persist last selected component


3lvis:
Persist last selected component so you don't have to select the same component over and over again when iterating over it.

henrikholmsen:
Solved: https://github.com/finn-no/FinniversKit/pull/86

PR #86, opened 2018-02-21
Persist last selected component

henrikholmsen:
Solving issue #80

### DIFF
--- a/Demo/Demo.swift
+++ b/Demo/Demo.swift
@@ -87,4 +87,18 @@ enum FinniversKitViews: String {
             .consentViewDemoView,
         ]
     }
+
+    static var lastSelectedView: FinniversKitViews? {
+        get {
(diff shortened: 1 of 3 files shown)


Demo/Demo.swift
```diff
@@ -87,4 +87,18 @@ enum FinniversKitViews: String {
             .consentViewDemoView,
         ]
     }
+
+    static var lastSelectedView: FinniversKitViews? {
+        get {
+            guard let lastSelectedViewRawValue = UserDefaults.standard.value(forKey: "lastSelectedView") as? String else {
```
3lvis: Could you abstract "lastSelectedView" into a separate variable so we don't have it hardcoded twice?

henrikholmsen: Absolutely 😊

PR #88, opened 2018-02-22
Add broadcast view container for displaying multiple broadcast views

henrikholmsen:
# Why
When displaying broadcasts we need a way to display multiple broadcasts/BroadcastViews.
The BroadcastContainerView that is added in this PR is a solution for that.

# What
This PR adds:
+ a BroadcastContainerView
+ a Broadcast struct that holds the message to be used in a BroadcastView
+ a BroadcastContainerDemoView that displays how the BroadcastContainerView can be implemented
as a header view in a UITableView

screen shot 2018-02-22 at 16 01 37


ulrikus:
Not that it's too important, but do we really need an additional demo for the broadcast component when it displays basically the same (`BroadcastView` and `BroadcastContainer`). To prevent the list of components in the demo app I think we could do with just the one that shows the two broadcasts. Sort of like the `ButtonDemoView`. Or what do you think?

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
```diff
@@ -0,0 +1,258 @@
...
+    ///   - index: the index for the broadcast
+    /// - Returns: The broadcast to display at the index
+    func broadcastContainerView(_ broadcastContainerView: BroadcastContainerView, broadcastForIndex index: Int) -> Broadcast
+}
+
+/// A view that is used to display multiple BroadcastViews
+public final class BroadcastContainerView: UIView {
+    private struct LayoutConstants {
+        struct ContentView {
+            static let insets = Insets(top: .mediumLargeSpacing, leading: .mediumLargeSpacing, bottom: 0, trailing: .mediumLargeSpacing)
```
3lvis: I usually use structs as models that hold data that is meant to be transporter or encapsulated, not sure if LayoutConstants.ContentView.insets would be a good candidate for this instead of for example having 

```
private let contentViewInsets = Insets(...
```

Same for the BroadcastView.spacing

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
```diff
@@ -0,0 +1,258 @@
...
+    public weak var delegate: BroadcastContainerViewDelegate?
+
+    private lazy var contentView: UIStackView = {
+        let view = UIStackView(frame: .zero)
+
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.axis = .vertical
+        view.spacing = LayoutConstants.BroadcastView.spacing
+        view.distribution = .fillProportionally
+
```
3lvis: We went for not having newlines here might be good to remove them for consistency.

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
```diff
@@ -0,0 +1,258 @@
...
+    public override func didMoveToSuperview() {
+        contentView.frame = CGRect(x: 0, y: 0, width: frame.width, height: 0)
+    }
+}
+
+// MARK: - Public
+
+public extension BroadcastContainerView {
+    /// Reloads the container by removing existing broadcasts and laying out all the broadcasts that the datasource provides
+    func reload() {
```
3lvis: I think this method was supposed to be public or not?

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    public override func didMoveToSuperview() {
+        contentView.frame = CGRect(x: 0, y: 0, width: frame.width, height: 0)
+    }
+}
+
+// MARK: - Public
+
+public extension BroadcastContainerView {
```
3lvis: I like the fact that you are using a public and private extensive to group methods 👌I think if we are not doing that already we should :D. Not a big fan of the `// MARK: -` in general, though, I like them better in Objective-C but in Swift they just feel like comments and I always try to make code that needs comments... We should talk about this with the rest of the team.

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
```diff
@@ -0,0 +1,258 @@
...
+    func remove(_ broadcastView: BroadcastView) {
+        guard let subView = contentView.arrangedSubviews.filter({ $0 == broadcastView }).first else {
+            return
+        }
+
+        contentView.removeArrangedSubview(subView)
+        subView.removeFromSuperview()
+    }
+
+    func reset() {
```
3lvis: First time I read this I thought it was resetting some data, might be good to rename it to `removeSubviews` or something that reflects more what it's doing.

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
```diff
@@ -0,0 +1,258 @@
...
+
+    public override var intrinsicContentSize: CGSize {
+        guard contentView.arrangedSubviews.count != 0 else {
+            return CGSize(width: frame.size.width, height: 0)
+        }
+
+        let broadcastViewsHorizontalSpacings = LayoutConstants.ContentView.insets.leading + abs(LayoutConstants.ContentView.insets.trailing)
+        let constrainedWidth = frame.size.width - broadcastViewsHorizontalSpacings
+
+        let broadcastViewsTotalHeight = contentView.arrangedSubviews.reduce(CGFloat(0), { acc, view in
```
3lvis: What's acc? Could we make this more verbose?

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
```diff
@@ -0,0 +1,258 @@
...
+            contentView.trailingAnchor.constraint(equalTo: trailingAnchor, constant: insets.trailing),
+        ])
+
+        contentView.setContentCompressionResistancePriority(.defaultLow, for: .vertical)
+        contentView.setContentHuggingPriority(.defaultHigh, for: .vertical)
+    }
+
+    func layoutBroadcastViews(from broadcasts: [Broadcast]) {
+        let broadcastViews = broadcasts.map { broadcastView(from: $0) }
+
```
3lvis: newline 🔪

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
```diff
@@ -0,0 +1,258 @@
...
+        broadcastViews.forEach { add($0, to: contentView) }
+    }
+
+    func broadcastView(from broadcast: Broadcast) -> BroadcastView {
+        let broadcastView = BroadcastView(frame: .zero)
+
+        // Set the message on the broadcast view and prepare it in its expanded state
```
3lvis: I don't think this comment is needed. Comments should say "why" not "what". :)

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
```diff
@@ -0,0 +1,258 @@
...
+
+    @objc private func broadcastViewTapped(_ sender: UIGestureRecognizer) {
+        guard let broadcastView = sender.view as? BroadcastView else {
+            return
+        }
+
+        if let delegate = delegate {
+            let newContainerSize: CGSize = {
+                let width = frame.width
+                if contentView.arrangedSubviews.count == 1 { // This is the last broadcast that soon will be removed
```
3lvis: We could use a named variable instead of a comment here.

For example:
```swift
let isBroadcastToBeRemoved = contentView.arrangedSubviews.count == 1
if isBroadcastToBeRemoved {
```

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
```diff
@@ -0,0 +1,258 @@
...
+                    return CGSize(width: width, height: 0)
+                } else {
+                    let height = frame.height - broadcastView.frame.height
+
```
3lvis: newline 🔪

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
```diff
@@ -0,0 +1,258 @@
...
+                    return CGSize(width: width, height: height)
+                }
+            }()
+
+            let broadcastViewMessageIndex = contentView.arrangedSubviews.index(of: broadcastView) ?? 0
+
```
3lvis: Newline 🔪. I guess we been using newlines to separate pieces of information. The `broadcastViewMessageIndex` is used just right after, so having a newline makes it seem like it's not used in the same context as the next piece of code.

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
henrikholmsen: Sure 👍  I just like it this way because of the readability that the name spacing of the nested structs and properties gives 😊 To me, when I read LayoutConstants.ContentView.insets, I know that it is LayoutConsants and that it is for the ContentView and that it's the insets for the ContentView 😊 Of course the LayoutConsants only gives meaning to have because there's layout constants for the Broadcast view spacing as well.

3lvis:
@ulrikus @henrikholmsen I think is good to have one demo per component (view). BroadcastView and BroadcastViewDemo could show rendering and styling while BroadcastContainerView could show stacking, presentation, and dismissal of Broadcasts.

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
henrikholmsen: Change pushed 😊

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
henrikholmsen: Do you mean the new line between the assignments and the return?

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
henrikholmsen: Yes, its supposed to be public. I'll fix that

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
3lvis: Correct :)

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
3lvis: 🔥 sweet!

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
henrikholmsen: changed pushed

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
henrikholmsen: change pushed

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
henrikholmsen: how about something like`clearContainer` or `removeBroadcastViews`?

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
henrikholmsen: accumulator, hehe. I'll make it more verbose 👍

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
3lvis: Maybe `removeAllBroadcastViews` then is also is consistent with the other `remove` method.

Sources/Components/Broadcast/BroadcastContainerView/BroadcastContainerView.swift
henrikholmsen: removed

3lvis:
giphy 11

PR #91, opened 2018-02-23
Make Broadcast equatable

henrikholmsen:
# Why
A Broadcast should be equatable so we can check if a Broadcast is equal or not equal to another broadcast and make it possible to use functions on arrays like contains(element:)

# What
+ Make Broadcast conform to Equatable


henrikholmsen:
That was fast @3lvis! 🏎😄

PR #92, opened 2018-03-02
Toggle switch with description

ulrikus:
# What?
Adds a view called `ToggleSwitchView` that consists of a `UISwitch` and two of the type `Label`; a header label and a description label.
The view model takes in three strings, `headerText`, `onDescriptionText` and `offDescriptionText`. Depending on the value of the switch it displays either the _on_ or the _off_ text. When the switch value changes a delegate method (`toggleSwitch(_ toggleSwitchView: ToggleSwitchView, didChangeValueFor toggleSwitch: UISwitch)`) is called and the text fades to the other string.

# Why?
This will be used in the consent sequence project.

# Screenshot gif
toggleswitch

### DIFF
--- a/Sources/Components/ToggleSwitch/ToggleSwitchView.swift
+++ b/Sources/Components/ToggleSwitch/ToggleSwitchView.swift
@@ -1,3 +1,7 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
 import UIKit
 
 public protocol ToggleSwitchDelegate: NSObjectProtocol {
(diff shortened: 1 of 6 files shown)


Sources/Components/ToggleSwitch/ToggleSwitchView.swift
```diff
@@ -0,0 +1,124 @@
...
+        label.translatesAutoresizingMaskIntoConstraints = false
+        label.numberOfLines = 0
+        return label
+    }()
+
+    private lazy var mySwitch: UISwitch = {
+        let mySwitch = UISwitch()
+        mySwitch.tintColor = .sardine
+        mySwitch.onTintColor = .secondaryBlue
+        mySwitch.onTintColor = .pea
```
axldyb: Is the `onTintColor` set twice here?

ulrikus: Yes it is, by mistake. Fixed now. :)

PR #93, opened 2018-03-05
Fix bug when reloading BroadcastContainerView

henrikholmsen:
Dont return before laying out broadcasts if there are no broadcasts to layou.
This will prevent the delegate beeing notfied about the new container size


3lvis:
💪

PR #94, opened 2018-03-05
Feature/distribute cocoapods

ag12:
**What?**
_Added FinniversKit.podspec_

**Why?**
It's nice to have the library through both Cocoapods and Carthage.


ag12:
Yes.

PR #97, opened 2018-03-06
Fix breaking constraint on showPasswordButton.

henrikholmsen:
⚠️  View debugger still complains about ambiguous _UIBaselineLayoutStrut.
This is a _UIBaselineLayoutStrut for a sublayer of the UITextFields layer and is possibly a UITextField bug
The warning only appears first time the view debugger is started and disappears on the next time.


ag12:
👍

PR #98, opened 2018-03-06
Added compatible topAnchor

ag12:
**What?**
if iOS 11
broadcastView.topAnchor.constraint(equalTo: safeAreaLayoutGuide.topAnchor)
else 
broadcastView.topAnchor.constraint(equalTo: topAnchor)

**Why?**
safeAreaLayoutGuide.topAnchor is only available on iOS 11.


henrikholmsen:
🙌

PR #100, opened 2018-03-06
Scroll login view content up if keyboard intersects input

henrikholmsen:
# Why?
On devices such as the iPhone 5s and 4 the keyboard will partially cover the inputs.

# What
scroll the content in the scrollview up when the keyboard appears

### DIFF
--- a/Sources/Components/Login/LoginView.swift
+++ b/Sources/Components/Login/LoginView.swift
@@ -68,6 +68,12 @@ public class LoginView: UIView {
         return button
     }()
 
+    private lazy var inputsContainer: UIView = {
+        let container = UIView(frame: .zero)
+        container.translatesAutoresizingMaskIntoConstraints = false
+        return container
+    }()
+
     private lazy var newUserStackView: UIStackView = {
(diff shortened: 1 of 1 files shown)


ag12:
We have to add this to RegisterView as well. 
Maybe we could join Register and Login into a base class or something?

henrikholmsen:
@ag12 I'll do the same with the RegisterView 👍 😊
I looked into how we could merge Register and Login, but I'm not sure if it's worth it considering the all the state handling we will have add to the "LoginAndRegisterView". The other option is to abstract the similarities of the two views ( the header text, input fields and the button) into one view, but that will still leave us with a RegistertView and a LoginView so I don't think it's worth the refactoring. It doesn't really hurt right now so I think that's something we could revisit when we start to have pain with two views 😊

ag12:
That sounds like a good plan 👍.

ulrikus:
Congrats with PR #100 btw! 💯🎇

ag12:
#sugarrush 👊

PR #103, opened 2018-03-08
Added UIAppFonts to info.plist

ag12:
**What?**

Added FinniversKit fonts to Info.plist (Fonts provided by application/UIAppFonts).
We don't need to call `UIFont.registerFinniversKitFonts()` anymore, it's called once automatically 

**Why?**

Support for CocoaPods.

### DIFF
--- a/Demo/AppDelegate.swift
+++ b/Demo/AppDelegate.swift
@@ -19,7 +19,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     var window: UIWindow?
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
-        UIFont.registerFinniversKitFonts()
         NotificationCenter.default.addObserver(self, selector: #selector(injected(notification:)), name: Notification.Name.InjectionNotification, object: nil)
 
         window = UIWindow(frame: UIScreen.main.bounds)
--- a/FinniversKit.podspec
+++ b/FinniversKit.podspec
@@ -25,7 +25,7 @@ Pod::Spec.new do |s|
   s.author             = "FINN.no"
   s.social_media_url   = "https://twitter.com/search?q=FINN.no"
 
-  s.platform = :ios, '9.0'
+  s.platform = :ios
   s.swift_version = "4.0"
   s.ios.deployment_target = "9.0"
   s.pod_target_xcconfig = { "SWIFT_VERSION" => "4.0" }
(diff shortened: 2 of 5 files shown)


Sources/Info.plist
```diff
@@ -2,6 +2,13 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>UIAppFonts</key>
+	<array>
+		<string>FINNTypeWebStrippet-Bold.ttf</string>
+		<string>FINNTypeWebStrippet-Light.ttf</string>
+		<string>FINNTypeWebStrippet-Medium.ttf</string>
+		<string>FINNTypeWebStrippet-Regular.ttf</string>
+	</array>
```
axldyb: Do we still need to do this?

Demo/AppDelegate.swift
```diff
@@ -19,7 +19,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     var window: UIWindow?
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
-        UIFont.registerFinniversKitFonts()
```
axldyb: Can we update the readme know that we don't need this line anymore?

Sources/Info.plist
ag12: I doubt, but why not be explicit?
Also more understandable if you’re reading the info file :)

Demo/AppDelegate.swift
ag12: Yes :)

Sources/Info.plist
axldyb: If they are there you expect them to do something, but they are actually not needed. Not sure how that is explicit?

FinniversKit.podspec
```diff
@@ -25,7 +25,7 @@ Pod::Spec.new do |s|
   s.author             = "FINN.no"
   s.social_media_url   = "https://twitter.com/search?q=FINN.no"
 
-  s.platform     = :ios
+  s.platform = :ios, '9.0'
```
ulrikus: Do we need the `, 9.0` part for the platform? Since the deployment_target is listed below.

Sources/Info.plist
ag12: good point.

FinniversKit.podspec
ag12: Removed 👍

PR #107, opened 2018-03-09
Cleaned podspec + added subspec DNA

ag12:


3lvis:
Why a subspec for DNA?

ag12:
Experimentig, maybe even have more subspecs. Right now the though was just to use DNA for font, color and spacing. Avoid adding more files than necessary, keep it clean and simply import what’s really needed.

3lvis:
Not sure if we should invest in extending our CocoaPods support seeing as our plan is to get rid of it.

ag12:
Fair enough :) 
Now that we have to, we might get the best out of it.

ag12:
@3lvis Can we merge this? This version of .podspec is cleaner anyways.

ag12:
Thanks 🗡💯🗡

PR #110, opened 2018-03-11
Use normal top anchor

3lvis:
I think using the normal top anchor instead of the `safeAreaLayoutGuide.topAnchor`. At least it works fine in the demo.


ag12:
should we use view.compatibleTopAnchor ?

3lvis:
I don’t think that’s needed

ulrikus:
I like the idea, but think we should be consistent and use it everywhere if we are to use it.

3lvis:
@ulrikus Not sure if I understand your comment, this PR changes to use topAnchor, which is the same thing we are doing everywhere.

ulrikus:
I was thinking that the `compatibleTopAnchor` was a nice idea to get to utilize the `safeAreaLayoutGuide` for iOS11, but it's fine not to use it also seeing as we are some time away from supporting only iOS11.

axldyb:
I agree to the last comment. Safe area is useful for UIViewController, but not between parent and child views.

PR #112, opened 2018-03-11
Add Foundation folder for Foundation extensions

3lvis:
screen shot 2018-03-11 at 9 17 07 am


ag12:
Since you're on it, should we also move these folders one level up? 
structure

3lvis:
I’ll make a different PR for that one

PR #114, opened 2018-03-13
Textfield design changes

axldyb:
# What?
Changes TextField floating label color and underline height

# Why?
...as per requested by the designers. [Trello card](https://trello.com/c/wukkVtNC)

### DIFF
--- a/Sources/Components/TextField/TextField.swift
+++ b/Sources/Components/TextField/TextField.swift
@@ -47,7 +47,7 @@ public class TextField: UIView {
     private let animationDuration: Double = 0.3
 
     private lazy var typeLabel: Label = {
-        let label = Label(style: .detail(.stone))
+        let label = Label(style: .detail(.licorice))
         label.translatesAutoresizingMaskIntoConstraints = false
         label.alpha = 0
         return label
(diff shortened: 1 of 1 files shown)


Sources/Components/TextField/TextField.swift
```diff
@@ -162,15 +167,16 @@ public class TextField: UIView {
             textField.topAnchor.constraint(equalTo: typeLabel.bottomAnchor, constant: .mediumSpacing),
             textField.leadingAnchor.constraint(equalTo: leadingAnchor),
             textField.trailingAnchor.constraint(equalTo: trailingAnchor),
-            textField.trailingAnchor.constraint(equalTo: trailingAnchor),
```
ulrikus: Uups.. :)

axldyb: Hehe, yeah noticed that it was set twice. Changed the last one to be bottom anchor. As the underline height now changes I didn't want that one to be the one attached to the bottom. That would cause the whole view to resize when activating the textfield.

ulrikus: Yeah, that was probably a brain fart from my part. Hence, the "Uups.." ;)

3lvis: giphy 12

PR #115, opened 2018-03-14
Refactor colors defined in Button.Style to UIColor extension

henrikholmsen:
# Why
All colors used in FinniversKit should be exposed in UIColor extensions equally as DNA colors


ag12:
👍

Sources/DNA/Color/Color+Button.swift
```diff
@@ -0,0 +1,27 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import Foundation
+
+public extension UIColor {
```
ag12: I think I made a mistake when I made this extension public. I don't think its actually necessary since its an extension. What do you think?
https://stackoverflow.com/questions/33925840/access-control-for-swift-extensions

ulrikus:
Maybe check with Ingrid that the colors are correct. I vaguely remember that there is a possibility that the colors may be off by a slight nuance, but maybe not. Otherwise, good idea. I was originally supposed to do something like this myself. Must have forgot.

Sources/DNA/Color/Color+Button.swift
```diff
@@ -0,0 +1,27 @@
...
+    public class var callToActionButtonHighlightedBodyColor: UIColor {
```
ag12: also, why not use static instead of class? :)

ulrikus:
You could also possibly use the convenience initializer in the _Color.swift_ file also.

Sources/DNA/Color/Color+Button.swift
henrikholmsen: I think so too, the extension should inherit the access level of UIColor so making it `public` is redundant.

Sources/DNA/Color/Color+Button.swift
henrikholmsen: I would probably use `public static let`, but the UIColor extensions in `Color.swift` used `class var` so I did it the same way for consistency.

3lvis: I think because this is how it's defined in UIKit. Might play better with ObjC but I'm not sure.

Sources/DNA/Color/Color+Button.swift
ag12: Nice!

Sources/DNA/Color/Color+Button.swift
henrikholmsen: ObjC interop might be the reason unless they actually want them to be overridable. Cause I guess thats the only other difference between class methods/properties and static methods/properties? 
At least I think I've read that runtime look up for class and static methods/properties are the same.

Sources/DNA/Color/Color+Button.swift
```diff
@@ -0,0 +1,27 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import Foundation
+
+extension UIColor {
+    public class var callToActionButtonHighlightedBodyColor: UIColor {
```
3lvis: Not sure if we need a new class for this, would be nice to have all the colors in one single file

Sources/DNA/Color/Color.swift
```diff
@@ -61,7 +61,7 @@ public extension UIColor {
         return UIColor(r: 104, g: 226, b: 184)!
     }
 
-    private convenience init?(r: CGFloat, g: CGFloat, b: CGFloat, a: CGFloat = 1.0) {
+    internal convenience init?(r: CGFloat, g: CGFloat, b: CGFloat, a: CGFloat = 1.0) {
```
3lvis: Internal is the default, so should be fine to remove it.

Sources/DNA/Color/Color+Button.swift
axldyb: I agree with @3lvis. Maybe keep it as a separate extension, but in the same file as the others. Also, I don't like the confusion in naming this file referencing `Color` which currently isn't a type. The existing definitions of colors is done in an extension on `UIColor`.

Sources/DNA/Color/Color.swift
henrikholmsen: Is it really when the extension has explicit public access control? 🤔
Since Amir said earlier that he didn't intend that the extension should be made explicitly public I'll remove that and and the internal for the initialiser 😄

henrikholmsen:
Bump 😄

PR #117, opened 2018-03-15
Add CGColor extension for DNA colors

henrikholmsen:
# Why?
When assigning DNA colors from our `UIColor` extension to `UIColor` properties we preferably should use the shorthand dot syntax. 
E.g: `view.backgroundColor = .primaryBlue` instead of `view.backgroundColor = UIColor.primaryBlue`

For properties that expect a`CGColor` we now would have to write this to provide the compiler with the context of `.primaryBlue`: `layer.borderColor = UIColor.primaryBlue.cgColor`

So to keep the nice shorthand dot syntax for`CGColor` properties we add a `CGColor` extension for all the UIColors so we can write this:
`layer.borderColor = .primaryBlue`

### DIFF
--- a/Sources/DNA/Color/Color.swift
+++ b/Sources/DNA/Color/Color.swift
@@ -66,6 +66,64 @@ extension UIColor {
     }
 }
 
+extension CGColor {
+    public class var ice: CGColor {
+        return UIColor.ice.cgColor
+    }
+
+    public class var milk: CGColor {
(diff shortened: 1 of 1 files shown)


ag12:
☝️👏🏼

henrikholmsen:
@ulrikus. The compiler would need the UIColor prefix to understand the context.

PR #118, opened 2018-03-19
Update broadcast view design and fix bug in broadcast container view

henrikholmsen:
# What?
+ Updates design on Broadcast according to Zeplin
+ Fixes bug in BroadcastContainerView where the calculation of the containers new height after dismissing a BroadcastView didn't mind the BroadcastView spacing to the next BroadcastView.


3lvis:
Might be good to get a screenshot when doing UI changes :)

henrikholmsen:
Absolutely @3lvis. I'll add a before/after screenshot for later reference 😊

PR #124, opened 2018-04-06
Rename BroadcastView to Broadcast

3lvis:
Why?
To keep consistency with other components, same as we don't do TextFieldView or LabelView, broadcast is a UI element so we can drop the View suffix.

For fullscreen views though, I think it makes sense to keep the View, for example, RegistrationView.


ulrikus:
Note @henrikholmsen. This will break the FINN app if things don't get renamed there as well (62 errors). Just sayin'. I'll take care of it though.

PR #125, opened 2018-04-10
Inline consent view

ulrikus:
# What?
A smaller consent view. Inline means that it is not a "popup", but a view that is present until you interact with it. Sort of in the same sense that _"Discover Cat"_ is an inline rating view.

# Why?
Extracts the view from the FINN app to be consistent with the practice of using FinniversKit.

# Screenshot
simulator screen shot - iphone x - 2018-04-10 at 10 40 21

### DIFF
--- a/Sources/Components/New Group/InlineConsentView.swift
+++ b/Sources/Components/New Group/InlineConsentView.swift
@@ -1,0 +1,126 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import UIKit
+
+public protocol InlineConsentViewDelegate: NSObjectProtocol {
+    func inlineConsentView(_ consentActivateDiscoverView: InlineConsentView, didSelectActivateButton button: Button)
+    func inlineConsentView(_ consentActivateDiscoverView: InlineConsentView, didSelectInfoButton button: Button)
+}
+
+public class InlineConsentView: UIView {
+
+    // MARK: - Internal properties
+
+    private lazy var descriptionTitleLabel: Label = {
+        let label = Label(style: .body(.licorice))
+        label.translatesAutoresizingMaskIntoConstraints = false
+        label.numberOfLines = 0
+        return label
+    }()
+
+    private lazy var yesButton: Button = {
+        let button = Button(style: .callToAction)
+        button.translatesAutoresizingMaskIntoConstraints = false
+        button.addTarget(self, action: #selector(yesButtonTapped), for: .touchUpInside)
+        return button
+    }()
+
+    private lazy var infoButton: Button = {
+        let button = Button(style: .flat)
+        button.translatesAutoresizingMaskIntoConstraints = false
+        button.addTarget(self, action: #selector(infoButtonTapped), for: .touchUpInside)
+        return button
+    }()
+
+    // MARK: - External properties / Dependency injection
+
+    public weak var delegate: InlineConsentViewDelegate?
+
+    public var descriptionText: String = "" {
+        didSet {
+            descriptionTitleLabel.text = descriptionText
+        }
+    }
+
+    public var yesButtonTitle: String = "" {
+        didSet {
+            yesButton.setTitle(yesButtonTitle, for: .normal)
+        }
+    }
+
+    public var infoButtonTitle: String = "" {
+        didSet {
+            infoButton.setTitle(infoButtonTitle, for: .normal)
+        }
+    }
+
+    // MARK: - Setup
+
+    public override init(frame: CGRect) {
+        super.init(frame: frame)
+        setup()
+    }
+
+    public required init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+        setup()
+    }
+
+    private func setup() {
+        addSubview(descriptionTitleLabel)
+        addSubview(yesButton)
+        addSubview(infoButton)
+
+        NSLayoutConstraint.activate([
+            descriptionTitleLabel.topAnchor.constraint(equalTo: topAnchor),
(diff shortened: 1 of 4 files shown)


Sources/Components/New Group/InlineConsentView.swift
```diff
@@ -0,0 +1,126 @@
...
+
+    // MARK: - Superclass overrides
+
+    public override var intrinsicContentSize: CGSize {
+        let intermediateSpacing = CGFloat.mediumLargeSpacing * 2
+
+        guard let descriptionText = descriptionTitleLabel.text else {
+            return CGSize.zero
+        }
+        let screenSize = UIScreen.main.bounds
```
henrikholmsen: I'm a little uncomfortable with using the `UIScreen.main` singleton to get the screen size here.🤔 Especially since this is a view created in a framework. Would it be better to ask the`delegate` to provide the screen size?

Sources/Components/New Group/InlineConsentView.swift
```diff
@@ -0,0 +1,126 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import UIKit
+
+public protocol InlineConsentViewDelegate: NSObjectProtocol {
+    func inlineConsentView(_ consentActivateDiscoverView: InlineConsentView, didSelectActivateButton button: Button)
+    func inlineConsentView(_ consentActivateDiscoverView: InlineConsentView, didSelectInfoButton button: Button)
```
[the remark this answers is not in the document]
ulrikus: A bit hasty by my part. I copied the file from the PR in the FINN-app. WillFix! :)

Sources/Components/New Group/InlineConsentView.swift
3lvis: I'm still having a hard time understanding why this is needed, feels like to goes against what Auto Layout stands for, we even have a few hardcoded dimensions on this method.

Sources/Components/New Group/InlineConsentView.swift
```diff
@@ -0,0 +1,126 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import UIKit
+
+public protocol InlineConsentViewDelegate: NSObjectProtocol {
```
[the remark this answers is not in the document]
ulrikus: Ok. I'm not really aware of the differences, but I was just following what seems to be the convention for the delegate protocols in FinniversKit. But I can of course change it to `public protocol InlineConsentViewDelegate: class`. If that's what you meant?

Sources/Components/New Group/InlineConsentView.swift
ulrikus: I don't like it either @3lvis , but there is a method called `calculatedGridHeaderViewFrame` in `MarketViewController` that needs it to calculate the height of the collectionView header based on whether or not the `inlineConsentView` and/or `broadcastContainerView` is included. Could probably be solved in a different way, but I did this to not have to rewrite to much and to follow the pattern the `broadcastContainerView` used when it was implemented.

Sources/Components/New Group/InlineConsentView.swift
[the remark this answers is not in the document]
henrikholmsen: Correct :) Using `class` limits the adopters of the protocol to class types and does the same thing you are trying to achieve by using `NSObjectProtocol`, making it possible to have a `weak` reference on your delegate. The difference is that `NSObjectProtocol` also adds a lot of other requirements to the adaptor.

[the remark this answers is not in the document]
henrikholmsen: Actually, having a look at the Swift 4.1 documentation it seems the textbook way of limiting protocols to classes has changed from using `class` to `AnyObject`. So I guess you should use that instead 😄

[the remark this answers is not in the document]
henrikholmsen: You mean `AnyObject`, right?😉 I guess it would be nice to be consistent, so updating other delegate protocols would be nice 🙂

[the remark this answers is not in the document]
ulrikus: Yes, I mean `AnyObject`. haha. 😬😅 That can be done another time I guess. I'll change it for this file. Thanks! 😃 👍

Sources/Components/New Group/InlineConsentView.swift
3lvis: I see, nevermind then :)

henrikholmsen: Would it be an option to not do the ` max(screenSize.width, widestButtonWidth)` and instead return the widest button width? I think this would be more "true" to the concept of intrinsic content size and instead make the client/implementation of the view handle how it should be laid out in relation to the max width. Does that make sense? 🤔😄

ulrikus: Actually, I think it would be more true to return the width of the `descriptionTitleLabel` since that would in most cases be the widest of the items in the view and, in fact, be what defines the width of the view. Or am I wrong here?

henrikholmsen: Sounds good :) Or perhaps the widest one of them? So something like `max(descriptionTitleLabelWidth, widestButtonWidth)`

3lvis: _Mirror, mirror, on the wall, who's the widest of them all?_

henrikholmsen: 😂

ulrikus:
All comments addressed @henrikholmsen and @3lvis. Care to check again? 😁

PR #130, opened 2018-04-24
EmptyView ❤️ Auto Layout

axldyb:
# What?
Fixes issues with initial position for the empty view shapes when the empty view is using Auto Layout

### DIFF
--- a/Sources/Fullscreen/EmptyView/EmptyView.swift
+++ b/Sources/Fullscreen/EmptyView/EmptyView.swift
@@ -9,39 +9,36 @@ public class EmptyView: UIView {
 
     // MARK: - Internal properties
 
-    private let screenSize: CGSize = UIScreen.main.bounds.size
     private let sizeOfTriangle = CGSize(width: 90, height: 90)
     private let sizeOfCircle = CGSize(width: 75, height: 75)
     private let sizeOfRoundedSquare = CGSize(width: 55, height: 55)
     private let sizeOfSquare = CGSize(width: 100, height: 100)
 
+    private var hasLayedOut = false
+
     private lazy var triangle: TriangleView = {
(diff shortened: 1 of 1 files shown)


Sources/Fullscreen/EmptyView/EmptyView.swift
```diff
@@ -161,6 +153,31 @@ public class EmptyView: UIView {
...
+        super.layoutSubviews()
+
+        // We only want to lay out once
+        guard hasLayedOut == false else {
+            return
+        }
+
+        let slice = frame.width / 8
+
+        // We resposition the shapes after the EmptyView itself has layed out its frame.
```
3lvis: Small typo: "reposition"

Sources/Fullscreen/EmptyView/EmptyView.swift
```diff
@@ -161,6 +153,31 @@ public class EmptyView: UIView {
...
+        // At this point we will have its size even if we use constraints to lay it out.
+        triangle.center = CGPoint(x: slice, y: frame.height - (sizeOfTriangle.height / 2))
+        circle.center = CGPoint(x: slice * 3, y: frame.height - (sizeOfCircle.height / 2))
+        roundedSquare.center = CGPoint(x: slice * 5, y: frame.height - (sizeOfRoundedSquare.height / 2))
+        rectangle.center = CGPoint(x: slice * 7, y: frame.height - (sizeOfSquare.height / 2))
```
ulrikus: The `slice` system is nice, but there is a possibility that an object will start outside the width of the screen if the object close to the edge is big enough. Just sayin'.. I don't think it will be a problem since we actually decide how large the objects are, but it's worth a mention.

PR #133, opened 2018-04-25
Adds action button to empty view

axldyb:
# What?
Adds action button to empty view

# Why?
In case the empty view is used as a placeholder for content when logged out we need an option to send the user to the login flow. 

# How?
The button is optional. If you set the button title it will show. Otherwise it will remain hidden.

# Screens
simulator screen shot - iphone 5s - 2018-04-25 at 10 42 36

### DIFF
--- a/Sources/Fullscreen/EmptyView/EmptyView.swift
+++ b/Sources/Fullscreen/EmptyView/EmptyView.swift
@@ -5,6 +5,10 @@
 import CoreMotion
 import UIKit
 
+public protocol EmptyViewDelegate: class {
+    func performAction(from emptyView: EmptyView)
+}
+
 public class EmptyView: UIView {
 
     // MARK: - Internal properties
@@ -62,6 +66,14 @@ public class EmptyView: UIView {
         return label
     }()
 
+    private lazy var actionButton: Button = {
+        let button = Button(style: .callToAction)
+        button.translatesAutoresizingMaskIntoConstraints = false
+        button.addTarget(self, action: #selector(performAction), for: .touchUpInside)
+        button.isHidden = true // Default is hidden. When a title is set it well be displayed.
+        return button
+    }()
+
     private lazy var animator: UIDynamicAnimator = {
         let animator = UIDynamicAnimator(referenceView: self)
         return animator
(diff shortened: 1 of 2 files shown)


Sources/Fullscreen/EmptyView/EmptyView.swift
```diff
@@ -5,6 +5,10 @@
 import CoreMotion
 import UIKit
 
+public protocol EmptyViewDelegate: class {
+    func performAction(from emptyView: EmptyView)
```
3lvis: I think this naming would be more consistent with the other delegates.

```swift
func emptyView(_ emptyView: EmptyView, didSelectActionButton button: Button)
```

Sources/Fullscreen/EmptyView/EmptyView.swift
```diff
@@ -62,6 +66,14 @@ public class EmptyView: UIView {
         return label
     }()
 
+    private lazy var actionButton: Button = {
+        let button = Button(style: .callToAction)
+        button.translatesAutoresizingMaskIntoConstraints = false
+        button.addTarget(self, action: #selector(performAction), for: .touchUpInside)
+        button.isHidden = true // Default is hidden. When a title is set it well be displayed.
```
3lvis: Small typo: `it will be displayed.`

Sources/Fullscreen/EmptyView/EmptyView.swift
axldyb: Ah, yes. Consistency is King 🤴🏻

3lvis:
>The button is optional. If you set the button title it will show. Otherwise it will remain hidden.

3lvis:
Looks pretty good just a few minor comments. Nice job!

ag12:
That was smart 😍

3lvis:
giphy 2

Sources/Fullscreen/EmptyView/EmptyView.swift
axldyb: Damn. There must always be one.

ag12:
💯 🎖

ingridvf:
Good job:) Be aware  that a new design for this will arrive in a couple of weeks or so.

axldyb:
Yes, sure. Will use this for now and then update when the new design is ready!

PR #135, opened 2018-04-25
Update/empty view

ag12:
**_Before:_**
before

**_After:_**
after

### DIFF
--- a/Sources/Fullscreen/EmptyView/EmptyView.swift
+++ b/Sources/Fullscreen/EmptyView/EmptyView.swift
@@ -16,7 +16,8 @@ public class EmptyView: UIView {
     private let sizeOfTriangle = CGSize(width: 90, height: 90)
     private let sizeOfCircle = CGSize(width: 75, height: 75)
     private let sizeOfRoundedSquare = CGSize(width: 55, height: 55)
-    private let sizeOfSquare = CGSize(width: 100, height: 100)
+    private let sizeOfRectangle = CGSize(width: 100, height: 100)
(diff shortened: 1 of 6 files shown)


ag12:
Yes :)

ag12:
Agile development 😆

Sources/Fullscreen/EmptyView/EmptyView.swift
```diff
@@ -16,7 +16,8 @@ public class EmptyView: UIView {
     private let sizeOfTriangle = CGSize(width: 90, height: 90)
     private let sizeOfCircle = CGSize(width: 75, height: 75)
     private let sizeOfRoundedSquare = CGSize(width: 55, height: 55)
-    private let sizeOfSquare = CGSize(width: 100, height: 100)
+    private let sizeOfRectangle = CGSize(width: 100, height: 100)
```
ulrikus: It's not really a rectangle when it's squared though..

axldyb:
Not sure I get the reasoning behind adding one more element? It looks prette clamped now, especially on the smaller screens. The distribution of the elements also feels unbalanced. Will the design really change here? We will still use this empty screen for the MinFINN screen even if the others will have a different design.

3lvis:
I had the impression that it felt a little bit empty before. I like the new colors better but they might be attracting too much attention from the main element.

axldyb:
Also, it now starts of with some movement as the circle will start a bit inside the triangle on screens with a width of 375 and smaller. I liked how it was more mysterious earlier when the user had to discover the movable elements by touching them or tilting the phone.

PR #137, opened 2018-05-01
ListViews

3lvis:
- Created the ListViews section which will hold cells, table views and collection views
- This change was partly because table view and collection views aren't necessarily part of fullscreen because a collection view might be just part of a view with other elements such as the Frontpage
- Inside each ListView type I've separated classes into: Demo, Cell, ListView and Model
- This inital PR just moves the existing ListViews (Market and GridPreview) into a separate section
- The plan is to afterwards extend ListViews to contain more types such as Favorites, YourAds, Notifications and so on

folder structure

demo app

demo views


3lvis:
💓

axldyb:
Why all the renaming of classes? Some, `PreviewGridView` to `GridPreviewListView` for example, reads a bit strange. A preview grid sounds like a grid displaying previews. A grid preview sounds like a preview of a grid.

axldyb:
Merged already? Ok, please note that this will cause a bunch of necessary updates to the FINN project as soon as anyone runs `carthage update`...

3lvis:
Could be CollectionPreviewListView and TablePreviewListView as well.

axldyb:
The repetition of the work _view_ messes with the readability, yes, but the representation of what the classes actually does is weakened. I regret introducing the "preview" naming to begin with 😅

3lvis:
Other alternatives:
- AdCollectionListView and AdTableListView
- CollectionAdListView and TableAdListView

3lvis:
- MarketListView works fine I guess the main issue is with the preview one that has multiple ways to be represented.

axldyb:
Well, I like the grid and table terms describing how the list views will be displayed. What are the TablePreviewListView going to be used for? Search result lists?

3lvis:
That's correct :)

3lvis:
- TablePreviewListView and GridPreviewListView
- TableAdsListView and GridAdsListView

I think the AdsListView might work better than the PreviewListView.

axldyb:
Yeah, the idea behind the "preview" naming was to be as generic as possible. I guess that has failed a bit. May be better to use "Ad" instead.

3lvis:
I can try with Ad, I think it would make it easier to understand.

ingridvf:
I have differentiated between GridView (https://zpl.io/bzym1NG), ListView (zpl.io/VkD31YG) and CardView, I'm guessing we will have them as well?

3lvis:
@ingridvf I wasn't able to find an example of CardView in Zeplin.

3lvis:
@ingridvf Those GridView, ListView and CardView have one name that represents them all?

3lvis:
- Recycling
  - Grid
    - MarketsGridView
    - AdsGridView
  - List
    - FavoritesListView
    - NotificationsListView
    - AdsListView

ingridvf:
@3lvis AdView maybe?

PR #141, opened 2018-05-03
Update FinniversKitView template

henrikholmsen:
# Why
This is a proposition for changes to the `FinniversKitView` template.

# What

## Make template class `final`
+ Classes should default to be final.
+ If the class needs to be a subclassable at a later stage we can remove the final keyword once we have made sure it will function properly as a superclass.
+ There are performance  gains by making classes `final` by reducing dynamic dispatch

## Make template class `public`
+ By default all FinniversKit views are intended to public. This saves us from returning to the class and making it public once its not available for us in outside the FinniversKit module😄

## Add `private extension` and `public extension`
+ Makes for a better overview of the classes `private` and `public` functions

# Discussion
What do we feel about the `// MARK: -`? Any changes we would like to make?

# Screenshots
screen shot 2018-05-03 at 11 51 08

### DIFF
--- a/Templates/FinniversKitView.xctemplate/___FILEBASENAME___.swift
+++ b/Templates/FinniversKitView.xctemplate/___FILEBASENAME___.swift
@@ -1,6 +1,6 @@
 import UIKit
 
-class ___FILEBASENAMEASIDENTIFIER___: <#View Class#> {
+final public class ___FILEBASENAMEASIDENTIFIER___: <#View Class#> {
 
     // MARK: - Internal properties
 
--- a/Templates/FinniversKitView.xctemplate/___FILEBASENAME___.swift
+++ b/Templates/FinniversKitView.xctemplate/___FILEBASENAME___.swift
@@ -18,14 +18,21 @@ final public class ___FILEBASENAMEASIDENTIFIER___: <#View Class#> {
         setup()
     }
 
-    private func setup() {
+    // MARK: - Superclass Overrides
+    
+}
+
+// MARK: - Private
+
+private extension ___FILEBASENAMEASIDENTIFIER___ {
+    func setup() {
         // Perform setup
         // Add child views as subviews
         // Setup constraints/frames
     }
+}
 
-    // MARK: - Superclass Overrides
+// MARK: - Public
 
-    // MARK: - Private
-
+public extension ___FILEBASENAMEASIDENTIFIER___ {
 }
--- a/Templates/FinniversKitView.xctemplate/___FILEBASENAME___.swift
+++ b/Templates/FinniversKitView.xctemplate/___FILEBASENAME___.swift
@@ -22,6 +22,11 @@ final public class ___FILEBASENAMEASIDENTIFIER___: <#View Class#> {
     
 }
 
+// MARK: - Public
+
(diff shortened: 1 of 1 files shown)


Templates/FinniversKitView.xctemplate/___FILEBASENAME___.swift
```diff
@@ -18,14 +18,21 @@ class ___FILEBASENAMEASIDENTIFIER___: <#View Class#> {
...
 
-    private func setup() {
+    // MARK: - Superclass Overrides
+    
+}
+
+// MARK: - Private
+
+private extension ___FILEBASENAMEASIDENTIFIER___ {
+    func setup() {
```
axldyb: This needs to be private.

Templates/FinniversKitView.xctemplate/___FILEBASENAME___.swift
```diff
@@ -18,14 +18,21 @@ class ___FILEBASENAMEASIDENTIFIER___: <#View Class#> {
...
         // Setup constraints/frames
     }
+}
 
-    // MARK: - Superclass Overrides
-
-    // MARK: - Private
+// MARK: - Public
 
+public extension ___FILEBASENAMEASIDENTIFIER___ {
```
axldyb: Can we have the public extension come before the private? Makes it easier when you open a view and start looking for what it can do.

Templates/FinniversKitView.xctemplate/___FILEBASENAME___.swift
axldyb: Never mind. Marking it private will render it inaccessible to the init methods. Not so sure about putting this in the private extension anymore. Should probably stay in the `Setup` section?

henrikholmsen: Yes it's innherited from the extension, so there is no need to explicitly use `private` 😊

Templates/FinniversKitView.xctemplate/___FILEBASENAME___.swift
henrikholmsen: I agree, thats better 👍

Templates/FinniversKitView.xctemplate/___FILEBASENAME___.swift
henrikholmsen: I think it would be nice to keep it in the `private extension` 🤔 I think its nice to keep the original class scope free of self defined methods and only have super class overrides there.

henrikholmsen: Also, setup code is long and boring so then you wouldn't have to scroll trough it as much 😂

henrikholmsen: Any thought @3lvis ?

axldyb: My initial thought was that the view setup (constraints etc) will be far away from the properties holding the views. Not sure if this is a big problem though. I agree that all methods created by us should stay in the public/private section.

3lvis:
Shippppp it! 🔥

PR #142, opened 2018-05-03
Remove unused frame

3lvis:
When using AutoLayout the frame is ignored.


ulrikus:
Will this look good on iPad, though?

3lvis:
I think it looks the same since the frames where ignored to begin with.

ulrikus:
But the `InlineConsentView` uses the frame to calculate its `intrinsicContentSize`. So when I try this in the FINN app the height (and width) is calculated wrong causing the view to look weird (compressed height) and I think it will look even worse when the broadcast is shown.

ulrikus:
Actually, it could be done like this in FinniversKit, but I did it the other way because I wanted it to be the same as the FINN app.

PR #143, opened 2018-05-04
Use UIEdgeInsets instead of Insets

3lvis:


Sources/Extensions/UIKit/LayoutHelpers.swift
```diff
@@ -4,32 +4,20 @@
...
 
-    public init(top: CGFloat = 0, leading: CGFloat = 0, bottom: CGFloat = 0, trailing: CGFloat = 0) {
-        self.top = top
-        self.leading = leading
-
-        // Required in order to ensure correct use against AutoLayout constants,
-        // where negative values are required in order to add a bottom and trailing inset.
-        self.bottom = bottom > 0 ? bottom * -1 : bottom
-        self.trailing = trailing > 0 ? trailing * -1 : trailing
+    public static var zero: UIEdgeInsets {
```
henrikholmsen: This class property already exists on `UIEdgeInsets`

PR #144, opened 2018-05-04
Popup consent

ulrikus:
# What?
Updated the former `ConsentView` to be more generic and reusable as a _popup_. This will also conform to the new consent version intended for MinFinn in a _popup-form_ both in design and naming, thusly named `PopupView`.

The _"topRightButton"_ and the _"linkButton"_ is set to hidden if the button text is set to `nil`/not set so that the same view can be used in more cases. 

In the demo you can easily switch between default data models by using one of the following examples in the demo:
`consentView.model = PopupConsentDefaultData.reccomendations.model`
`consentView.model = PopupConsentDefaultData.reccomendationsRejected.model`
`consentView.model = PopupConsentDefaultData.transparency.model`

### Discussion point:
Should we use the `PopupViewModel` approach so that a model is injected into the class or do people prefer to set the individual parameters, in lack of better explanation? 😅

- [ ] Clarify usage of the datamodel
- [x] Clarify design

# Why?
To update the new design, name and functionality for the `PopupView`. It is supposed to be as generic and reusable as possible.

# Screenshots
### iPhone X (`PopupConsentView`) vs iPhone 4 (Former `ConsentTransparencyView`)
iPhone X iPhone 4

### iPad Pro (PopupConsentView after pressing "Nei takk" in `PopupConsentView`)
iPad Pro


3lvis:
Is that how is going to look in iPad and iPhone? There's a lot of whitespace.

ulrikus:
For the iPhones: Yes
For the iPad: Not sure. I have asked Mona (the designer in the GDPR team), but haven't heard anything yet. I'll talk with her on Monday and clarify.

Sources/Fullscreen/PopupConsentView/PopupView.swift
```diff
@@ -4,15 +4,16 @@
...
 
-// MARK: - ConsentViewDelegate
+// MARK: - PopupViewDelegate
 
-public protocol ConsentViewDelegate: NSObjectProtocol {
-    func consentView(_ consentView: ConsentView, didSelectYesButton button: Button)
-    func consentView(_ consentView: ConsentView, didSelectNoButton button: Button)
-    func consentView(_ consentView: ConsentView, didSelectCancelButton button: Button)
+public protocol PopupViewDelegate: NSObjectProtocol {
+    func popupView(_ popupView: PopupView, didSelectBottomRightButton button: Button)
```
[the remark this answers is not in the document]
ulrikus: Yes, as generic as possible. As you see from the screenshots the buttonTitles are different for every case. Therefore, I think it can be confusing to use names like that.

ulrikus:
That's a good idea. I like those names better. 👍

Sources/Fullscreen/PopupConsentView/PopupViewModel.swift
```diff
@@ -0,0 +1,15 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import Foundation
+
+public protocol PopupViewModel {
```
axldyb: I vote for keeping the model. It will keep the implementation point leaner and it also matches how other views are set up in FinniversKit. However, I'm open to changing this, but then we should change all models, not just this one.

ulrikus: What do you think @3lvis ?

3lvis:
I think we were a bit too fast in deleting the ConsentTransparencyView, running 'carthage update' isn't working now.

ulrikus:
I have a branch that fixes that while adding the `PopupView` coming pretty soon.

3lvis:
Awesome, thanks! If it takes longer than expected would be nice to split the purpose of the PRs so they are more specific. One for fixing and one for adding the new component. :)

PR #146, opened 2018-05-04
[WIP] FrontpageGridView

3lvis:
Not sure why we didn't make this as a component to begin with but I think it would be good to encapsulate all the UI logic into one place.

artboard


axldyb:
Cool. Not sure why we didn't do this in the first place. I guess the ide of the FinniversKit has changed a bit since the beginning. How do you plan to incorporate the broadcast and the consent views into this?

3lvis:
Need to take a closer look at the broadcast implementation, ideally the logic isn’t handled here, when it comes to consent I think we need to include the required logic here.

ulrikus:
Not sure if it even belongs in the `FrontpageGridView`, but does the class support the `InlineConsentView`? And I guess the `Broadcast` would be inserted above the entire view?

3lvis:
No, it's hanging in limbo. 

What's missing here is adding the relevant GDPR bits as well as the broadcast. :)

3lvis:
I think this serves as a good starting place that we could expand later on, for now I would like to use it in my experiment project, could you have a look at it @finn-no/ios?

3lvis:
Closing in favor of https://github.com/finn-no/FinniversKit/pull/272

PR #150, opened 2018-05-10
Add dismiss instructions

3lvis:
a


Demo/Helpers.swift
```diff
@@ -36,3 +36,17 @@ extension IndexPath {
         }
     }
 }
+
+struct Helpers {
```
henrikholmsen: Maybe we could call this something like `Settings` instead of `Helpers`? 🤔

henrikholmsen: Then, perhaps, we could move the `lastSelectedIndexPath` in there as well.

3lvis: Settings sound something like preferences and this is just state. Maybe call it 	`State`?

henrikholmsen: True, `State` sounds right 😁

3lvis:
Changed one 🎆

PR #160, opened 2018-05-24
Changed font sizes to the new spec.

ingridvf:
**What?**
Changed font sizes

**Why?**
To make text more readable and flexible

_Note:
Update finnbox and filterkit._

After:
font-after

Before:
font-before

### DIFF
--- a/Sources/DNA/Font/Font.swift
+++ b/Sources/DNA/Font/Font.swift
@@ -12,7 +12,7 @@ enum FontType: String {
 }
 
 public extension UIFont {
-    /// FINNTypeWebStrippet-Medium with a size of 36 scaled for UIFontTextStyle.title1
+    /// FINNTypeWebStrippet-Medium with a size of 34 scaled for UIFontTextStyle.title1
     /// ## Usage:
     /// - It should only be used one T1 and it should be the first text element that catches the users attention.
     /// - It shall give the user an overview of which page he or she is located.
@@ -20,7 +20,10 @@ public extension UIFont {
     public static var title1: UIFont {
         registerCustomFonts()
 
-        let font = UIFont(name: FontType.medium.rawValue, size: 36.0)!
+        let font = UIFont(name: FontType.medium.rawValue, size: 34.0)!
+        if #available(iOS 11.0, *) {
+            return font.scaledFont(forTextStyle: .largeTitle)
+        }
         return font.scaledFont(forTextStyle: .title1)
     }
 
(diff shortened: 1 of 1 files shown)


3lvis:
Nice initiative, Ingrid, keep the PRs coming! 🔥

Sources/DNA/Font/Font.swift
```diff
@@ -34,66 +37,78 @@ public extension UIFont {
         registerCustomFonts()
 
         let font = UIFont(name: FontType.light.rawValue, size: 28.0)!
-        return font.scaledFont(forTextStyle: .title2)
+        return font.scaledFont(forTextStyle: .title1)
```
3lvis: What's the reason for changing this .title1 to .title2?

Sources/DNA/Font/Font.swift
```diff
@@ -34,66 +37,78 @@ public extension UIFont {
...
     /// ## Usage:
     /// - If a T2 have more sublevels, you can use T3.
     /// - This always has the weight Light.
     public static var title3: UIFont {
         registerCustomFonts()
 
-        let font = UIFont(name: FontType.light.rawValue, size: 22.5)!
-        return font.scaledFont(forTextStyle: .title3)
+        let font = UIFont(name: FontType.light.rawValue, size: 22)!
+        return font.scaledFont(forTextStyle: .title2)
```
3lvis: Same here, shouldn't this be title3?

ingridvf: Because in Human Interface guidelines their Title2 is the same side as our title3, so just to her the dynamic type scaling to work in the way they have put it

Sources/DNA/Font/Font.swift
ag12: https://developer.apple.com/ios/human-interface-guidelines/visual-design/typography/

Sources/DNA/Font/Font.swift
3lvis: Ahhh, I see!

3lvis:
The changes look pretty good, the only feedback is that it would be easier to work with the Apple system if we tried matching their conventions which I personally think are pretty good. If we added the "Large title" font type or "Heading" type then we could match title1, title2 and so on.

screen shot 2018-05-25 at 09 44 50

ag12:
Yes, very agreed :)

ingridvf:
Good idea. We should check with Android and Web to see if it works for them, so we are talking about the same size when we say Title1

ag12:
@henrikholmsen og @osanoj  Remember to update FilterKit when this is merged :)

henrikholmsen:
Will do @ag12  😁

PR #161, opened 2018-06-06
Snap behavior in EmptyView

ulrikus:
# What?
Adds a `UISnapBehavior` to all the movable objects in the `EmptyView` when the 3D touch force value surpasses **95%** of the maximum value. For devices not supporting 3D touch a longpress for **2.5 seconds** does the same thing.

To release the movable objects (remove the snap behavior) the user can either tap the screen outside of the blocks or shake the device (`shakeMotion`).

Please let me know if I should keep doing PRs like this or if you don't want it.

# Why?
It's fun and I haven't started on a new project yet. And also, I miss FINN.

# Gif
snapbehavior

### DIFF
--- a/Sources/Fullscreen/EmptyView/EmptyView.swift
+++ b/Sources/Fullscreen/EmptyView/EmptyView.swift
@@ -110,6 +110,12 @@ public class EmptyView: UIView {
         return itemBehavior
     }()
 
+    private var rectangleSnapBehavior: UISnapBehavior?
+    private var triangleSnapBehavior: UISnapBehavior?
+    private var roundedSquareSnapBehavior: UISnapBehavior?
+    private var circleSnapBehavior: UISnapBehavior?
+    private var squareSnapBehavior: UISnapBehavior?
+
     private lazy var motionManager: CMMotionManager = {
         let motionManager = CMMotionManager()
         motionManager.accelerometerUpdateInterval = 1 / 60
@@ -162,6 +168,12 @@ public class EmptyView: UIView {
 
     private func setup() {
         backgroundColor = .milk
+
+        let tapGesture = UITapGestureRecognizer(target: self, action: #selector(userHasTapped))
+        let longPressGesture = UILongPressGestureRecognizer(target: self, action: #selector(userDidLongPress))
+        longPressGesture.minimumPressDuration = 2.0
+        addGestureRecognizer(tapGesture)
+        addGestureRecognizer(longPressGesture)
 
         addSubview(rectangle)
         addSubview(triangle)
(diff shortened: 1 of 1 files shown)


3lvis:
This is awesome!

ag12:
I don't understand what you have done, but the code looks great. Also really cool and awesome feature. I say let's go for it.

ulrikus:
@ag12 A `UISnapBehavior` is basically a way of attaching an object to a point. The objects are using the same animator for the _"snappy animations"_ that creates the _"gravity", "collision"_ and the _"itemBehavior"_ (bouncyness, density, and other attributes on each object) in the same view. The rest is mainly detection of `3D touch` and `shakeMotion and dismissal of the snapBehavior`.

axldyb:
Why not release the objects when the 3D Touch / long tap ends? With the gravity normally controls the objects it feels like they should drop when the tap/handle cease to exist.

Sources/Fullscreen/EmptyView/EmptyView.swift
```diff
@@ -163,6 +169,11 @@ public class EmptyView: UIView {
     private func setup() {
         backgroundColor = .milk
 
+        let tapGesture = UITapGestureRecognizer(target: self, action: #selector(userHasTapped))
+        addGestureRecognizer(tapGesture)
+
+        becomeFirstResponder()
```
henrikholmsen: Just a question, why are you calling `becomeFirstResonder()` here?

henrikholmsen: Ah, ok. Just so you know, it does say in Apples documentation that you should 
[_never call this method on a view that is not part of an active view hierarchy_](https://developer.apple.com/documentation/uikit/uiresponder/1621113-becomefirstresponder).
So maybe you should move this call to after you've added the view as a subview? I'm sure everything works as it should now though, but it might cause unexpected behaviour.

ulrikus: I just read that myself. 😅Not entirely sure what it means, but I could move it to after the adding of subviews yes. Also, I can guard the `window` property in the `canBecomeFirstResponder` override so that it returns `false` if `window == nil`.

ulrikus:
@axldyb I guess it's because I wanted to have more weird/fun hidden features. I guess it's a bit _"gimmicky"_. I could implement your suggestion instead, in fact I have a version ready to be committed and pushed if desirable. That's a fun one too! 🙂If so I can remove the whole firstResponder stuff also. What do you think @henrikholmsen ?

Sources/Fullscreen/EmptyView/EmptyView.swift
henrikholmsen: Calling `becomefirstresponder` on the view after adding its subview wouldn't make any difference. You'd have to wait for calling it after the view itself has been added as a subview to a view thats in an active view hierarchy.

henrikholmsen:
Sounds like a good idea, makes more sense @ulrikus  😊

3lvis:
I think it's better to keep this codebase simple we might reconsider this in the future but for now we would have to skip this PR.

PR #163, opened 2018-06-08
Particle Emitter for EmptyView

ulrikus:
# What?
Adds a particle emitter that looks like sparks flying for when the blocks collide in EmptyView. The velocity must be greater than 500 (whatever that means) to produce a spark. The blocks will travel just fast enough if they are dropped from the topp of the screen and straight down on another object. There are three different sparks. The smallest sparks are for the lowest velocity on impact.

### Note
The sparking is highly inconsistent, as you can see on the gif below, a few times when the velocity are sufficiently high, there are no sparks and sometimes there are. I haven't figured out why, but I have some suggestions:
- The `collisionBehavior(_ behavior: UICollisionBehavior, endedContactFor item1: UIDynamicItem, with item2: UIDynamicItem)` method is called upon too soon so that the sparks have no time to show
- The way the emitter layer stopps the emitter cells (maybe setting the `birthRate` the way I did is incorrect).
- Black magic

**However**, I think it's fun that it does not show every time so that it's more of a hidden thing. The more exploring you do in the EmptyView the more fun stuff you discover. Although we want the user to actually use FINN, it's fun to make this an engaging little game.

I haven't tested how this works with the snap behavior in PR https://github.com/finn-no/FinniversKit/pull/161. Will wait until it is merged if desirable.

# Why?
Make EmptyView great (..again?).

# Gif
**NB!** The frame rate i a bit low so the sparks don't look as good on this gif as they do IRL. Also, the simulator was laggy, poor and annoying AF.

sparkcollision2


3lvis:
I think it's better to keep this codebase simple we might reconsider this in the future but for now we would have to skip this PR.

PR #166, opened 2018-06-25
updated sardine color

ingridvf:
What:
Changed Sardine Color to match Zeplin Spec

Why:
The color was outdated, and not very visible as a disabled state.

Before:
sardine before

After:
sardine after


ag12:
Great jobs 👋🏻🙏

PR #168, opened 2018-06-27
Add padding in text fields

ingridvf:
**_What_**
Added class for creating padding in text fields, and increased the padding.
**_Why:_** 
to match design spec, and have the same height as buttons.

Before: text field before

After:
text field after

### DIFF
--- a/Sources/Components/TextField/TextField.swift
+++ b/Sources/Components/TextField/TextField.swift
@@ -120,8 +120,8 @@ public class TextField: UIView {
 
     // MARK: - External properties
 
-    public lazy var textField: FKTextField = {
-        let textField = FKTextField()
+    public lazy var textField: UITextField = {
+        let textField = UITextField()
         textField.translatesAutoresizingMaskIntoConstraints = false
         textField.font = UIFont.body
         textField.textColor = .licorice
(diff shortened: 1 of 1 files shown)


ag12:
Could we have a second opinion here? 🗡

axldyb:
Yeah, looking at this now. I don't get the purpose of the FKTextField? Do we just want to have more space/padding inside the blue box?

ag12:
Yes, we need to override `textRect`, `placeholderRect` and `editingRect`

axldyb:
Why do we need to override that exactly? If we just want to have more padding then why not simply edit the constraints between the `textField` and the blue `textFieldBackgroundView`?

ag12:
Changing these?

```
            textField.topAnchor.constraint(equalTo: textFieldBackgroundView.topAnchor, constant: .mediumSpacing),
            textField.leadingAnchor.constraint(equalTo: textFieldBackgroundView.leadingAnchor, constant: .mediumSpacing),
            textField.trailingAnchor.constraint(equalTo: textFieldBackgroundView.trailingAnchor, constant: -.mediumSpacing),
            textField.bottomAnchor.constraint(equalTo: textFieldBackgroundView.bottomAnchor, constant: -.mediumSpacing)
```
We could. But Is that very "iOS" like? I guss this is a custom component so, why not?

ag12:
But does it make sense to change constraints.. for adding padding to a textViews text and placeholder? 🤔

axldyb:
The `textFieldBackgroundView` was added in the first place in order to achieve this. It already adds som padding. Now we will have padding in two places.

ag12:
Alright 🎖

ag12:
Then we need to `Address PR comments` @ingridvf  😅

axldyb:
Here you see the outlines of the views. The text field has some padding inside the ice blue `textFieldBackgroundView`. 
 
skjermbilde 2018-06-28 kl 13 14 56

ag12:
Yeah, it makes sense now. Think we need to ass some comment in the code.

ag12:
https://gph.is/2iDik3x

ingridvf:
I fix

ag12:
Next level.

PR #174, opened 2018-07-23
Selection boxes

shredlocker:
Created two types of selection boxes, single selection (Radiobox) and multi selection (Checkbox). 

Everything seems to work well and without any bugs for now. 

Can't preview the reverse animation in demo because of the double tap gesture recogniser, but it's there! 

Radio button animation should be shorter


Sources/Components/SelectionBoxes/Demo/CheckboxDemoView.swift
```diff
@@ -0,0 +1,94 @@
...
+        ])
+    }
+
+    @objc func dismiss() {
+        guard let doubleTap = superview?.gestureRecognizers?.first else { return }
+
+        if touchEnabled {
+            dismissButton.setTitle("Enable Double Tap", for: .normal)
+            touchEnabled = false
+            doubleTap.isEnabled = false
```
3lvis: I was thinking instead of this we could have a property in ViewController that says if it should support double tap or not. Then that property can be overwritten by the people that subclass it.

shredlocker: Aha! I'll fix that :)

Sources/Components/SelectionBoxes/AnimatedImageView.swift
```diff
@@ -0,0 +1,51 @@
...
+        super.startAnimating()
+        startTime = CACurrentMediaTime()
+    }
+
+    func cancelAnimation() {
+        super.stopAnimating()
+        // Get current animation frame
+        guard let startTime = startTime else { return }
+        let elapsed = CACurrentMediaTime() - startTime
+        let frame = Int(60.0 * elapsed)
```
osanoj: Can this be renamed? Anything that is called `frame` is making me think `CGRect` and views. Even just renaming it to `animationFrame` helps.

shredlocker: Yep, I'll rename it ;)

Sources/Components/SelectionBoxes/AnimatedImageView.swift
```diff
@@ -0,0 +1,51 @@
...
+        super.startAnimating()
+        startTime = CACurrentMediaTime()
+    }
+
+    func cancelAnimation() {
+        super.stopAnimating()
+        // Get current animation frame
+        guard let startTime = startTime else { return }
+        let elapsedTime = CACurrentMediaTime() - startTime
+        let currentAnimationFrame = Int(60.0 * elapsedTime)
```
3lvis: Would be nice to abstract 60 into a named variable.

Sources/Components/SelectionBoxes/AnimatedImageView.swift
```diff
@@ -0,0 +1,51 @@
...
+
+        let animationSequence = images[..<currentAnimationFrame].reversed()
+
+        let reverse = UIImageView(frame: bounds)
+        addSubview(reverse)
+
+        reverse.image = animationSequence.last
+        reverse.animationRepeatCount = 1
+        reverse.animationImages = Array(animationSequence)
+        reverse.animationDuration = Double(animationSequence.count) / 60.0
```
3lvis: Maybe we can reuse the variable here instead of the hardcoded number if that makes sense?

Sources/Components/SelectionBoxes/Demo/CheckboxDemoView.swift
```diff
@@ -0,0 +1,60 @@
...
+
+class CheckboxDemoView: UIView {
+    let strings = [
+        "Mistanke om svindel",
+        "Regelbrudd",
+        "Forhandler opptrer som privat",
+    ]
+
+    lazy var checkbox: Checkbox = {
+        let box = Checkbox(strings: self.strings)
```
3lvis: I think we are going to see what things we need when we implement this on the FINN app but we might need more than title strings then we might need to look into the DataSource pattern used in other pars of FinniversKit.

ag12:
+2,373 −4 
A major PR IMHO :)

Sources/Components/SelectionBoxes/Demo/CheckboxDemoView.swift
```diff
@@ -0,0 +1,60 @@
...
+            checkbox.heightAnchor.constraint(equalToConstant: CGFloat(strings.count + 1) * 44),
+        ])
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+extension CheckboxDemoView: SelectionboxDelegate {
```
3lvis: I think we need to rename the delegate to CheckboxDelegate and RadiobuttonDelegate using Selection feels strange.

Sources/Components/SelectionBoxes/Demo/CheckboxDemoView.swift
```diff
@@ -0,0 +1,60 @@
...
+        return box
+    }()
+
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+
+        Selectionbox.appearance().font = .body
+        Selectionbox.appearance().textColor = .licorice
+
+        let checkboxSelected = UIImage.animatedImageNamed("checkbox-selected-", duration: 20 / 60.0)
```
3lvis: I think we should replace this hardcoded numbers with named variables.

Demo/ViewController.swift
```diff
@@ -2,7 +2,7 @@
 //  Copyright © FINN.no AS, Inc. All rights reserved.
 //
 
-import UIKit
+import FinniversKit
 
 class ViewController<View: UIView>: UIViewController {
```
ag12: Should this be called `ViewController`? 
In `Demo/`? :)

Sources/Components/SelectionBoxes/Demo/CheckboxDemoView.swift
```diff
@@ -0,0 +1,60 @@
...
+        checkbox.unselectedAnimationImages = checkboxUnselected?.images
+
+        backgroundColor = .white
+        addSubview(checkbox)
+
+        NSLayoutConstraint.activate([
+            checkbox.leftAnchor.constraint(equalTo: leftAnchor),
+            checkbox.rightAnchor.constraint(equalTo: rightAnchor),
+            checkbox.topAnchor.constraint(equalTo: topAnchor),
+            checkbox.heightAnchor.constraint(equalToConstant: CGFloat(strings.count + 1) * 44),
```
3lvis: Same with the 44, might be worth to see how would this work blowing up the fonts in the accessibility settings, UISwitch might be a good guide for what to do here.

Sources/Components/SelectionBoxes/Selectionbox.swift
```diff
@@ -0,0 +1,178 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import UIKit
+
+/* Selection box for selecting a singel item */
```
3lvis: Small typo here :) should be single.

Sources/Components/SelectionBoxes/Selectionbox.swift
```diff
@@ -0,0 +1,178 @@
...
+
+public class RadioButton: Selectionbox {
```
3lvis: I think we should place RadioButton in its own file, same as Checkbox makes it easier to find the files using Quick Open.

Sources/Components/SelectionBoxes/Demo/RadioButtonDemoView.swift
```diff
@@ -0,0 +1,48 @@
...
+    ]
+
+    lazy var radioButton: RadioButton = {
+        let box = RadioButton(strings: self.strings)
+        box.title = "Radio Box Title"
+        box.translatesAutoresizingMaskIntoConstraints = false
+        return box
+    }()
+
+    private var touchEnabled = false
```
3lvis: I think we aren't using this one anymore

Sources/Components/SelectionBoxes/Selectionbox.swift
```diff
@@ -0,0 +1,178 @@
...
+
+public protocol SelectionboxDelegate: class {
+    func selectionbox(_ selectionbox: Selectionbox, didSelectItem item: SelectionboxItem)
+    func selectionbox(_ selectionbox: Selectionbox, didUnselectItem item: SelectionboxItem)
+}
+
+public class Selectionbox: UIView {
+    public var unselectedImage: UIImage? {
+        didSet {
+            for item in stack.arrangedSubviews as! [SelectionboxItem] {
```
3lvis: As a best practice I think it's better to avoid having forced unwrapping.

Sources/Components/SelectionBoxes/Selectionbox.swift
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    public override func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent?) {
+        guard let touch = touches.first else { return }
+        let location = touch.location(in: self)
+        guard let item = hitTest(location, with: event) as? SelectionboxItem else { return }
+        highlightedItem = item
+    }
+
+    public override func touchesEnded(_ touches: Set<UITouch>, with event: UIEvent?) {
```
3lvis: I really like how you handle touch detection, works quite well 👏

Sources/Components/SelectionBoxes/Selectionbox.swift
```diff
@@ -0,0 +1,178 @@
...
+            let item = SelectionboxItem(index: i)
+            item.label.text = string
+            stack.addArrangedSubview(item)
+        }
+
+        addSubview(titleLabel)
+        addSubview(stack)
+
+        NSLayoutConstraint.activate([
+            titleLabel.leftAnchor.constraint(equalTo: leftAnchor, constant: 16),
```
3lvis: Try using the constants in Spacing.swift instead of this hardcoded values. :)

Sources/Components/SelectionBoxes/Selectionbox.swift
shredlocker: I only put them in the same folder so I could use the fileprivate keyword for the method these classes are suppose to override

Sources/Components/SelectionBoxes/Selectionbox.swift
shredlocker: Thanks! :)

Demo/ViewController.swift
```diff
@@ -36,9 +54,21 @@ class ViewController<View: UIView>: UIViewController {
...
+            view.addGestureRecognizer(doubleTap)
+        } else {
+            let button = Button(style: .callToAction)
+            button.setTitle("Dismiss", for: .normal)
+            button.addTarget(self, action: #selector(didDoubleTap), for: .touchUpInside)
+            button.translatesAutoresizingMaskIntoConstraints = false
+            view.addSubview(button)
+            NSLayoutConstraint.activate([
+                button.centerXAnchor.constraint(equalTo: view.centerXAnchor),
+                button.bottomAnchor.constraint(equalTo: view.bottomAnchor, constant: -64),
```
3lvis: Can you change this to use `veryLargeSpacing` instead?

Sources/Components/SelectionBoxes/Selectionbox.swift
```diff
@@ -0,0 +1,199 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import UIKit
+
+/* Selection box for selecting a single item */
+
+public protocol RadioButtonDelegate: class {
+    func radioButton(_ radioButton: RadioButton, didSelectItem item: SelectionboxItem)
```
3lvis: I think this should be RadioButtonItem instead of SelectionboxItem same for the Checkbox.

Sources/Components/SelectionBoxes/Selectionbox.swift
```diff
@@ -0,0 +1,199 @@
...
+            let item = SelectionboxItem(index: i)
+            item.label.text = string
+            stack.addArrangedSubview(item)
+        }
+
+        addSubview(titleLabel)
+        addSubview(stack)
+
+        NSLayoutConstraint.activate([
+            titleLabel.leftAnchor.constraint(equalTo: leftAnchor, constant: Spacing.mediumLargeSpacing),
```
3lvis: Since we've implemented this as an extension of CGFloat as well you can drop the `Spacing` and just do `.mediumLargeSpacing`.

Sources/Components/SelectionBoxes/Selectionbox.swift
shredlocker: I thought maybe use index instead of item. There's not really anything useful in the items at the moment so maybe just make it an internal class for now?

Sources/Components/SelectionBoxes/SelectionboxItem.swift
```diff
@@ -0,0 +1,87 @@
...
+
+    let imageView: AnimatedImageView = {
+        let view = AnimatedImageView(frame: .zero)
+        view.contentMode = .scaleAspectFit
+        view.animationRepeatCount = 1
+        view.translatesAutoresizingMaskIntoConstraints = false
+        return view
+    }()
+
+    let label: UILabel = {
```
3lvis: I think we should rename this as "titleLabel" or "headingLabel" just label by itself doesn't have much meaning.

Sources/Components/SelectionBoxes/AnimatedImageView.swift
```diff
@@ -0,0 +1,71 @@
...
+    private var startTime: TimeInterval?
+    private var reverseImageView: UIImageView // Used for reversing an animation
+
+    // MARK: Implementation
+
+    override init(frame: CGRect) {
+        reverseImageView = UIImageView(frame: .zero)
+        super.init(frame: frame)
+        
+        reverseImageView.isHidden = true
```
3lvis: Much cleaner 👏

Sources/Components/SelectionBoxes/Selectionbox.swift
3lvis: I think you might need the item in case it has an ID for example and we'll most likely make it a protocol like we did with the other components. For now I think it would be nice the types so the radio button doesn't have to deal with Selectionbox items.

shredlocker:
checkboxanimation

radiobuttonanimation

3lvis:
Sweet tappable buttons!

200w_d

PR #178, opened 2018-07-30
Remove headers from framework

3lvis:
As Amir pointed out, it's not needed to have other headers than these ones, tested in the main iOS app. It works fine 👌.

screen shot 2018-07-30 at 15 14 29


3lvis:
As Amir pointed out, it's not needed to have other headers than these ones, tested in the main iOS app. It works fine 👌.

screen shot 2018-07-30 at 15 14 29

PR #179, opened 2018-07-31
Radio checkbox imageview

shredlocker:
Classes for creating RadioButton and Checkbox image animation. This way we can use these animation in multiple places. 

This feature was requested.


3lvis:
@shredlocker Can you close this PR now that we have discussed this issue?

PR #185, opened 2018-08-02
Adds confirmation view

shredlocker:
View confirming that your message is sent

### DIFF
--- a/Sources/Fullscreen/AdReporter/AdReporterConfirmationView.swift
+++ b/Sources/Fullscreen/AdReporter/AdReporterConfirmationView.swift
@@ -1,0 +1,93 @@
+//
+//  Copyright © 2018 FINN.no. All rights reserved.
+//
+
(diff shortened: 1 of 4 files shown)


Sources/Fullscreen/AdReporter/AdReporterConfirmationView.swift
```diff
@@ -0,0 +1,93 @@
...
+    }
+    
+    public var buttonTitle: String? {
+        get { return closeButton.titleLabel?.text }
+        set { closeButton.setTitle(newValue, for: .normal) }
+    }
+    
+    // MARK: - Action
+    
+    public var action: (() -> Void)?
```
3lvis: Can we use a delegate pattern instead of a closure so it's consistent with the rest of FinniversKit?

3lvis:
Good job!

PR #197, opened 2018-08-10
Remove unused LoginView and Register view

3lvis:
These have been replaced with Schibsted Account SDK https://github.com/schibsted/account-sdk-ios


3lvis:
a

PR #198, opened 2018-08-13
Snapshot tests

shredlocker:
Snapshot tests for every component. Missing BroadcastContainer test. Could potentially change from using the ViewController class to making stand alone views with frames.


3lvis:
I wonder if it would make sense to just add all the snapshot tests under one file. 🤔

3lvis:
So it's consistent with what we're doing with Demo.swift as well.

shredlocker:
That can be done

3lvis:
Much better 👌

PR #202, opened 2018-08-13
Make Cartfile.private and point to release instead of master

3lvis:
Changes Cartfile to be Cartfile.private this will indicate Carthage that it's an internal dependency, otherwise it will be included when running `carthage update` in the main FINN app.

Also changed Cartfile.private to point to a release instead of master, it's more secure.


3lvis:
Merging to make a PR in the main FINN app.

PR #208, opened 2018-08-14
Change frontpage column sizing in iPad

3lvis:
ipad

### DIFF
--- a/Demo/DemoHelpers.swift
+++ b/Demo/DemoHelpers.swift
@@ -163,7 +163,7 @@ enum Sections: String {
         case .dna, .components, .fullscreen:
             return .fullscreen
         case .recycling:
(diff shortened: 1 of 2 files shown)


ingridvf:
Looks much better! Have you considered 3 ads instead of 4 in landscape mode? With bigger images, like what we have on desktop? Can still be a bit overwhelming with that many ads visible in the screen.

3lvis:
# 4 columns
simulator screen shot - ipad air - 2018-08-16 at 07 54 56

# 3 columns
simulator screen shot - ipad air - 2018-08-16 at 07 55 13

3lvis:
I don't have a device at hand but we could try 3 columns as well :)

PR #209, opened 2018-08-20
Refactor broadcast

shredlocker:
Made it work pretty well. I'm open for any suggestions if there are things that could be done better! I removed the datasource approach, I think it only makes everything a bit more complex


Cartfile
```diff
@@ -0,0 +1,2 @@
+# FBSnapshotTestCase
+github "uber/ios-snapshot-test-case" "master"
```
3lvis: This should be Cartfile.private

shredlocker:
I'll redo it

shredlocker:
Jepp, something have gone wrong with this one at some point!

PR #210, opened 2018-08-21
Improved Broadcast

shredlocker:
This PR should not be broken, I hope...


3lvis:
I think we should rename BroadcastContainer to just Broadcast, since we don't plan to use a single Broadcast.

Sources/Components/Broadcast/Broadcast.swift
```diff
@@ -4,250 +4,147 @@
...
-    public override init(frame: CGRect = .zero) {
-        super.init(frame: frame)
-        setup()
-    }
-
-    public required init?(coder aDecoder: NSCoder) {
-        super.init(coder: aDecoder)
-        setup()
-    }
+public protocol BroadcastContainerDelegate: class {
```
3lvis: Here still says Container :)

3lvis:
Epic work, @shredlocker! 👏

shredlocker:
Should I merge this now? I'm not really happy about how it all turned out code wise, it should work fine in the app as it is now

3lvis:
🚀 🚀 🚀 🚀 🚀

PR #212, opened 2018-08-22
Extend README with interesting things

3lvis:
# Why?

Important to share the value behind following an approach like FinniversKit.

# What?

Adds an initial version on reasoning on what interesting things could be found in our project.

# Show me

No UI changes.


README.md
```diff
@@ -24,13 +24,35 @@ Import the framework to access all the components.
...
-## Snapshot Testing
+## Interesting things
+
+### Folder structure (tests and demo)
+
+Inspired by Kickstarter (and Ruby on Rails) we place relevant files next to each other, even if these files are placed in different targets. We believe that having files closer together makes you aware of the different elements that might be affect after a change.
+
+### Delegates and data sources (instead blocks and injection)
+
+In order to maintain consistency we have opted out for using data sources for giving data to our recyclabe views and using delegates to interact for actions inside views.
```
osanoj: Should be "... we have opted for using data sources ..." without "out"?

3lvis: Good point, fixed :)

PR #213, opened 2018-08-30
Fix crash when a cached index path is not found

3lvis:


3lvis:
Thanks for the inspiration for this @osanoj 🙈🙈

PR #216, opened 2018-08-30
SavedSearchesListView

3lvis:
screen shot 2018-08-30 at 04 19 18

### DIFF
--- a/Sources/Recycling/ListViews/SavedSearches/Cell/SavedSearchesListViewCell.swift
+++ b/Sources/Recycling/ListViews/SavedSearches/Cell/SavedSearchesListViewCell.swift
@@ -1,0 +1,104 @@
+import UIKit
+
+@objc protocol SavedSearchesListViewCellDelegate: class {
+    func savedSearchesListViewCell(_ savedSearchesListViewCell: SavedSearchesListViewCell, didPressSettingsButton button: UIButton)
+}
+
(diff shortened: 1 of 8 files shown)


Sources/Recycling/ListViews/SavedSearches/Cell/SavedSearchesListViewCell.swift
```diff
@@ -0,0 +1,104 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    @objc func showSettings(button: UIButton) {
+        delegate?.savedSearchesListViewCell(self, didPressSettingsButton: button)
+    }
+
+    // MARK: - Dependency injection
+
+    public var model: SavedSearchesListViewModel? {
```
osanoj: There isn't much reason for the cell to keep a reference to the model is it? This could then just be a func to set values to show. But no big deal :)

3lvis: I agree but I think it's better to be consistent instead of being pragmatic when it comes to building the components makes it quicker to make decisions and also to understand code if all is build in the "same" way.

osanoj: Yes, if it is a consistent pattern it is good to keep doing it the same everywhere 👍

PR #217, opened 2018-08-30
Split style and color

3lvis:
Doing 

```swift
let label = Label(style: .detail)
```

Feels normal in the iOS world, UIButton works this way for example (check `UIButton(type: ...)`), here you split shape and form from other attributes such as tinting/color and frame size.

What we have instead is this 

```swift
let label = Label(style: .detail(.stone))
```

This is a strange pattern for a few reasons:
- You're mixing scope here and it's not necessarily clear what styles will accept which colors. 
- You can always overwrite the font color even if you try to disallow it.

Example: 
```swift
let label = Label(style: .detail(.stone))
label.textColor = .blue
```

It's also when working with labels more natural to separate color from the typeface, for this reasons, I propose this PR as a change in how we deal with the Label component.


Sources/Components/Label/Label.swift
```diff
@@ -44,11 +48,28 @@ public class Label: UILabel {
...
+            textColor = color
+        }
     }
 
     // MARK: - Dependency injection
 
     public var style: Style?
+
+    // Required for backwards compatibility https://github.com/finn-no/FinniversKit/pull/217
+    private func color(for style: Style?) -> UIColor? {
```
osanoj: Instead of keeping this for backwards compatibility, how about setting licorice as default color? We should normally never use labels with nil/black textColor anyway, right?

shredlocker:
I agree, this is alot better!

Sources/Components/Label/Label.swift
```diff
@@ -44,11 +48,28 @@ public class Label: UILabel {
...
         font = style?.font
+
+        if let color = color(for: style) {
+            textColor = color
+        }
     }
 
     // MARK: - Dependency injection
 
     public var style: Style?
```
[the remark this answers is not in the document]
3lvis: Good point, I can see both cases being valid. I'll go for making the private(set) to not modify too much the behavior.

Sources/Components/Label/Label.swift
3lvis: Good point, I'll change it :D

PR #219, opened 2018-08-30
Improve tests so it warns you about untested components

3lvis:
A common issue at the moment with snapshot testing is that it won't nag you when you add a new component and forget to add a test for it. Thanks to this PR if you add a component and forget to add the respective test for it you'll get the following message.

screen shot 2018-08-30 at 13 14 53


3lvis:
Tried refactoring to remove the duplicated logic between test classes but then it made everything more complex for little value.

PR #221, opened 2018-09-06
Removed print

ag12:
**_What?_** 
Removed print("UIFont+:  Failed to register font - register graphics font failed - this font may have already been registered in the main bundle.")

Until we have set up loggers we should remove this specific print line.


3lvis:
There's a whole empty print 

```swift
if CTFontManagerRegisterGraphicsFont(fontRef, &errorRef) == false {
    print("UIFont+:  Failed to register font - register graphics font failed - this font may have already been registered in the main bundle.")
}
```

PR #222, opened 2018-09-07
HackDay: Error view

3lvis:


Sources/EasterEgg/FlappyCat/Demo/FlappyCatViewDemoView.swift
```diff
@@ -0,0 +1,47 @@
...
+                break
+            }
+        }
+        return (responder as? UIViewController)!
+    }
+}
+
+extension FlappyCatViewDemoView: FlappyCatViewDelegate {
+    public func flappyCatViewDidPressBackButton(_ flappyCatView: FlappyCatView) {
+        if let viewController = FlappyCatViewDemoView.viewController(self) as? DemoViewController<FlappyCatViewDemoView> {
```
vadymmarkov: Do we really need to access view controller from a view?

3lvis: It’s a hack, I’ll find a better way to handle this.

PR #223, opened 2018-09-09
Bottom sheet

3lvis:
artboard

Moved from FilterKit.


3lvis:
It doesn't work as well as I would like but it's a start :)

Sources/Components/BottomSheet/BottomSheetInteractiveDismissalController.swift
```diff
@@ -0,0 +1,83 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import UIKit
+
+public final class BottomSheetInteractiveDismissalController: UIPercentDrivenInteractiveTransition {
+    let dismissalTransitioningRect: CGRect
```
vadymmarkov: Could those 4 properties be private? or they're used somewhere outside of this class?

Sources/Components/BottomSheet/BottomSheetInteractiveDismissalController.swift
```diff
@@ -0,0 +1,83 @@
...
+        case .ended:
+            isInteractionActive = false
+            if dismissalPercentage < 0 {
+                cancel()
+            } else if percentComplete >= dismissalPercentageThreshold {
+                finish()
+            } else {
+                if #available(iOS 11.0, *) {
+                } else {
+                    // There is a bug for version earlier than iOS 11 that needs completion speed to be something else than 1 to avoid "double" cancel dismissal animation
```
vadymmarkov: 😱

Sources/Components/BottomSheet/BottomSheetInteractiveDismissalController.swift
3lvis: Good point! I should take some time to go through the access control.

3lvis:
Closing seeing as this implementation isn't good enough.

PR #226, opened 2018-09-10
Rename ViewController and improve initalizers

3lvis:
Having a class called "ViewController" isn't necessarily the best idea, it's too generic and can cause confusion, this PR aims to fix that. Also, this PR changes the wording of the initializer `usingDoubleTap` to `withDismissButton`. 

By default components in FinniversKit are dismissed using double tap, for some controllers where testing double tap is important (like checkboxes) this was an issue, that's why we introduced the initializer `usingDoubleTap: false`, sadly, this didn't show the true intention, since using `usingDoubleTap: false` would mean that instead of double tapping for dismiss, you would get a "Dismiss" button instead. On this PR this will change so this behavior is more clear since it entails what will happen if you trigger this effect. Also if you check the initializers you can see that the default behavior is `usingDoubleTapToDismiss = true`, which is what will happen if you don't do anything.

image

### DIFF
--- a/Sources/Resources/FinniversImageAsset.m
+++ b/Sources/Resources/FinniversImageAsset.m
@@ -7,7 +7,6 @@
 
 #import "FinniversImageAsset.h"
 
-NSString *const FinniversImageAssetNoImage = @"NoImage";
 NSString *const FinniversImageAssetarrowDown = @"arrowDown";
 NSString *const FinniversImageAssetarrowRight = @"arrowRight";
 NSString *const FinniversImageAssetarrowUp = @"arrowUp";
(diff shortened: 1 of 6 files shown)


Sources/Resources/FinniversImageAsset.m
```diff
@@ -52,6 +51,7 @@
 NSString *const FinniversImageAssetmore = @"more";
 NSString *const FinniversImageAssetmoreImg = @"moreImg";
 NSString *const FinniversImageAssetmoteplassen = @"moteplassen";
+NSString *const FinniversImageAssetnoImage = @"noImage";
```
3lvis: I have no idea why this changed but it changed.

PR #228, opened 2018-09-17
Feature/saved search settings

3lvis:


3lvis:
final

3lvis:
This PR is becoming something else. I'll reopen a new with only the saved search settings.

PR #230, opened 2018-09-17
Proposal for app friendly text fields

3lvis:
The current text fields are a bit web-ish, I tried to make them more rounded and app-ish as a proposal on what we could do. Thoughts?

screen


shredlocker:
Looks good to me!

PR #236, opened 2018-09-26
Easter eggs: drum machine view and trigger button

vadymmarkov:
## Why?
We're adding Easter eggs to the FINN app to make it more playful, so users can find hidden features when they search for specific things.

## What?
- `EasterEggButton` component to start playing with discovered Easter egg. The button also shows how many times you can play with the current Easter egg.
- `DrumMachineView` fullscreen view that lets users create beats 🥁 

## How it looks
| `EasterEggButton` | `DrumMachineView` |
| --- | --- |
| simulator screen shot - iphone x - 2018-09-26 at 12 31 43 | simulator screen shot - iphone x - 2018-09-26 at 12 31 53 |


Sources/Components/EasterEggButton/Demo/EasterEggButtonDemoView.swift
```diff
@@ -0,0 +1,37 @@
...
+
+    // MARK: - Setup
+
+    private func setup() {
+        addSubview(easterEggButton)
+
+        NSLayoutConstraint.activate([
+            easterEggButton.centerXAnchor.constraint(equalTo: centerXAnchor),
+            easterEggButton.centerYAnchor.constraint(equalTo: centerYAnchor),
+            easterEggButton.widthAnchor.constraint(equalToConstant: 64),
```
3lvis: Could you change it to use `.veryLargeSpacing` instead of 64?

Sources/Fullscreen/EasterEggs/DrumMachine/Demo/DrumMachineDemoView.swift
```diff
@@ -0,0 +1,48 @@
...
+            drumMachineView.start()
+        }
+    }
+
+    // MARK: - Setup
+
+    private func setup() {
+        addSubview(drumMachineView)
+
+        NSLayoutConstraint.activate([
```
3lvis: You could use `drumMachineView.fillInSuperview()` instead of this. We didn't add more AutoLayout convenience methods but for sure we're open for suggestions :)

vadymmarkov: I think it's the most useful one 😉

Sources/Fullscreen/EasterEggs/DrumMachine/DrumMachineView.swift
```diff
@@ -0,0 +1,206 @@
...
+        slider.tintColor = .pea
+        return slider
+    }()
+
+    // MARK: - Init
+
+    public override init(frame: CGRect) {
+        super.init(frame: frame)
+        setupSubviews()
+        setupConstraints()
```
3lvis: We've been combining both `setupSubviews` and `setupConstraints` in a `setup` method, I guess the reason we've been doing that it's mostly because you aren't supposed to use one method without the other.

Sources/Fullscreen/EasterEggs/DrumMachine/DrumMachineView.swift
```diff
@@ -0,0 +1,206 @@
...
+        for (_, instrumentComposition) in compositions.enumerated() {
+            if instrumentComposition.value[currentPad] {
+                instrumentComposition.key.play()
+            }
+        }
+    }
+
+    private func makeEmptyCompositions() -> [Instrument: [Bool]] {
+        return [
+            .kick: Array(repeating: false, count: self.numberOfPads),
```
3lvis: Is the `self.` needed here? 🤔

Sources/Fullscreen/EasterEggs/DrumMachine/DrumMachineView.swift
```diff
@@ -0,0 +1,206 @@
...
+        bpmSlider.addTarget(self, action: #selector(sliderValueChanged(_:)), for: .valueChanged)
+        collectionView.reloadData()
+    }
+
+    private func setupConstraints() {
+        NSLayoutConstraint.activate([
+            selectorView.topAnchor.constraint(equalTo: topAnchor),
+            selectorView.leadingAnchor.constraint(equalTo: leadingAnchor),
+            selectorView.trailingAnchor.constraint(equalTo: trailingAnchor),
+            selectorView.heightAnchor.constraint(equalToConstant: 100),
```
3lvis: Could we abstract `100` into a variable?

Sources/Fullscreen/EasterEggs/DrumMachine/DrumMachineView.swift
```diff
@@ -0,0 +1,206 @@
...
+    private lazy var collectionViewLayout: UICollectionViewLayout = {
+        let layout = UICollectionViewFlowLayout()
+        layout.scrollDirection = .vertical
+        layout.minimumInteritemSpacing = cellSpacing
+        layout.minimumLineSpacing = cellSpacing
+        layout.sectionInset = UIEdgeInsets(top: padSpacing, left: padSpacing, bottom: padSpacing, right: padSpacing)
+        return layout
+    }()
+
+    private lazy var bpmSlider: UISlider = {
```
3lvis: Could we rename `bpmSlider` to `beatsPerMinuteSlider`? Is that what stands for? 😁

Sources/Fullscreen/EasterEggs/DrumMachine/DrumPadCollectionViewCell.swift
```diff
@@ -0,0 +1,57 @@
+//
+//  Copyright © 2018 FINN.no. All rights reserved.
+//
+
+import UIKit
+
+final class DrumPadCollectionViewCell: UICollectionViewCell {
+    private lazy var overlayLayer: CALayer = {
```
3lvis: What's the reason to use a CALayer instead of a UIView?

Sources/Fullscreen/EasterEggs/DrumMachine/DrumMachineView.swift
vadymmarkov: It probably was a lazy var at first, but now it's not needed since it's a function.

Sources/Fullscreen/EasterEggs/DrumMachine/InstrumentSelectorView.swift
```diff
@@ -0,0 +1,76 @@
...
+        return segmentedControl
+    }()
+
+    // MARK: - Init
+
+    init(instrument: Instrument) {
+        self.selectedInstrument = instrument
+        super.init(frame: .zero)
+        setupSubviews()
+        setupConstraints()
```
3lvis: Same comment here for the `setup` method :)

Sources/Fullscreen/EasterEggs/DrumMachine/DrumMachineView.swift
vadymmarkov: I usually split it in two only when `setup` grows out of proportion. But you're right that they are binded together.

PR #237, opened 2018-09-26
LoadingView: SVProgressHUD replacement.

3lvis:
# Why?

We've been using a third-party library for a quite straightforward element. In order to get rid of that library, we need to introduce a branded alternative.

# What?

A `LoadingView` is a screen h that works in a very similar manner to `SVProgressHUD` with the main difference that it uses our branded `LoadingIndicatorView`.

# Show me

### `LoadingView.show()`
new

### `LoadingView.shouldUseOldIndicator = true`
old


Sources/Fullscreen/LoadingView/LoadingView.swift
```diff
@@ -0,0 +1,101 @@
...
+
+    /// Stops the animation of the loading indicator and removes the loading view.
+    public class func hide() {
+        LoadingView.shared.stopAnimating()
+    }
+}
+
+// MARK: - Private methods
+
+extension LoadingView {
```
vadymmarkov: You could also make this extension private `private extension LoadingView` and remove private from methods in it.

3lvis: I made the extension private but I prefer to keep methods private unless they *need* to be internal or public.

vadymmarkov: Yeah, but I meant that methods in `private extension` become private automatically.

osanoj: This is probably something we should add to coding guidelines then. For instance we have tried to only specify on extension level as much as possible in filters. 

But I agree, if the extension is bigger than a couple of methods it makes it harder to see quickly when it is only on extension level.

PR #238, opened 2018-09-27
Xcode 10

3lvis:


Sources/Components/Broadcast/BroadcastItem.swift
```diff
@@ -139,3 +139,9 @@ extension BroadcastItem: UITextViewDelegate {
         return false
     }
 }
+
+// Helper function inserted by Swift 4.2 migrator.
+fileprivate func convertToOptionalNSAttributedStringKeyDictionary(_ input: [String: Any]?) -> [NSAttributedString.Key: Any]? {
```
vadymmarkov: This function looks weird 😅 Should we perhaps change `BroadcastItem.Style.linkTextAttributes` to `[NSAttributedStringKey.foregroundColor: UIColor.primaryBlue]` (without rawValue)?

Sources/Fullscreen/EmptyView/EmptyView.swift
```diff
@@ -225,7 +225,7 @@ public class EmptyView: UIView {
 
         let location = sender.location(in: self)
         let touchLocation = sender.location(in: objectView)
-        let touchOffset = UIOffsetMake(touchLocation.x - objectView.bounds.midX, touchLocation.y - objectView.bounds.midY)
+        let touchOffset = UIOffset.init(horizontal: touchLocation.x - objectView.bounds.midX, vertical: touchLocation.y - objectView.bounds.midY)
```
vadymmarkov: We can drop `init` here.

Sources/Components/Broadcast/BroadcastItem.swift
3lvis: wow, Xcode!

3lvis:
Snapshot tests not working.

Torsph:
@3lvis It became green for me if I did one run with `recordMode = true`.

3lvis:
@finn-no/ios This is ready for review. :D

3lvis:
Tests run normally on my computer.

3lvis:
We're having some issues with Broadcast snapshot testing because since it's using a Set the order keeps changing when the tests are done. I'm disabling snapshot testing for that component and will submit a PR to move from Set to Array in a different one/

PR #242, opened 2018-09-30
Improve loading view and improve demo

3lvis:
# Why?

Improved support for going from loading to success

# What?

Improves transition between states, this potentially could be refactored for better state handling but not sure if we should wait until we add "failed" for example.

# Show me

all


3lvis:
Okay, this time I'm pointing to the branch.

Sources/Fullscreen/LoadingView/Demo/LoadingViewDemoView.swift
```diff
@@ -22,18 +69,37 @@ public class LoadingViewDemoView: UIView {
...
 
     private func setup() {
-        addSubview(startLoading)
-        NSLayoutConstraint.activate([
-            startLoading.bottomAnchor.constraint(equalTo: bottomAnchor, constant: -.largeSpacing),
-            startLoading.centerXAnchor.constraint(equalTo: centerXAnchor)
-            ])
+        addSubview(tableView)
+        tableView.fillInSuperview()
+        tableView.dataSource = self
```
robinsalehjan: Nitpicking, but can we move this into the computed `tableView` property?

Sources/Fullscreen/LoadingView/Demo/LoadingViewDemoView.swift
```diff
@@ -22,18 +69,37 @@ public class LoadingViewDemoView: UIView {
...
+        tableView.delegate = self
```
robinsalehjan: And this one as well.

3lvis:
Adressed PR comments :)

3lvis:
🎉

PR #243, opened 2018-10-01
Review design changes

Torsph:
# Why?
Needed to match the current design of the pick-a-buyer functionality. 

# What?
I have updated the code to match the new design from Nowik. Would be nice to get some discussion around the `StandaloneSelectionboxItem.swift`. The `Selectionbox` currently needs to be have a header and a title for each item. I just need the button functionality. How this should be solved need to be up to you guys. 

# Show me

### Before
screenshot 2018-10-01 at 12 39 44

_...._

### After
screenshot 2018-10-01 at 12 22 53
_...._


Torsph:
Are going to update the screenshot tests :)

Sources/Components/SelectionBoxes/SelectionIndicatorView.swift
```diff
@@ -0,0 +1,123 @@
...
+    private func setupSubviews() {
+        translatesAutoresizingMaskIntoConstraints = false
+        addSubview(imageView)
+        imageView.fillInSuperview()
+    }
+
+    private func setupStyle() {
+        let framesPerSecond = 60.0
+
+        let radiobuttonSelected = UIImage.animatedImageNamed(type.selected, duration: 13 / framesPerSecond)
```
Torsph: FYI, this wont work when the framework is used since `animatedImageNamed` uses the main bundle.

Sources/Components/SelectionBoxes/SelectionIndicatorView.swift
```diff
@@ -0,0 +1,123 @@
...
+            return
+        }
+
+        imageView.isHighlighted = selected
+        imageView.animationDuration = selected ? imageView.selectedDuration : imageView.unselectedDuration
+        imageView.startAnimating()
+    }
+}
+
+public enum SelectionBoxType {
```
3lvis: Could you move this into `SelectionIndicatorView` and just call it `Type`?

3lvis: SelectionIndicatorView works so well ❤️

Sources/Components/SelectionBoxes/SelectionIndicatorView.swift
3lvis: mmmm you say it doesn't work but it works when I tried... so what do you mean?

3lvis:
Discussed with Thomas and Ingrid on trying a different approach with less "See chat", "See chat", "See chat", "See chat", "See chat", "See chat", "See chat", "See chat", "See chat", "See chat", "See chat", "See chat", "See chat".

Sources/Components/SelectionBoxes/SelectionIndicatorView.swift
Torsph: It does not work from the FINN app because it uses the FinniverseKit as a Framework. But the Demo app is using the code directly.

3lvis: If you try to report an ad from the FINN app it seems to be working. 🤔

Torsph:
Hmm, ok. I got views without images when I used it. Will have a look
tomorrow. :)

Sources/Components/SelectionBoxes/SelectionIndicatorView.swift
Torsph: I found out why it works in the ReporterView. It manually gets all the images instead of using `UIImage.animatedImageNamed`
```
for i in 0 ..< 13 {
    if let image = UIImage(named: "radiobutton-select-\(i)", in: FinniversKit.bundle, compatibleWith: nil) {
        radioButtonSelected.append(image)
    }
}

for i in 0 ..< 10 {
    if let image = UIImage(named: "radiobutton-unselected-\(i)", in: FinniversKit.bundle, compatibleWith: nil) {
        radioButtonUnselected.append(image)
    }
}
```

3lvis: Ahhhh sneaky!

Torsph:
@3lvis Is this ok now?

3lvis:
Great job, Thomas! Clean, simple and functional. :clap:

PR #248, opened 2018-10-03
Update Favorite icon

3lvis:
# Why?

Request by Thuy. Also, it looks better 😁. 

# What?

Changed gray button (usually used for the list that shows up when you select favorites). 

# Show me

screen shot 2018-10-03 at 16 03 40


3lvis:
Merging since it's not much to review.

PR #249, opened 2018-10-03
Fix collision with method name

3lvis:


3lvis:
There's an issue with using 

`@objc public class func named(_ imageName: String) -> UIImage {`

The Swift API

`UIImage(named: "String")`

Gets confused... renaming to `UIImage(assetNamed: "String")` fixes this issue.

PR #250, opened 2018-10-09
Add dismissal gesture recognizer for horizontal slide

3lvis:
# Why?

When a screen gets presented it feels natural to try to dismiss it by pulling to the opposite side from where it came from.

# What?

Adds a swipe gesture recognizer to dismiss the controller, tried using `UIViewControllerInteractiveTransitioning` but couldn't get the canceling part to work the way I wanted so I'm using this as a fallback, I'll keep trying mostly for learning purposes and see if I make a new PR when I nail it.

# Show me

No UI changes really, just behavior.


osanoj:
It would be nice with a delegate protocol so that the presenter can get a callback whenever the view disappears (also when disappear because of button tap). Not required for this PR of course 🙂

PR #253, opened 2018-10-09
Make HorizontalSlider available on ObjC

3lvis:


3lvis:
Merging since it's minor changes.

PR #256, opened 2018-10-16
Fixes broken behaviour on iOS version below 11

shredlocker:
# Why?

Cell and header views in settings view are acting weird on iOS version lower than 11.

# What?

- Set estimated row height and section header height for table view.
- Set hugging priorities to make everything look good.


3lvis:
Can you add a comment saying that this is needed only on iOS 11 and bellow?

PR #257, opened 2018-10-16
Adds comment

shredlocker:
# Why?

Makes it easier for the future when we drop iOS 10

# What?

Adds a comment saying which properties are only needed for iOS version below 11


osanoj:
I think the `setContentHuggingPriority` had more to do with small screens than iOS version, this is the only stuff needed for pre iOS 11

```
         if #available(iOS 11.0, *) {
             // Not necessary >= iOS 11
         } else {
             tableView.rowHeight = UITableViewAutomaticDimension
             tableView.estimatedRowHeight = 44
         }
```

Is how I did in filters.

PR #258, opened 2018-10-17
Component/tableviewcells

bstien:
# Why?
Many of the cells in the FINN app are inconsistent in both size, font and spacing. To get us back on track I've had a collab with Thuy to standardize how most of our cells should look and feel.

The plan is to make the cells:
- Modifiable
- Simple to use
- A template for new types of cells

I think the way to go with this is to make some base classes that fits most of the use cases, and sub-class/modify if needed.

# What?
In this PR I've started implementing 3 of the cells, along with a demo of them. The three celltypes are:

- Basic
- Checkbox
- Checkbox w/ subtitle

I'd love to get feedback on how we can make them even more modifiable and reusable, along with naming conventions 👌

# Show me
### Before
Nothing before..!

### After
screen shot 2018-10-17 at 13 16 45
screen shot 2018-10-17 at 13 16 54
screen shot 2018-10-17 at 13 16 56


Sources/Components/TableViewCells/BasicTableViewCell.swift
```diff
@@ -0,0 +1,43 @@
...
+            titleLabel.topAnchor.constraint(equalTo: contentView.topAnchor, constant: 13),
+            titleLabel.leadingAnchor.constraint(equalTo: contentView.leadingAnchor, constant: .mediumLargeSpacing),
+            titleLabel.bottomAnchor.constraint(equalTo: contentView.bottomAnchor, constant: -13),
+            ])
+    }
+
+    public func configure(with viewModel: BasicTableViewCellViewModel) {
+        titleLabel.text = viewModel.title
+        separatorInset = .leadingInset(.mediumLargeSpacing)
+
```
robinsalehjan: Maybe remove the extra space?

3lvis:
In order for snapshots to work properly we might need to either have all the cells in one table, or have one Demo per each cell type. What do you think we should do, Bastian?

Sources/Components/TableViewCells/Demo/TableViewCellsDemoView.swift
```diff
@@ -0,0 +1,112 @@
...
+            basicCellDemoView.fillInSuperview()
+            checkboxCellDemoView.fillInSuperview()
+            checkboxSubtitleCellDemoView.fillInSuperview()
+            
+            currentCellType = .basic
+        }
+    }
+    
+    @objc private func segmentedControlChangedValue() {
+        self.currentCellType = TableViewCellType.init(rawValue: self.segmentedControl.selectedSegmentIndex)!
```
robinsalehjan: Remove the `.init` for consistency :)

Sources/Components/TableViewCells/Demo/TableViewCellsDemoView.swift
```diff
@@ -0,0 +1,112 @@
...
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import FinniversKit
+
+fileprivate enum TableViewCellType: Int {
+    case basic, checkbox, checkboxSubtitle
+}
+
+protocol HapticFeedbackProducer: AnyObject {
```
3lvis: Why using a protocol for this?

3lvis:
Really like the fact that you’ve done different cells for each thing instead of “one cell to rule them all”.

bstien:
> In order for snapshots to work properly we might need to either have all the cells in one table, or have one Demo per each cell type. What do you think we should do, Bastian?

We may add a couple of more cells, which could be customizable and have different variants. So to both show these off and have the snapshot tests succeed I think it's best to create a separate demo for each cell type.

3lvis:
I’ve been thinking that might be worth to place this components under Recyclable instead of Components (with all the file structure that involves adding this to Recyclable). How does that sound to you?

3lvis:
Let’s do one demo for each then :)

Sources/Components/TableViewCells/Demo/TableViewCellsDemoView.swift
bstien: Protocols are cheap, and I wanted a single place to change/test how much feedback a selection should produce.

Sources/Components/TableViewCells/BasicTableViewCell.swift
bstien: Removed 👌

Sources/Components/TableViewCells/Demo/TableViewCellsDemoView.swift
bstien: Done 🍰

bstien:
I was unsure where to put all of this, but `Recycling/ListViews/Cells` may be a good place :)

ag12:
Nice job @bstien 👏. 
How are we going to be able to change customization on the cell? Let's say I want to use a different font/text size/color etc? :D 
Do we a Base Theme, default Theme? etc. Or is it possible to add Themes? Thanks

3lvis:
Maybe Recycling/ListViews/Generic/Cells?

bstien:
> Maybe Recycling/ListViews/Generic/Cells?

Even better 👏

bstien:
> Nice job @bstien 👏.
> How are we going to be able to change customization on the cell? Let's say I want to use a different font/text size/color etc? :D
> Do we a Base Theme, default Theme? etc. Or is it possible to add Themes? Thanks

I'm not really sure if we need a style-engine-thingy for these. For most usecases the font, size and colors will be the same. I'm thinking that it may be easier to just subclass one of the generic ones and modify it as needed. If we at some point notice that we are aaaalways subclassing these and modifying, then we may consider either the need to add some more cell types to FinniversKit or make them `even more modifiable`™.

The main point is, as I see it, to create a basis for cells we use throughout the app :)

bstien:
The tests breaks because we've replaced the checkbox image with a 24x24 image, instead of 20x20. I'll look into these changes early next week.

PR #264, opened 2018-10-26
Changed icons for radiobuttons and checkboxes

bstien:
# What?
Icons for radiobuttons and checkboxes should have size 24x24.
This fits into Thuy's design where elements and spacing should be dividable by 8 -> `8, 16, 24, 32, ... n*8`


3lvis:
@bstien You might need to refresh the screenshots

bstien:
Ah, forgot that😣 Updated now 👍

PR #265, opened 2018-10-26
Refactor/selectionboxes and animated imageview

bstien:
# Why?
Radiobuttons and checkboxes were pretty hard to use before outside of a `SelectionBox` before. Now there are two separate classes for this, which can potentially be used anywhere it's needed without much customization.

# What?
- Added `AnimatedRadiobuttonView` and `AnimatedCheckboxView`
- Consistent spacing between elements (8,16,24,32, ... n*8)
- Updated snapshot tests

# Show me
### Checkbox (before -> after)

checkbox_gammel
checkbox_ny

### Radiobutton (before -> after)
radiobutton_gammel
radiobutton_ny

### Review (before -> after)
review_gammel
review_ny

### Report (before -> after)
report_gammel
report_ny


Sources/Components/SelectionBoxes/Selectionbox.swift
```diff
@@ -183,46 +156,28 @@ extension Selectionbox {
             titleLabel.leadingAnchor.constraint(equalTo: leadingAnchor, constant: .mediumLargeSpacing),
             titleLabel.topAnchor.constraint(equalTo: topAnchor, constant: .mediumLargeSpacing),
 
-            stack.leadingAnchor.constraint(equalTo: leadingAnchor, constant: .mediumLargeSpacing),
+            //            stack.heightAnchor.constraint(equalToConstant: CGFloat(fields.count * 48)),
```
3lvis: Can we remove this commented code?

bstien: Removed 💥

PR #269, opened 2018-11-07
Fix/bundle for animated image

bstien:
# Why?
Assets were loaded from the main bundle, instead of `FinniversKit`.


Sources/Components/SelectionBoxes/AnimatedCheckboxView.swift
```diff
@@ -30,10 +30,18 @@ public class AnimatedCheckboxView: AnimatedSelectionView {
     }
 
     private func setImages() {
-        let framesPerSecond = AnimatedSelectionView.framesPerSecond
-        guard let selectedImages = UIImage.animatedImageNamed("checkbox-selected-", duration: 20 / framesPerSecond)?.images,
-            let unselectedImages = UIImage.animatedImageNamed("checkbox-unselected-", duration: 14 / framesPerSecond)?.images else {
-                fatalError("Could not instantiate animation images")
+        var selectedImages = [UIImage]()
+        for index in 0 ..< 20 {
+            if let image = UIImage(named: "checkbox-selected-\(index)", in: FinniversKit.bundle, compatibleWith: nil) {
```
osanoj: It would maybe be nicer to have these defined in a `[FinniversImageAsset]`, so we can still use our rename-proof way of accessing resource images

3lvis: I don't think we can since that's autogenerated. We should still maybe extract into a variable.

osanoj: Ah, wasn't aware these images were not part of the assets catalog that generates `FinniversImageAsset`

3lvis:
Did you check if this fixes the FINN app?

bstien:
Yes, it's not crashing anymore and the icons are loaded correctly.

3lvis:
👏 👏 👏 👏 👏

PR #271, opened 2018-11-12
Add sold page.

Torsph:
This will be used when you click the sold marking on the AdManagmentPage

# Why?

There should be a clear indication that the item has been set to sold when you click the `Mark as sold` text.

# Show me

screenshot 2018-11-12 at 10 41 01


3lvis:
Checking with Thuy since we have a similar UI when reporting an ad and we should have some sort of consistency.

3lvis:
screenshot 2018-11-12 at 12 58 00

thuygianguyen:
I've adjusted spacing and font sizes that conform with the Finnivers and made this

image

zpl://project?pid=581858889668fdfc012811c8&seid=5bf3d123ccaa613e867a5bf0

3lvis:
@Torsph Can you have a second look following the design change?

3lvis:
We should make this the "end of flow" screen and have the different alternatives since this can be reused in many places where the flow ends.

Torsph:
Sorry, had some vacation.

The first page is inside the trust sdk. The other two looks ok to me.

Sources/Fullscreen/SoldView/SoldView.swift
```diff
@@ -0,0 +1,88 @@
...
+    }()
+
+    private lazy var stack: UIStackView = {
+        let stack = UIStackView(arrangedSubviews: [
+                imageView,
+                title,
+                button
+            ])
+        stack.translatesAutoresizingMaskIntoConstraints = false
+        stack.spacing = 24
```
Torsph: Should there be added an option to send a multiplier to the Spacing class? I cannot find a spacing option of the 24pt.

3lvis: That would be cool! I don't remember doing multipliers, actually I don't even know how to use them. Feel free to add support for it and show us how it's used :)

ag12:
Looks like @Torsph have implemented changes @thuygianguyen requested. 👏

PR #278, opened 2018-12-04
Adds files from FinnLabs project, plus some refactor to make it work ok

shredlocker:
# Why?

We need a bottom sheet implementation in FK and this one is a good starting point.

# What?

- Bottom sheet view controller for presenting view controllers easily
- Custom spring animator specifically for bottom sheet animation
- All transition objects needed for modal presentation
- Gesture controller for interacting with transition and presentation

Using SafeLayout constraints is messing a bit with the bottom sheet presentation because content will not follow the above these safe margins. 

# Show me

bottom-sheet-animation

### DIFF
--- a/Demo/Demo.swift
+++ b/Demo/Demo.swift
@@ -1,7 +1,7 @@
 //
 //  Copyright © FINN.no AS, Inc. All rights reserved.
 //
-import UIKit
+import FinniversKit
 
 public enum DnaViews: String, CaseIterable {
     case color
@@ -38,6 +38,7 @@ public enum ComponentViews: String, CaseIterable {
     case loadingIndicator
     case horizontalSlide
     case easterEggButton
+    case bottomSheet
 
     public var viewController: UIViewController {
         switch self {
@@ -73,6 +74,9 @@ public enum ComponentViews: String, CaseIterable {
             return secondViewController
         case .easterEggButton:
             return DemoViewController<EasterEggButtonDemoView>()
+        case .bottomSheet:
+            let controller = DemoViewController<ButtonDemoView>()
+            return BottomSheet(rootViewController: controller)
         }
     }
 }
--- a/Sources/Components/BottomSheet/Animator/SpringAnimator.swift
+++ b/Sources/Components/BottomSheet/Animator/SpringAnimator.swift
@@ -1,0 +1,104 @@
+//
+//  SpringAnimator.swift
+//  spring-physics
+//
+//  Created by Granheim Brustad , Henrik on 27/10/2018.
+//  Copyright © 2018 Henrik Brustad. All rights reserved.
+//
+
+import UIKit
+
+extension SpringAnimator {
+    enum State {
+        case animating, paused, cancelled, stopped
+    }
+}
+
+class SpringAnimator: NSObject {
+
+    // Spring properties
+    let damping: CGFloat
+    let stiffness: CGFloat
+
+    // View properties
+    private var velocity = 0.0 as CGFloat
+    private var position = 0.0 as CGFloat
+    var initialVelocity: CGFloat = 0 {
+        didSet { velocity = -initialVelocity }
+    }
+
+    // Animation properties
+    var state: State = .stopped
+
+    var targetPosition = 0 as CGFloat
+    var constraint: NSLayoutConstraint?
+
+    var completion: ((Bool) -> Void)?
+
+    private let scale = 1 / UIScreen.main.scale
+    private var displayLink: CADisplayLink?
+
+    init(dampingRatio: CGFloat, frequencyResponse: CGFloat) {
+        self.stiffness = pow(2 * .pi / frequencyResponse, 2)
+        self.damping = 2 * dampingRatio * sqrt(stiffness)
+    }
+
+    func startAnimation() {
+        if state == .paused { continueAnimation() }
+
+        guard let constraint = constraint else { return }
+        position = targetPosition - constraint.constant
+
+        guard position != 0, displayLink == nil else { return }
+        displayLink = CADisplayLink(target: self, selector: #selector(step(displayLink:)))
+        displayLink?.add(to: .current, forMode: .default)
+        state = .animating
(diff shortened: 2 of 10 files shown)


Demo/Demo.swift
```diff
@@ -73,6 +74,9 @@ public enum ComponentViews: String, CaseIterable {
             return secondViewController
         case .easterEggButton:
             return DemoViewController<EasterEggButtonDemoView>()
+        case .bottomSheet:
+            let controller = DemoViewController<ButtonDemoView>()
```
vadymmarkov: Not sure if it's a problem or not, but does it play well with double tap gestures we're using to dismiss demo controller? Maybe it's better to have a view with "present" button like you did in FinnLabs project.

Sources/Components/BottomSheet/Animator/SpringAnimator.swift
```diff
@@ -0,0 +1,100 @@
...
+extension SpringAnimator {
+    enum State {
+        case animating, paused, cancelled, stopped
+    }
+}
+
+class SpringAnimator: NSObject {
+
+    // Spring properties
+    let damping: CGFloat
```
vadymmarkov: Could these properties be private?

Sources/Components/BottomSheet/Animator/SpringAnimator.swift
```diff
@@ -0,0 +1,100 @@
...
+    let stiffness: CGFloat
+
+    // View properties
+    private var velocity = 0.0 as CGFloat
```
vadymmarkov: Nitpicking here, but I prefer `private var velocity: CGFloat = 0`

Sources/Components/BottomSheet/Animator/SpringAnimator.swift
```diff
@@ -0,0 +1,100 @@
...
+        position += velocity * CGFloat(displayLink.duration)
+        constraint?.constant = targetPosition - position
+
+        if abs(position) < scale, abs(velocity) < scale {
+            stopAnimation(didComplete: true)
+        }
+    }
+
+    func stopAnimation(didComplete: Bool) {
+        if didComplete { constraint?.constant = targetPosition }
```
vadymmarkov: Not a big fan of one line if statements 😄

Sources/Components/BottomSheet/Controller/BottomSheet.swift
```diff
@@ -0,0 +1,86 @@
...
+//  Copyright © 2018 FINN.no. All rights reserved.
+//
+
+import UIKit
+
+public class BottomSheet: UIViewController {
+
+    // MARK: - Public properties
+
+    public var rootViewController: UIViewController
```
vadymmarkov: I think it should be private or at least let because if you set new root view controller from outside it's gonna break things.

Sources/Components/BottomSheet/Animator/SpringAnimator.swift
shredlocker: Yes, these should be private

Sources/Components/BottomSheet/Animator/SpringAnimator.swift
shredlocker: Hehe, I can change it

Sources/Components/BottomSheet/Controller/BottomSheet.swift
```diff
@@ -0,0 +1,86 @@
...
+        }
+        view.addSubview(notch)
+
+        addChild(rootViewController)
+        view.insertSubview(rootViewController.view, belowSubview: notch)
+        rootViewController.didMove(toParent: self)
+        rootViewController.view.translatesAutoresizingMaskIntoConstraints = false
+
+        NSLayoutConstraint.activate([
+            notch.centerXAnchor.constraint(equalTo: view.centerXAnchor),
```
vadymmarkov: Should we use default spacings from `CGFloat` extension we have?

Sources/Components/BottomSheet/Controller/BottomSheet.swift
shredlocker: We can make it private for now and then change things as we need

Sources/Components/BottomSheet/Controller/BottomSheet.swift
shredlocker: Maybe ask Thuy? I've just used some random values that I thought looked ok

Sources/Components/BottomSheet/Animator/SpringAnimator.swift
shredlocker: Things were so great in C and Java...

Sources/Components/BottomSheet/Transition/BottomSheetGestureController.swift
```diff
@@ -0,0 +1,82 @@
+//
+//  Copyright © 2018 FINN.no. All rights reserved.
+//
+
+import UIKit
+
+protocol BottomSheetGestureControllerDelegate: class {
+    func gestureDidBegin() -> CGFloat // Expects to get the current position of the bottom sheet
```
vadymmarkov: We have a convention for delegate methods to be in format like 
```swift
func bottomSheetGestureControllerDidBeginGesture(_ controller: BottomSheetGestureController)
```

Demo/Demo.swift
shredlocker: It doesn't seem to be a problem at the moment. It will just dismiss if you double tap inside the bottom sheet. Can of course consider having a separate demo view controller for the bottom sheet

Sources/Components/BottomSheet/Transition/BottomSheetGestureController.swift
shredlocker: I'm too lazy 😊

shredlocker: I'll fix though!

PR #280, opened 2018-12-07
Use state controller in bottom sheet

shredlocker:
# Why?

The presentation and transition are sharing the state of the bottom sheet. Having a separate object handling the state and defining sizes for each state makes sense.

# What?

- Added state controller to handle states
- Simplified gesture controller to simply handle gestures
- Tweaked transition animation, no overshoot on present and dismiss, only when the user pans the bottom sheet.

# Show me

No UI

### DIFF
--- a/Sources/Components/BottomSheet/Transition/BottomSheetGestureController.swift
+++ b/Sources/Components/BottomSheet/Transition/BottomSheetGestureController.swift
@@ -7,23 +7,21 @@ import UIKit
 protocol BottomSheetGestureControllerDelegate: class {
     // Expects to get the current position of the bottom sheet
     func bottomSheetGestureControllerDidBeginGesture(_ controller: BottomSheetGestureController) -> CGFloat
-    func bottomSheetGestureController(_ controller: BottomSheetGestureController, didChangeGesture position: CGFloat)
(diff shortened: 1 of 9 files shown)


Sources/Components/BottomSheet/Transition/BottomSheetGestureController.swift
```diff
@@ -34,7 +32,8 @@ class BottomSheetGestureController {
 
 private extension BottomSheetGestureController {
     @objc func handlePan(gesture: UIPanGestureRecognizer) {
-        let transition = gesture.translation(in: containerView)
+        let translation = gesture.translation(in: containerView)
+        self.translation = translation.y
```
vadymmarkov: Should it perhaps be called `translationY`?

Sources/Components/BottomSheet/Transition/BottomSheetStateController.swift
```diff
@@ -0,0 +1,57 @@
...
+
+    var state: State = .compressed
+    var frame: CGRect = .zero
+    var targetPosition: CGFloat {
+        get {
+            return targetPosition(for: state)
+        }
+    }
+    
+    private var minValue = 44 as CGFloat
```
vadymmarkov: This `as CGFloat` again 😄

PR #283, opened 2018-12-12
Christmas themed EmptyView

bstien:
# Why?
_FINN users_: We want to throw christmas-themed shapes around when we're not logged in or when we don't have a network connection!

# When do we want it?
_FINN users_: **NOW🎉🎄🎁**

# Show me
### Before

Nothing, empty.

### After
screen shot 2018-12-12 at 17 58 10


ag12:
How did you fix the triangle challenge? :D

bstien:
The triangle is an image this time. I copied the `collisionBoundingPath` from the existing `TriangleView`.

Looks good on iPad! 👌🎉

Sources/Fullscreen/EmptyView/ChristmasViews/ChristmasOrnamentView.swift
```diff
@@ -0,0 +1,46 @@
...
+    case red, blue
+}
+
+class ChristmasOrnamentView: UIImageView, AttachableView {
+    var attach: UIAttachmentBehavior?
+    private let ornamentColor: ChristmasOrnamentColor
+
+    // MARK: - Init
+
+    public init(frame: CGRect, ornamentColor: ChristmasOrnamentColor) {
```
vadymmarkov: I guess it shouldn't be public since class itself is not.

Sources/Fullscreen/EmptyView/ChristmasViews/RectangleGiftView.swift
```diff
@@ -0,0 +1,37 @@
...
+import UIKit
+
+class RectangleGiftView: UIImageView, AttachableView {
+    var attach: UIAttachmentBehavior?
+    private let cornerRadius: CGFloat = 4.0
+    private let imageAsset: FinniversImageAsset
+
+    // MARK: - Init
+
+    public init(frame: CGRect, image: FinniversImageAsset) {
```
vadymmarkov: Same comment about public in this file.

Sources/Fullscreen/EmptyView/ChristmasViews/RoundedRectangleGiftView.swift
```diff
@@ -0,0 +1,53 @@
...
+//
+
+import UIKit
+
+class RoundedRectangleGiftView: UIImageView, AttachableView {
+    var attach: UIAttachmentBehavior?
+
+    // MARK: - Init
+
+    public override init(frame: CGRect) {
```
vadymmarkov: Same comment about public in this file.

Sources/Fullscreen/EmptyView/ChristmasViews/StarView.swift
```diff
@@ -0,0 +1,85 @@
...
+//
+
+import UIKit
+
+class StarView: UIImageView, AttachableView {
+    var attach: UIAttachmentBehavior?
+
+    // MARK: - Init
+
+    public override init(frame: CGRect) {
```
vadymmarkov: Same comment about public in this file.

Sources/Fullscreen/EmptyView/ChristmasViews/TriangleGiftView.swift
```diff
@@ -0,0 +1,58 @@
...
+//
+
+import UIKit
+
+class TriangleGiftView: UIImageView, AttachableView {
+    var attach: UIAttachmentBehavior?
+
+    // MARK: - Init
+
+    public override init(frame: CGRect) {
```
vadymmarkov: Same comment about public in this file.

Sources/Fullscreen/EmptyView/EmptyView.swift
```diff
@@ -20,6 +21,9 @@ public class EmptyView: UIView {
     private let sizeOfSquare = CGSize(width: 50, height: 50)
 
     private var hasLayedOut = false
+    private let isChristmasThemed: Bool
```
vadymmarkov: I would probably make a separate view or something like this, but this flag should work for now

bstien: Yes, I thought of making a separate view, but that means more work in the iOS app just to remove it when christmas is over.

PR #286, opened 2018-12-14
Updated Fronpage Favorite Button

ag12:
# Why?
New design.

# What?
New Frontpage Favorite Button.

### Before
before

### After
after


bstien:
Approved as long as the tests run green 💚

PR #287, opened 2018-12-17
Set bottom sheet height on creation

shredlocker:
# Why?

One should be able to set the height of the bottom sheet in both compact and expanded modes.

# What?

- Height struct for setting the height of the bottom sheet in both modes.
- Default value is the filter size
- Use navigation controller in bottom sheet demo

# Show me

No UI


Sources/Components/BottomSheet/Controller/BottomSheet.swift
```diff
@@ -4,12 +4,33 @@
...
+        public let expanded: CGFloat
+
+        public static var defaultFilterHeight: Height {
+            let screenSize = UIScreen.main.bounds.size
+            let compact: CGFloat = screenSize.height >= 812 ? 570 : 510
+            let expanded = screenSize.height - 64
+            return Height(compact: compact, expanded: expanded)
+        }
+
+        public static var zero = Height(compact: 0, expanded: 0)
```
vadymmarkov: Should it be `let`?

shredlocker: Yes!

shredlocker:
It should fix the warnings

PR #294, opened 2018-12-19
Adds dim view with basic animation, move setup code to separate method

shredlocker:
# Why?

Background should be dimmed to focus user attention on the bottom sheet.

# What?

Adds a dim view which fades in on presentation and fades out on dismissal. Very basic.
Bottom sheet will dismiss if tapping on in this view.

# Show me

simulator screen shot - iphone xs - 2018-12-19 at 09 31 00

### DIFF
--- a/Sources/Components/BottomSheet/Transition/BottomSheetPresentationController.swift
+++ b/Sources/Components/BottomSheet/Transition/BottomSheetPresentationController.swift
@@ -28,7 +28,7 @@ class BottomSheetPresentationController: UIPresentationController {
     }
 
     private lazy var dimView: UIView = {
-        let view = UIView(frame: UIScreen.main.bounds)
+        let view = UIView(frame: .zero)
         view.backgroundColor = UIColor.black.withAlphaComponent(0.4)
         view.alpha = 0
         return view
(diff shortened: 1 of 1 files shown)


Sources/Components/BottomSheet/Transition/BottomSheetPresentationController.swift
```diff
@@ -27,6 +27,13 @@ class BottomSheetPresentationController: UIPresentationController {
         return .overCurrentContext
     }
 
+    private lazy var dimView: UIView = {
+        let view = UIView(frame: UIScreen.main.bounds)
```
vadymmarkov: Can we use auto layout instead? Because I think it's gonna break if you rotate to landscape on iPad, for example.

shredlocker: Sure

osanoj:
Would it somehow be possible to let the user tap where the status bar is to initiate scroll to top (If this would actually work when the scrollview is inside a bottomsheet)? Not required to fix this in this PR, but it would be awesome if we could make the scroll to top shortcut work for our bottomsheets.

shredlocker:
I think that would be up to the view controller which is displayed in the bottom sheet.

PR #296, opened 2018-12-19
Adds refactored spring animator class

shredlocker:
# Why?

The current spring animation is too specific toward the bottom sheet position. This PR separates the spring animation from the constraint used to animate the bottom sheet. The spring animation will now calculate the position of the spring and call any animation blocks added to the animation.

Current way:
`springAnimator.constraint = bottomSheet.constraint`

New way:
`springAnimator.addAnimation { position in
          bottomSheet.constraint.constant = position.y
}`

This allows us to animate more than just the bottom sheet position in the transition. We could sync the animation of the dimView alpha with the animation of the bottom sheet position.

# What?

- Refactored SpringAnimator class. It is like the UIViewAnimating protocol, except I didn't use that protocol because some of it need iOS 10.0 and up.
- Tried to make it look more like UIView.animate API
- Added a bunch of CGPoint operator function for easier implementation

# Show me

Animate both bottom sheet position and bottom sheet background color during transition:

bottom-sheet-color


shredlocker:
That was quick!

PR #304, opened 2019-01-03
Add custom refresh control

vadymmarkov:
# Why?

Because we want to use branded refresh control instead of the default one.

# What?

- Implement `RefreshControl` component based on existing `LoadingIndicatorView` component
- Use new refresh control in `AdsGridView` on the front page
- Slightly modified `LoadingIndicatorView` because the start angle of layer's bezier path needs to be `3π/2` for refresh control. It doesn't make a big difference in loading animation though.

It's super easy to use `RefreshControl` since it's is a subclass of `UIRefreshControl`:

```swift
if #available(iOS 10.0, *) {
    tableView.refreshControl = refreshControl
} else {
    tableView.addSubview(refreshControl)
}
```

The only difference is `RefreshControlDelegate` that should be implemented in order to be notified when refresh control begins refreshing.

# Show me

refresh_control


ag12:
Impressive clean and nice coding here 😍

vadymmarkov:
It's 5 seconds in Demo just to show loading animation. Normally you need to call `reloadData` to stop refreshing.

PR #306, opened 2019-01-04
Update/icons

ag12:
# Why?
Part of the new design icons for 2019.

# What?
New icons.


ag12:
Thanks! We just have to wait before we merge here and in the iOS app. Have to sync it 😅

PR #308, opened 2019-01-09
Bottom sheet limit gesture

shredlocker:
# Why?

Users should not be able to pan the bottom sheet higher than the expanded height.

# What?

- As the bottom sheet reaches its maximum height it will smoothly animate to the maximum height
- Once it has reached the maximum height the user won't be able to expand it any further

Other fixes:
- Fix dismissing the BottomSheetMechanics demo
- Fix dimming the background when canceling the presentation transition
- Change the parameters of the spring animation during the transition once the user starts interacting with the bottom sheet. This makes it feel more consistent when interacting with it during transition part.

# Show me

Wasn't able to make a recording showing where my finger was and it won't be very clear without that.


ag12:
Awesome job dude!

PR #311, opened 2019-01-16
Moved BottomSheetMechanicsDemoViewController into Demo-folder

bstien:
# Why?
One of the demo-files was placed within `Sources` instead of `Demo`.
This was causing an issue when pulling the project through Cocoapods.

# What?
- Moved `BottomSheetMechanicsDemoViewController` into `Demo`-folder.


ag12:
cool!

PR #314, opened 2019-01-22
Bottom sheet optional compact height

shredlocker:
# Why?

You might not always want to expand the bottom sheet.

# What?

- Made compact height optional. Setting this to nil makes the bottom sheet expanded by default.
- By adjusting the expanded height you will get the effect that you cannot expand the bottom sheet.

### DIFF
--- a/Sources/Components/BottomSheet/Controller/BottomSheet.swift
+++ b/Sources/Components/BottomSheet/Controller/BottomSheet.swift
@@ -6,19 +6,23 @@ import UIKit
 
 extension BottomSheet {
     public struct Height {
-        public let compact: CGFloat
+        public let compact: CGFloat?
         public let expanded: CGFloat
 
         public static var defaultFilterHeight: Height {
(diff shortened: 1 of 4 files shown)


Sources/Components/BottomSheet/Controller/BottomSheet.swift
```diff
@@ -6,19 +6,23 @@ import UIKit
 
 extension BottomSheet {
     public struct Height {
-        public let compact: CGFloat
+        public let compact: CGFloat?
```
vadymmarkov: Not sure if it's actually better, but I think we could achieve the same by making `compact` equal to `expanded`. Then you can use something like `isExpandedByDefault: Bool { return compact == expanded }` when needed. What do you think?

shredlocker: Good idea! It would be nice to avoid the optional!

PR #319, opened 2019-01-30
 Add accessory label to AdsGridViewCell

vadymmarkov:
# Why?

In order to show more information below the title, for example total price.

# What?

Add optional accessory label to `AdsGridViewCell`

# Show me

### Before

simulator screen shot - iphone x - 2019-01-30 at 17 02 21

### After

simulator screen shot - iphone x - 2019-01-30 at 17 04 35


Sources/Recycling/GridViews/Ads/Cell/AdsGridViewCell.swift
```diff
@@ -147,7 +162,11 @@ public class AdsGridViewCell: UICollectionViewCell {
             titleLabel.leadingAnchor.constraint(equalTo: leadingAnchor),
             titleLabel.trailingAnchor.constraint(equalTo: trailingAnchor),
             titleLabel.heightAnchor.constraint(equalToConstant: AdsGridViewCell.titleHeight),
-            titleLabel.bottomAnchor.constraint(equalTo: bottomAnchor, constant: -AdsGridViewCell.titleBottomMargin),
+
+            accessoryLabel.topAnchor.constraint(equalTo: titleLabel.bottomAnchor),
+            accessoryLabel.leadingAnchor.constraint(equalTo: leadingAnchor),
+            accessoryLabel.trailingAnchor.constraint(equalTo: trailingAnchor),
+            accessoryLabel.bottomAnchor.constraint(equalTo: bottomAnchor, constant: -AdsGridViewCell.bottomMargin),
```
bstien: Should the `accessoryLabel` also have its `heightAnchor` constrained? I see the others have it.

vadymmarkov: No, because this label is optional in contrast with others, so I want the height to be zero when there is no text set.

bstien: Ah, of course! Don't mind me😎

PR #322, opened 2019-02-01
Refactor/bottom sheet draggable area

bstien:
# Why?
Ref. the discussion in [this `charcoal-ios` PR](https://github.com/finn-no/charcoal-ios/pull/132), we should be able to choose which part of the bottomSheet that should be draggable.

These changes lets you choose between these three cases when creating a `BottomSheet`:
- The whole view
- Only navigationBar
- A custom `CGRect` within the bounds of the view

The next step is to remove the magic number `20`, which is the height of the top area that includes the notch. I'll make those changes in another PR soon.

# What?
- Added enum `BottomSheet.DraggableArea`.
- Added custom implementation of `UIPanGestureRecognizer`.

# Show me
### Before
draggable_everything

### After (navBar & custom)
draggable_navbar
draggable_custom


Demo/Components/BottomSheet/BottomSheetMechanicsDemoViewController.swift
```diff
@@ -12,7 +12,14 @@ protocol RootViewControllerDelegate: class {
 
 class RootViewController: UIViewController {
 
+    // MARK: - Public properties
+
     weak var delegate: RootViewControllerDelegate?
+    lazy var draggableLabelFrame = CGRect(origin: CGPoint(x: 0, y: 150), size: CGSize(width: view.frame.width, height: 44))
```
vadymmarkov: Could it be that `view.frame` is not set yet when we access `draggableLabelFrame`?

bstien: I'm not completely sure, but I think it always will be as long as it happens after we've called `init`.

This is from the documentation for [`UIViewController.loadView()`](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621454-loadview):
> The view controller calls this method when its view property is requested but is currently nil. This method loads or creates a view and assigns it to the view property.

shredlocker: It will be set after init, but it might not be the size we want. I guess we can only be sure the frame is correct after layoutSubviews is called when using Auto Layout.

vadymmarkov: Yeah, what @shredlocker said. By "not set" I meant zero or not correct frame that you might get in `viewDidLoad` for example. How about making `draggableLabelFrame` a computed var, then at least you always get the current view width:

```swift
var draggableLabelFrame: CGRect { 
    return CGRect(
        origin: CGPoint(x: 0, y: 150), 
        size: CGSize(width: view.frame.width, height: 44)
    )
}
```

bstien: Yeah, that makes sense 👍 I'll change it

PR #323, opened 2019-02-01
Pull the notch into a separate view

bstien:
# Why?
The bottomSheet's notch was created within `BottomSheet`, it should be in a separate view.

# What?
- Create private class `Notch` within `BottomSheet` file.

# Show me
No UI changes.


vadymmarkov:
Why should it be a separate view?

bstien:
Yeah, I should've made an effort to explain that this is related to #322 😅

I want to place this within a separate view so we can easily set the notch's height and constrain the `rootViewController` to the bottom of this view.

That way we don't have to have the magic number `20`™ around, as mentioned in the other PR.

PR #324, opened 2019-02-01
Add infobox component

vadymmarkov:
# Why?

Because we want to use it in the app.

# What?

- Add new `InfoboxView` component
- Make it possible to make a small version of every button style we have. Currently it's for `InfoboxView` only, but we'll start using it more in the future.

# Show me

simulator screen shot - iphone xs - 2019-02-01 at 15 27 42

simulator screen shot - iphone xs - 2019-02-01 at 15 43 41


Sources/Components/Infobox/InfoboxView.swift
```diff
@@ -0,0 +1,113 @@
...
+    required init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+        setup()
+    }
+
+    // MARK: - Overrides
+
+    public override func layoutSubviews() {
+        super.layoutSubviews()
+        layer.cornerRadius = 8
```
shredlocker: You could put this in `setup()`, right?

vadymmarkov: Absolutely!

PR #325, opened 2019-02-04
Add ConfirmationView to Demo

3lvis:
# Why?

Because it was missing from the Demos.

# What?

Adds ConfirmationView to Demo.

# Show me

simulator screen shot - iphone xs - 2019-02-04 at 14 48 14


3lvis:
🎉

PR #333, opened 2019-02-08
Fix/bottomsheet dismissal delegate

bstien:
# Why?
We need a delegate callback when `BottomSheet` is dismissed.

# What?
- Added protocols `BottomSheetDelegate` _(public)_ and `BottomSheetDismissalDelegate` _(internal)_.
- Call `BottomSheetDelegate.bottomSheetDidDismiss(_:)` when state is set to `.dismissed`, when the swipe gesture dismisses bottomSheet, or a tap happens within `dimView`.

### DIFF
--- a/Sources/Components/BottomSheet/Controller/BottomSheet.swift
+++ b/Sources/Components/BottomSheet/Controller/BottomSheet.swift
@@ -103,6 +103,7 @@ public class BottomSheet: UIViewController {
             self.bottomSafeAreaInset = 0
         }
         super.init(nibName: nil, bundle: nil)
+        transitionDelegate.dismissalDelegate = self
         transitioningDelegate = transitionDelegate
         modalPresentationStyle = .custom
     }
--- a/Sources/Components/BottomSheet/Controller/BottomSheet.swift
+++ b/Sources/Components/BottomSheet/Controller/BottomSheet.swift
@@ -4,6 +4,14 @@
 
 import UIKit
 
+public protocol BottomSheetDelegate: class {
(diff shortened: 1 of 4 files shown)


Sources/Components/BottomSheet/Transition/BottomSheetPresentationController.swift
```diff
@@ -125,15 +126,18 @@ private extension BottomSheetPresentationController {
     }
 
     @objc func handleTap() {
+        guard stateController.state != .dismissed else { return }
         stateController.state = .dismissed
         gestureController?.velocity = .zero
         presentedViewController.dismiss(animated: true)
+        dismissalDelegate?.bottomSheetDidDismiss()
```
shredlocker: Should this be called in the completion handler in the dismiss function instead maybe?

Sources/Components/BottomSheet/Controller/BottomSheet.swift
```diff
@@ -4,6 +4,14 @@
 
 import UIKit
 
+public protocol BottomSheetDelegate: class {
+    func bottomSheetDidDismiss(_ bottomSheet: BottomSheet)
+}
+
+protocol BottomSheetDismissalDelegate: class {
```
vadymmarkov: How about this instead:

```swift
protocol BottomSheetPresentationControllerDelegate {
    func bottomSheetPresentationController(
        _ presentationController: BottomSheetPresentationController, 
       didDismissPresentedViewController presentedViewController: UIViewController
    )
}
```

bstien: Even better👍🎉

PR #341, opened 2019-02-26
Feedbackview, aka. RatingCat

bstien:
# Why?
We want to make RatingCat into a reusable component. It should also need to handle a set of questions based on our need for feedback, instead of hardcoded strings which we use today.

`FeedbackView` can present a title and one, two or zero buttons. It has 4 states, which can be used by the handling view/viewController to determine which text or buttons to show next based on which button the user selected. The states are: `initial`, `accept`, `decline` and `finished`.

## Configuring
It takes an instance of `FeedbackViewModel` to configure its views. If `nil` is provided to either `positiveButtonTitle` or `negativeButtonTitle`, that button won't be presented. Only `title` is required.

```swift
struct FeedbackViewModel {
    let title: String
    let positiveButtonTitle: String?
    let negativeButtonTitle: String?
}
```

## Handle selections
When the user taps a button, `FeedbackView` notifies its delegate with the current state and the button that was tapped, so the next state can be configured.

```swift
public func feedbackView(_ feedbackView: FeedbackView, didSelectButtonOfType buttonType: FeedbackView.ButtonType, forState state: FeedbackView.State) {
        let newState: FeedbackView.State
        switch (state, buttonType) {
        case (.initial, .positive):
            newState = .accept
        case (.initial, .negative):
            newState = .decline
        case (_, _):
            newState = .finished
        }

        let viewModel = // Fetch next viewModel for newState.
        feedbackView.setState(newState, withViewModel: viewModel)
    }
```

# Show me
RatingCat should work well in both a list and grid, so I added the button at the top to switch between the two, as would happen within the search results in the ios-app.

ratingcat
screen shot 2019-02-26 at 10 22 22
screen shot 2019-02-26 at 10 23 05


Demo/Components/Feedback/FeedbackDemoView.swift
```diff
@@ -0,0 +1,145 @@
...
+
+    private var hasSwitchedViews = false
+
+    private let gridItemHeight: CGFloat = 300
+    private let listItemHeight: CGFloat = 100
+
+    lazy var listItems = [singleQuestionFeedback.createFeedbackView(), multiQuestionFeedback.createFeedbackView()]
+    lazy var gridItems = [singleQuestionFeedback.createFeedbackView(), multiQuestionFeedback.createFeedbackView()]
+
+    lazy var listFeedbackStackView: UIStackView = {
```
vadymmarkov: Should stack views be private?

Sources/Components/Feedback/FeedbackView.swift
```diff
@@ -0,0 +1,341 @@
...
+
+            titleView.transform = CGAffineTransform(translationX: 0, y: 30)
+            titleView.alpha = 0
+            titleView.setTitle(viewModel.title)
+            UIView.animate(withDuration: 0.2, animations: { [weak self] in
+                self?.titleView.transform = .identity
+                self?.titleView.alpha = 1
+                snapshotTitleView.alpha = 0
+                snapshotTitleView.transform = CGAffineTransform(translationX: 0, y: -30)
+
```
vadymmarkov: Extra line?

Sources/Components/Feedback/FeedbackViewModel.swift
```diff
@@ -0,0 +1,17 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import Foundation
+
+public struct FeedbackViewModel {
+    let title: String
```
vadymmarkov: Should it be `public`? Just for case.

PR #344, opened 2019-02-26
Add Earth Hour component

vadymmarkov:
# Why?

Because we want to run "Earth Hour" campaign in March.

# What?

- Add expandable `EarthHourView`.
- Animation should be started manually when the view is visible on the screen.  

# Show me

earth


ag12:
Du må bare søke jobb i `NASA` snart ™️

vadymmarkov:
@osanoj Demo should be fixed for iPhone4 and iPad.

PR #345, opened 2019-02-27
Happiness rating

bstien:
# Why?
We need a new component to show users when we want to ask them how happy they are with the app.

This view consists of 5 ratings/faces: love, happy, neutral, dissatisfied and angry.

The user may either tap each face or swipe over them to select the one they want. Haptic feedback are enabled for users with iOS 10, or later.

# Show me
happinessrating
screen shot 2019-02-27 at 13 48 11
screen shot 2019-02-27 at 13 48 13


ingridvf:
We just got complaints about the color contrast of the unselected smileys for Trust & Rep, so there I have chosen to invert the selected smiley instead. Could we do the same in this usecase? 

image

bstien:
Sure, are they in Zeplin?

ingridvf:
https://zpl.io/brl3yGL here is the trust project, sent you an invite. I guess @thuy could make the inverted ones for the feedback?

thuyranosaurus:
@ingridvf We should NOT do inverted faces/blackface for this interface. 
In our solution we do not lead you to the next page automatically like in your example. So you also get a clear CTA upon pressing. 
I also believe that with the scaling (15%) of the selected smiley, haptic feedback and also the darker blue colour for the selected we are doing plenty to clarify which icon is selected.

TL;DR: I will not make inverted faces

ingridvf:
In the video example above, the smileys turn grey when they are not selected, and this color does not have good enough contrast to the white background. So if a user with poor sight changes their mind, they cannot see the other smileys available, correct? Or does it not work like in the video above?

ingridvf:
It is not the selected icon that is the issue, it is the other ones @thuyranosaurus

thuyranosaurus:
contrast comparison

@ingridvf I think just testing out with a 60% Primary blue increases the contrast a lot.
We also shrink the the unselected slightly now to make the increase the contrast between selected and unselected.

ingridvf:
I think that could work @thuyranosaurus 👍 I'll do the same for feedback input for Trust & rep

PR #346, opened 2019-02-27
New fullscreen gallery

pimms:
~~**WIP!** There are a few visual final touches missing that I'll need to get sorted before this can be merged. I've held off opening a PR on this task until I've got a fully working integration with the FINN app running, as there are a lot of old moving parts I needed to take into account. They should all be sorted out now, and only the polish remains.~~

# Why?

The old image gallery has visual and functional issues that are hard to track down and harder to fix. The new implementation has has the same functionality, but with more playful animations and in general feels a lot smoother.

# What?

The new `FullscreenGalleryViewController` is created by providing a `FullscreenGalleryViewModel`, a `FullscreenGalleryImageSource`. The optional `FullscreenGalleryTransitionController` _(requires a `FullscreenGalleryTransitionPresenterDelegate`)_ makes the transition buttery smooth 🥞

### Two way communication

The way the `FullscreenGalleryViewModel` is used to communicate information about the selected image. The `FullscreenGalleryViewController` sets the `selectedIndex` attribute, which is being handled by `GalleryItem` and `AdPageGalleryViewModel` in the main app. I don't want to rewrite the logic behind this in the main app for this feature, but we **can** quite easily outsource the changing of the `selectedIndex` to `FINObjectViewController` through a delegate of `FullscreenGalleryViewController`.

Typing this out, I'm fairly certain that a delegate _(and setting the `selectedIndex` in `FINObjectViewController`)_ is the way to go, but a second opinion would be nice 🙂

### FullscreenGalleryMassiveViewController

The primary View Controller is almost certainly too large as-is. Pointers on where it can be divided are very welcome 🙌

# Show me

fullscreen-gallery

# Diagrams

The interactions on the FINN-side of things is difficult to explain, so I drew up a quick diagram of how this component is currently integrated with the app.

screenshot 2019-03-01 at 08 38 50


Demo/Fullscreen/FullscreenGallery/FullscreenGalleryDemoViewController.swift
```diff
@@ -0,0 +1,10 @@
+//
+// Copyright (c) 2019 FINN AS. All rights reserved.
+//
+
+class FullscreenGalleryDemoViewController: DemoViewController<FullscreenGalleryDemoView> {
```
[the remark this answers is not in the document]
pimms: > Is it because of limitations of the Demo project where we work mostly with views?

This was indeed the reasoning behind this, but moving this into the custom view-controller sounds like a much better solution 🙂 👍

Sources/Fullscreen/FullscreenGallery/FullscreenGalleryViewController.swift
```diff
@@ -0,0 +1,472 @@
...
+    public override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
+        super.viewWillTransition(to: size, with: coordinator)
+        previewView.superviewWillTransition(to: size)
+    }
+
+    // MARK: - View interactions
+
+    @objc private func dismissButtonTapped() {
+        galleryTransitioningController?.dismissVelocity = nil
+        dismiss(animated: true)
```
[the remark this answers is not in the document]
pimms: Sounds like a good idea 👍 This would also make the call to `DeviceOrientation.lockPhone()` from `FINObjectViewController` happen in a more natural place 🙂 It currently happens in `FullscreenGalleryTransitionPresenterDelegate::fullscreenGalleryTransitionOutCompleted`, which only gets called when using the **optional** transition controller.

Demo/Fullscreen/FullscreenGallery/FullscreenGalleryDemoViewController.swift
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+        setup()
+    }
+
+    private func setup() {
+        contentView.addSubview(imageView)
+        imageView.fillInSuperview(insets: UIEdgeInsets(all: .mediumSpacing))
```
bstien: Worth noting that `fillInSuperview(insets:)` adds constraints directly from the values from `UIEdgeInsets` to the corresponding anchors. I.e. it does not add a minus `-` to values for `trailingAnchor` or `bottomAnchor`, where you have set the value `.mediumSpacing`, aka. 8 points.

I guess what you really want is `-.mediumSpacing`, aka. -8 points.

Demo/Fullscreen/FullscreenGallery/FullscreenGalleryDemoViewController.swift
```diff
@@ -0,0 +1,304 @@
...
+            ImageDownloader.shared.simulatedDelayMs = 200
+        } else {
+            ImageDownloader.shared.simulatedDelayMs = 0
+        }
+    }
+
+    // MARK: - UICollectionView
+
+    public func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
+        return 5
```
bstien: Nitpicking, but could potentially be `min(imageUrls.count, imageCaptions.count)` instead.

FinniversKit.xcodeproj/project.pbxproj
```diff
@@ -2804,7 +2876,7 @@
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
 				CODE_SIGN_STYLE = Automatic;
-				DEVELOPMENT_TEAM = D34XZHQLE3;
+				DEVELOPMENT_TEAM = WSQ25TFDK3;
```
bstien: This should probably be changed back to its original value😃

FinniversKit.xcodeproj/project.pbxproj
```diff
@@ -2829,7 +2901,7 @@
...
```
bstien: And this one.

Sources/Fullscreen/FullscreenGallery/FullscreenGalleryViewController.swift
```diff
@@ -0,0 +1,508 @@
...
+    private var galleryTransitioningController: FullscreenGalleryTransitioningController? {
+        return transitioningDelegate as? FullscreenGalleryTransitioningController
+    }
+
+    // MARK: - UI properties
+
+    private lazy var backgroundView: UIView = {
+        let view = UIView(frame: .zero)
+        view.backgroundColor = .black
+        view.isUserInteractionEnabled = false
```
bstien: Not sure if you need this one, it's `false` by default.

Demo/Fullscreen/FullscreenGallery/FullscreenGalleryDemoViewController.swift
pimms: Hmmm, nice spot – is the issue in `fillInSuperview(insets:,isActive:)`? I would expect `UIEdgeInsets(leading: x, top: x, trailing: x, bottom: x)` to be equivalent to `x` pixels of padding, and the bottom & trailing constraints thus having `constant: -x`.

Sources/Fullscreen/FullscreenGallery/FullscreenGalleryViewController.swift
```diff
@@ -0,0 +1,508 @@
...
+        button.setImage(removeImage, for: .normal)
+        button.contentHorizontalAlignment = .fill
+        button.contentVerticalAlignment = .fill
+        button.imageView?.contentMode = .scaleAspectFit
+        return button
+    }()
+
+    private lazy var previewView: GalleryPreviewView = {
+        let previewView = GalleryPreviewView()
+        previewView.translatesAutoresizingMaskIntoConstraints = false
```
bstien: Most views which you instantiate with their default constructor, i.e. `init(frame:)`, can be switched to `init(withAutoLayout:)`. This calls `init(frame: .zero)` and sets `translatesAutoresizingMaskIntoConstraints` to `false`.

Not a big deal, but it's a handy init that saves a few lines of code :)

Demo/Fullscreen/FullscreenGallery/FullscreenGalleryDemoViewController.swift
bstien: Yup, I agree. I've run into the same issue, and when I saw you added `init(all:)` I had to check the documentation.

For `UIEdgeInsets` it says that positive values inset towards the center, while negative values expand from the center. This is not the case for `fillInSuperview(insets:)`, where it does not set trailing and bottom to negative values.

We could potentially rewrite that method to adhere to `UIEdgeInsets` documentation, but I guess we'll break some stuff.

pimms: `UIEdgeInsets` does seem to be equivalent to `padding` in HTML/CSS, indicating to me that the error lies in how we set up the constraints from them.

Example: https://stackoverflow.com/questions/27459746/adding-space-padding-to-a-uilabel

pimms: I'll just do this the old fashioned way for now to remove ambiguity.

bstien: Yes, the problem lies in `fillInSuperview(insets:)`.

Sources/Fullscreen/FullscreenGallery/FullscreenGalleryViewController.swift
```diff
@@ -0,0 +1,508 @@
...
+
+    private lazy var previewViewVisibleConstraint: NSLayoutConstraint = {
+        return previewView.bottomAnchor.constraint(equalTo: view.safeLayoutGuide.bottomAnchor)
+    }()
+
+    private lazy var previewViewHiddenConstraint: NSLayoutConstraint = {
+        // The constant exists to prevent the preview-view from jumping back into the visible area
+        // during the dismissal animation.
+        return previewView.topAnchor.constraint(equalTo: view.bottomAnchor, constant: .mediumLargeSpacing)
+    }()
```
bstien: Could rather be:

```swift
// The constant exists to prevent the preview-view from jumping back into the visible area
// during the dismissal animation.
private lazy var previewViewHiddenConstraint = previewView.topAnchor.constraint(equalTo: view.bottomAnchor, constant: .mediumLargeSpacing)
```

Sources/Fullscreen/FullscreenGallery/FullscreenGalleryViewController.swift
```diff
@@ -0,0 +1,508 @@
...
+
+    private lazy var singleTapGestureRecognizer: UITapGestureRecognizer = {
+        let recognizer = UITapGestureRecognizer()
+        recognizer.numberOfTapsRequired = 1
```
bstien: The default value is 1 tap. Could be removed.

Sources/Fullscreen/FullscreenGallery/FullscreenGalleryViewController.swift
pimms: It's actually `true` by default

bstien: Yeah sorry, you're right. Don't mind me😅

Sources/Fullscreen/FullscreenGallery/FullscreenGalleryViewController.swift
```diff
@@ -0,0 +1,508 @@
...
+            dataCallback(image)
+        })
+    }
+
+    func title(forImageViewController vc: FullscreenImageViewController) -> String? {
+        if vc.imageIndex < 0 || vc.imageIndex >= viewModel.imageCaptions.count {
+            return nil
+        }
+
+        return viewModel.imageCaptions[vc.imageIndex]
```
bstien: I see you use this alot in your code. We have a custom subscript in `ios-app` which you could probably use here as well.

```swift
return viewModel.imageCaptions[safe: vc.imageIndex]
```

The extension:
```swift
private extension Array {
    // Returns an Optional that will be nil if index < count
    public subscript(safe index: Int) -> Element? {
        return indices.contains(index) ? self[index] : .none
    }
}
```

Sources/Fullscreen/FullscreenGallery/FullscreenGalleryOverlayView.swift
```diff
@@ -0,0 +1,231 @@
...
+            return
+        }
+
+        if visible {
+            NSLayoutConstraint.deactivate([previewViewHiddenConstraint])
+            NSLayoutConstraint.activate([previewViewVisibleConstraint])
+        } else {
+            NSLayoutConstraint.deactivate([previewViewVisibleConstraint])
+            NSLayoutConstraint.activate([previewViewHiddenConstraint])
+        }
```
bstien: This could be rewritten as

```swift
previewViewVisibleConstraint.isActive = visible
previewViewHiddenConstraint.isActive = !visible
```

Sources/Fullscreen/FullscreenGallery/FullscreenImageView.swift
```diff
@@ -0,0 +1,162 @@
...
+
+// MARK: - Double tap handling
+
+extension FullscreenImageView {
+    @objc private func onDoubleTap(_ recognizer: UIGestureRecognizer) {
+        // When double tapping, only alternate between 1x and 2x. Further zooming may be permitted,
+        // but that has to be done explicitly. Double tapping an already zoomed in image should always
+        // result in zooming back out to 1x.
+        var newZoom: CGFloat
+        if zoomScale >= minimumZoomScale * 1.05 {
```
bstien: What does `1.05` mean? Could this be placed into a variable, perhaps?

pimms: This is a check to see if the view is zoomed in more than 5% – not sure if this warrants its own constant tbh.

pimms: The 5% is also completely arbitrary – what I'm really wanting to do is get rid of floating-point errors, so the check could _(should?)_ **really** have been:

`if zoomScale >= minimumZoomScale * 1.001 {`

bstien: If not a variable, a comment should be enough. Just to explain to people in the future why you're multiplying `minimumZoomScale` by an arbitrary value😅

Sources/Fullscreen/FullscreenGallery/FullscreenImageViewController.swift
```diff
@@ -0,0 +1,252 @@
...
+            }
+
+        default:
+            break
+        }
+    }
+
+    // MARK: - UIGestureRecognizerDelegate
+
+    public func gestureRecognizerShouldBegin(_ gestureRecognizer: UIGestureRecognizer) -> Bool {
```
bstien: This could be pulled into an extension.

Sources/Fullscreen/FullscreenGallery/FullscreenImageViewController.swift
```diff
@@ -0,0 +1,252 @@
...
+
+    public func updateLayout(withPreviewViewVisible previewViewVisible: Bool) {
+        self.previewViewVisible = previewViewVisible
+        fullscreenImageView.frame = calculateImageFrame()
+        fullscreenImageView.recalculateLimitsAndBounds()
+    }
+
+    func imageViewForDismissiveAnimation() -> UIImageView {
+        return panStateController?.panView ?? fullscreenImageView.imageView
+    }
```
bstien: This could potentially be a computed property

Sources/Fullscreen/FullscreenGallery/FullscreenImageView.swift
osanoj: Floating point errors like value being 0.9999997 (or something) instead of 1? You could try .round() or .rounded() then

Sources/Fullscreen/FullscreenGallery/GalleryPreviewView.swift
```diff
@@ -0,0 +1,170 @@
...
+        cell.clipsToBounds = false
+        cell.layer.masksToBounds = false
+        cell.layer.shadowColor = UIColor.black.cgColor
+        cell.layer.shadowOpacity = 0.3
+        cell.layer.shadowOffset = CGSize(width: 0, height: 0)
+        cell.layer.shadowRadius = 3
+
+        cell.layer.shadowPath = UIBezierPath(rect: estimatedFrame).cgPath
+        cell.layer.shouldRasterize = true
+        cell.layer.rasterizationScale = UIScreen.main.scale
```
bstien: Could this setup be placed into `GalleryPreviewCell.setup()` instead?

Sources/Fullscreen/FullscreenGallery/FullscreenImageView.swift
pimms: `zoomScale` and `minimumZoomScale` will hold arbitrary float values though _(often in the range `0.3...1.0`)_, so rounding them would remove all information from them.

Sources/Fullscreen/FullscreenGallery/FullscreenGalleryOverlayView.swift
pimms: I didn't know that, thanks! This makes the constraints invalid when showing the thumbnails however, as the `hidden` and `visible` constraints are active simultaneously. This makes auto-layout complain and attempt to break one constraint to resolve the conflict - it seems to always break the correct constraint, but I don't think that's a die we want to roll in production... 🙂

Sources/Fullscreen/FullscreenGallery/FullscreenImageView.swift
pimms: Played around with this a bit more – if the user has accidentally zoomed in slightly _(2-3%)_, it feels "off" that double-tapping zooms out. I'm changing the limit to 10% – if you're zoomed in less than this limit a double-tap will zoom **in**.

Sources/Fullscreen/FullscreenGallery/GalleryPreviewView.swift
pimms: The cell doesn't know his future size at all though, so it would have to be through a separate method on the form `GalleryPreviewCell.addShadow(withEstimatedSize: CGRect)`. What do you think of this?

bstien: How about overriding `layoutSubviews` and set it there? I think the cell's frame should already be set.

pimms: Ooh, that works! 🙂

Sources/Fullscreen/FullscreenGallery/Transition/FullscreenGalleryTransitionPresenterDelegate.swift
```diff
@@ -0,0 +1,22 @@
+//
+// Copyright (c) 2019 FINN AS. All rights reserved.
+//
+
+import UIKit
+
+/// The Fullscreen Gallery Transition is a round-trip, and the Transition Presenter Delegate will be used
+/// both when transitioning in, and when transitioning out.
+public protocol FullscreenGalleryTransitionPresenterDelegate: class {
```
bstien: As @osanoj mentioned in another PR, we should start using `AnyObject` over `class` for protocols. As mentioned in [this Swift Evolution proposal](https://github.com/apple/swift-evolution/blob/master/proposals/0156-subclass-existentials.md#class-and-anyobject), `class` may be removed at some point.

No worries at this point at all, but it could save us some changes when/if it'll be removed :)

Demo/Fullscreen/FullscreenGallery/FullscreenGalleryDemoViewController.swift
```diff
@@ -0,0 +1,309 @@
...
+
+        playgroundView.addSubview(collectionView)
+        playgroundView.addSubview(thumbnailSwitch)
+        playgroundView.addSubview(thumbnailLabel)
+        playgroundView.addSubview(simulateLoadingSwitch)
+        playgroundView.addSubview(simulateLoadingLabel)
+        playgroundView.addSubview(helpLabel)
+
+        NSLayoutConstraint.activate([
+            collectionView.heightAnchor.constraint(equalToConstant: collectionCellHeight + CGFloat.largeSpacing),
```
vadymmarkov: Can be just `.largeSpacing`

Demo/Fullscreen/FullscreenGallery/FullscreenGalleryDemoViewController.swift
```diff
@@ -0,0 +1,309 @@
...
+            simulateLoadingSwitch.topAnchor.constraint(equalTo: thumbnailSwitch.bottomAnchor, constant: .mediumSpacing),
+            simulateLoadingSwitch.leadingAnchor.constraint(equalTo: playgroundView.leadingAnchor, constant: .largeSpacing),
+
+            simulateLoadingLabel.leadingAnchor.constraint(equalTo: simulateLoadingSwitch.trailingAnchor, constant: .mediumLargeSpacing),
+            simulateLoadingLabel.centerYAnchor.constraint(equalTo: simulateLoadingSwitch.centerYAnchor),
+
+            helpLabel.leadingAnchor.constraint(equalTo: playgroundView.leadingAnchor, constant: .mediumLargeSpacing),
+            helpLabel.trailingAnchor.constraint(equalTo: playgroundView.trailingAnchor, constant: -.mediumLargeSpacing),
+            helpLabel.topAnchor.constraint(equalTo: playgroundView.safeLayoutGuide.topAnchor, constant: .largeSpacing),
+
```
vadymmarkov: Extra comma and line?

Sources/Fullscreen/FullscreenGallery/FullscreenGalleryViewController.swift
```diff
@@ -0,0 +1,411 @@
...
+                }
+                imageVc.updateLayout(withPreviewViewVisible: visible)
+            })
+        })
+    }
+
+    // MARK: - View modifications
+
+    private func transitionToImage(atIndex index: Int, animated: Bool) {
+        if let currentIndex = currentImageViewController()?.imageIndex {
```
vadymmarkov: How about `guard currentImageViewController()?.imageIndex != index else { return }`?

Sources/Fullscreen/FullscreenGallery/FullscreenImageView.swift
```diff
@@ -0,0 +1,167 @@
...
+// Copyright (c) 2019 FINN AS. All rights reserved.
+//
+
+import UIKit
+
+class FullscreenImageView: UIScrollView {
+
+    // MARK: - Public properties
+
+    public var image: UIImage? {
```
vadymmarkov: I think you can remove `public` from vars and functions in this class since it's internal.

Sources/Fullscreen/FullscreenGallery/FullscreenImageViewController.swift
```diff
@@ -0,0 +1,256 @@
...
+    /// - Returns
+    ///   True if the FullscreenImageViewController should animate the primary image view back into position.
+    ///   False otherwise.
+    func fullscreenImageViewControllerDidEndPan(_: FullscreenImageViewController, withTranslation translation: CGPoint, velocity: CGPoint) -> Bool
+}
+
+class FullscreenImageViewController: UIViewController {
+    // MARK: - Public properties
+
+    public weak var dataSource: FullscreenImageViewControllerDataSource?
```
vadymmarkov: The same comment about `public` for this class.

Sources/Fullscreen/FullscreenGallery/Transition/FullscreenGalleryDismissalTransition.swift
```diff
@@ -0,0 +1,116 @@
...
+// Copyright (c) 2019 FINN AS. All rights reserved.
+//
+
+import UIKit
+
+class FullscreenGalleryDismissalTransition: NSObject, UIViewControllerAnimatedTransitioning {
+
+    // MARK: - Public properties
+
+    public var dismissVelocity: CGPoint?
```
vadymmarkov: No need for `public`?

Sources/Fullscreen/FullscreenGallery/FullscreenImageView.swift
pimms: Still getting used to the Swift-modifiers... 🙃

bstien:
I'm merging this one since @pimms is not yet a member of the finn-no/ios team.
Do you have access to add him, @osanoj?

osanoj:
@bstien: Can't add @pimms to the team since he is not part of the organization yet, so we need to fix that first

PR #350, opened 2019-03-01
Fix TextView delegate, part two

bstien:
# Why?
After my last PR (#349) I noticed it was ambiguous which delegate method would be called from `TextView` when it called `delegate?.textViewDidChange(_:)`. This was due to the delegate protocol inheriting from `UITextViewDelegate`.

# What?
- Removed inheritance from `UITextViewDelegate`, so we may send a `TextView` instance to the delegate calls, instead of `UITextView`.
- Added new methods to protocol, with empty implementations.

### DIFF
--- a/Sources/Components/TextView/TextView.swift
+++ b/Sources/Components/TextView/TextView.swift
@@ -6,6 +6,9 @@ import UIKit
 
 public protocol TextViewDelegate: class {
     func textViewDidChange(_ textView: TextView)
+    func textViewDidBeginEditing(_ textView: TextView)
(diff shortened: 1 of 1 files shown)


Sources/Components/TextView/TextView.swift
```diff
@@ -4,8 +4,22 @@
 
 import UIKit
 
-public protocol TextViewDelegate: UITextViewDelegate {
+public protocol TextViewDelegate: class {
```
osanoj: You could change to `AnyObject`, I think that is preferable even though they at the moment gives the same result :)

PR #351, opened 2019-03-01
Add section property in UserAdsListHeaderView

robinsalehjan:
# Why?

- To check which section the header belongs to.

# What?

- Add a property to the `UserAdsListHeaderView` class which has to be provided at initialisation.

# Show me
No UI changes


Sources/Recycling/ListViews/UserAds/ListView/UserAdsListHeaderView.swift
```diff
@@ -13,6 +13,7 @@ public class UserAdsListHeaderView: UIView {
     // MARK: - Public properties
 
     public weak var delegate: UserAdsListHeaderViewDelegate?
+    public var section: Int
```
vadymmarkov: Could it be `let`?

robinsalehjan: Fixed 😃

PR #361, opened 2019-03-08
Removed badge-functionality from Market Grid

pimms:
# Why?

The FINN Shopping badge _"Kun nytt"_ is going to be removed, and we therefore don't need the badge functionality anymore.

# Show me

image


bstien:
You could probably also remove the image `onlyNew`. I don't think we use it anywhere else🤔

PR #364, opened 2019-03-11
Notify BottomSheet's delegate on drag

bstien:
# Why?
I'm in the situation where I need to know when the user begins dragging the bottomSheet so I can dismiss the keyboard.

# What?
- Added methods to both `BottomSheetDelegate` and `BottomSheetPresentationControllerDelegate`.


bstien:
@shredlocker @vadymmarkov 
Do you guys think we should have a default blank implementation of this method, so it doesn't force something unneeded/unwanted into Charcoal?

vadymmarkov:
I'm not a big fan of optional protocol methods. I think we could either add a blank implementation in `Charcoal` or introduce a new delegate protocol here, like `BottomSheetDragDelegate`.

bstien:
Yeah, probably a better idea to put it into a separate protocol 👍 There might be some other drag events we need later, so it's better to have them separated.

PR #365, opened 2019-03-11
Feature/add/dialogue

ag12:
# Why?
https://app.zeplin.io/project/59b68125aaf5b20574170f3f/screen/5c585ffbb74a443081a0881a

# What?
Added simple Dialogue View, Dialogue View Controller, and the new Inline Consent View. 

### New UI
Simulator Screen Shot - iPad Air - 2019-03-11 at 10 06 15

Simulator Screen Shot - iPad Air - 2019-03-11 at 10 06 22

Simulator Screen Shot - iPhone X - 2019-03-11 at 10 06 35

Simulator Screen Shot - iPad Air - 2019-03-11 at 10 13 47

Simulator Screen Shot - iPhone X - 2019-03-11 at 10 13 54


Sources/Components/Dialogue/DialogueView.swift
```diff
@@ -0,0 +1,172 @@
...
+
+// MARK: - DialogueViewDelegate
+
+/**
+ - Note:
+    Protocol to call back the Dialogue view Delegate
+
+ */
+
+public protocol DialogueViewDelegate: class {
```
bstien: We should start using `AnyObject`, instead of `class`, for class-only protocols since `class` might be [removed in the next version of Swift](https://github.com/apple/swift-evolution/blob/master/proposals/0156-subclass-existentials.md#class-and-anyobject).

Demo/Components/Dialogue/DialogueViewController.swift
```diff
@@ -0,0 +1,51 @@
...
+
+public struct DialogueDefaultData: DialogueViewModel {
+    public let title = "Slå på anbefalinger"
+    public let detail = "Vi viser deg relevante FINN-annonser og tilpasser FINN etter din bruk. For å gjøre dette lagrer vi info om hva du ser på hos oss."
+    public var link: String? = "Mer om samtykke"
+    public let primaryButtonTitle = "Aktiver personlige anbefalinger"
+    public var secondaryButtonTitle: String? = "Alternativ knapp"
+}
+
+public protocol DialogueViewControllerDelegate: class {
```
bstien: Same here regarding `AnyObject` vs. `class`.

Sources/Components/Dialogue/DialogueView.swift
ag12: Alright! nice :)

PR #369, opened 2019-03-18
Refactor ReviewView and Button

bstien:
# Why?
We need a way to disable/enable the `selectButton` in `ReviewView` after a selection has been made, seeing as the network request we make in the app takes a long time. I've added the method `setSelectButtonEnabled(_: Bool)` to that view to control this.

After disabling the button in the demo I noticed the button didn't change its colors, even though it was disabled. The reason for this was `Button` not correctly setting its colors based on the state. This can be seen in the before gif below.

# What?
- Correct access control and `MARK`s for `ReviewView`.
- Added method to control enabled state for button in `ReviewView`.
- Set correct colors for `Button`, based on state.
- Refactor Demo to simulate a 3 second load time, by disabling the button and presenting a `LoadingView`.

# Show me
### Before & after
_In both videos I set the button to disabled on click, and enable it again after 3 secs. Only after has the loading view added._

reviewview before
reviewview after


Demo/Fullscreen/Review/ReviewViewDemoView.swift
```diff
@@ -31,7 +31,13 @@ public class ReviewViewDemoView: UIView {
 
 extension ReviewViewDemoView: ReviewViewDelegate {
     public func reviewView(_ reviewView: ReviewView, didSelect profile: ReviewViewProfileModel) {
+        LoadingView.show()
+        reviewView.setSelectButtonEnabled(false)
         print("Did select: \(profile.name) for review")
```
robinsalehjan: Debug code?

bstien: Demo code 😄 Did not change this line, though

robinsalehjan: Yes probably leftover

PR #373, opened 2019-03-20
Set ribbonView to have highest resistance priority

robinsalehjan:
# Why?

- The `ribbonView` should not be compressed if the date or price happen to be very long.

# What?

- Add `contentCompressionResistancePriority` to `titleLabel`, `detailLabel` and `ribbonView`
- Tweak the `detailLabel` and the `priceLabel` trailing in relation to the `ribbonView`

# Show me

### Before

before

### After

after


pimms:
It seems you may be able to remove the compression resistances on the labels – I removed them, and I'm getting this result on a iPhone 7+ _(after changing the detail & prices to be super long)_:

Screen Shot 2019-03-21 at 12 01 07

bstien:
Thinking about it, it may only be necessary to set either `contentHuggingPriority` or `contentCompressionResistancePriority` to `.required` for `ribbonLabel`

robinsalehjan:
Cool, will fix 👍

robinsalehjan:
This `UITableView` has a trick up it sleeves, it can render the `inactive section` with a separate cell design, so if I remove the resistance priority it will compress the `ribbonView`.

On a iPhone 5s the `titleLabel` will compress the `ribbonView` without that resistance (both for `inactive` and `active` cells)

Simulator Screen Shot - iPhone 5s - 2019-03-21 at 13 52 25

PR #376, opened 2019-03-25
Feature/add/inline consent dialog to front page

ag12:
# What?
New Dialogue Component 
And Inline Consent.


Demo/Extensions/UIKit/UIDevice.swift
```diff
@@ -6,7 +6,7 @@ import UIKit.UIDevice
 
 // MARK: - UIDevice
 
- extension UIDevice {
+ internal extension UIDevice {
```
vadymmarkov: It should be `internal` by default.

Sources/Extensions/UIKit/UIDevice.swift
```diff
@@ -0,0 +1,26 @@
+//
+//  Copyright © 2019 FINN AS. All rights reserved.
+//
+
+import UIKit.UIDevice
+
+// MARK: - UIDevice
+
+internal extension UIDevice {
```
vadymmarkov: The same comment about `internal` as above.

Sources/Fullscreen/FrontPage/FrontPageView.swift
```diff
@@ -185,28 +186,51 @@ public final class FrontPageView: UIView {
...
         adsRetryView.frame.origin = CGPoint(x: 0, y: headerView.frame.height + .veryLargeSpacing)
         adsRetryView.frame.size = CGSize(width: bounds.width, height: 200)
         boundsForCurrentSubviewSetup = bounds
         adsGridView.invalidateLayout()
+        setupFrameForDialogue(height: height)
     }
 
-    private func updateHeaderTitle() {
-        headerLabel.text = inlineConsentView.isHidden ? model?.adsGridViewHeaderTitle : model?.inlineConsentTitle
+    func setupFrameForDialogue(height: CGFloat) {
```
vadymmarkov: Maybe `y` or `yPosition` is better seeing you don't use it as a height.

Sources/Extensions/UIKit/UIDevice.swift
```diff
@@ -0,0 +1,26 @@
...
+//  Copyright © 2019 FINN AS. All rights reserved.
+//
+
+import UIKit.UIDevice
+
+// MARK: - UIDevice
+
+extension UIDevice {
+
+    internal class func isIPad() -> Bool {
```
vadymmarkov: You can also remove `internal` from class function here.

Sources/Components/Dialogue/DialogueView.swift
```diff
@@ -147,25 +138,42 @@ public class DialogueView: UIView {
...
+            link.topAnchor.constraint(equalTo: primaryButton.bottomAnchor, constant: .mediumSpacing),
             link.leadingAnchor.constraint(equalTo: leadingAnchor, constant: .largeSpacing),
             link.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -.largeSpacing),
-
-            primaryButton.topAnchor.constraint(equalTo: link.bottomAnchor, constant: .mediumLargeSpacing),
-            primaryButton.centerXAnchor.constraint(equalTo: centerXAnchor),
             ])
     }
 
+    func heightWithConstrainedWidth(width: CGFloat) -> CGFloat {
```
pimms: What about `heightWithConstrainedWidth(_ width: CGFloat)` instead?

Sources/Components/Dialogue/DialogueView.swift
```diff
@@ -147,25 +138,42 @@ public class DialogueView: UIView {
...
+        let constraintRect = CGSize(width: width, height: .greatestFiniteMagnitude)
+        let boundingBox = text.boundingRect(with: constraintRect, options: [.usesLineFragmentOrigin], attributes: [NSAttributedString.Key.font: detail.font], context: nil)
+
+        var moreSpacing: CGFloat = 0
+
+        if UIDevice.isSmallScreen() {
+            moreSpacing = .largeSpacing
+        }
+
+        if UIDevice.isLandscape() {
```
pimms: Should this be an `else if UIDevice.isLandscape` instead? It looks like they are exclusive.

Sources/Extensions/UIKit/UIViewExtensions.swift
```diff
@@ -33,4 +33,13 @@ public extension UIView {
             return layoutMarginsGuide
         }
     }
+
+    public func dropShadow(color: UIColor, opacity: Float = 0.5, offset: CGSize = CGSize.zero, radius: CGFloat = 10.0, usingPath: Bool = true) {
```
pimms: `usingPath` is not used

ag12:
Thanks for the feedback @finn-no/ios

PR #378, opened 2019-03-25
Update/rename bold fonts

bstien:
# Why?
Some of the fonts should have their names updated to be in sync with the names our designers use. See spec here: https://zpl.io/aMd1exr

These fonts have been renamed:
- `title4` → `bodyStrong`
- `title5` → `detailStrong`
- `captionHeavy` → `captionStrong`

# What?
- Renamed fonts and all usages.

# Show me
_No UI changes._


bstien:
‼️ This PR can't be merged before the same changes has been made to FinnBox and Charcoal. I'm working on those now.

bstien:
Since it would be kind of a hassle to merge these changes in Finnbox, Charcoal and ios-app in sync, I've re-added the removed font types and added a deprecation warning so we'll know that we should use the new names.

PR #379, opened 2019-03-25
Stepper indicator view

pimms:
# Why?

We need a nice way to indicate the progress through a discrete process. This implementation is part of the T&R design sketches.

# What?

 - All positioning are handled via constraints
 - Simple interface to set current step & handling taps on "step dots"
 - Supports an arbitrary _([2..∞>)_ number of dots

*Note:* The sketches are ambiguous in what to do when the user would navigate backwards. I think the implementation I made looks pretty good. I tried a few different approaches, but found this to look the most elegant. @ingridvf, do you have any input on this?

# Show me

Simple step-through
stepindicator-simple

"Random access"
stepindicator-tapping

Orientation change
stepindicator-orientation


Sources/Components/Broadcast/StepIndicator.swift
```diff
@@ -0,0 +1,123 @@
...
+            sendSubviewToBack(connector)
+            connectors.append(connector)
+        }
+
+        updateCurrentStep(previousValue: 0)
+    }
+
+    // MARK: - Private methods
+
+    private func updateCurrentStep(previousValue: Int) {
```
bstien: `previousValue` doesn't seem to be used. Can it be removed?

pimms: Yup, nice spot 🙂 Leftover from an earlier iteration.

ingridvf:
Awesome! When navigating back, I think it should be styled the way it is when it is not completed, this way it is more clear what is the current step:)

ingridvf:
Maybe we can make the Sardine ones filled when navigating back, to indicate that the step is done?

PR #380, opened 2019-03-26
Made StepIndicator an open class

pimms:
# Why?

In order to properly integrate the `StepIndicator` with the T&R SDK, I need to extend it.

# What?

`StepIndicator` is now open.

# Show me

```
open class StepIndicator: UIView {
```


pimms:
Nevermind, I think this is easier solved through composition.

PR #381, opened 2019-03-26
Improvements to StepIndicator highlighting

pimms:
# Why?

Ingrid had a good idea for how to better visualize that the user has stepped backwards in the `StepIndicator` view.

# What?

The furthermost step is now sardine, filled, and the connector is sardine.

# Show me

### Before

Screen Shot 2019-03-26 at 09 21 01

### After

Screen Shot 2019-03-26 at 09 21 15

stepindicator-improve


ingridvf:
Great job :)

PR #382, opened 2019-03-26
Re-add removed enum cases for Label.Style, with deprecation warnings

bstien:
# Why?
The enum `Label.Style` had a 1:1 relationship with our custom fonts, which I didn't see yesterday.

This caused an issue with the iOS-app not building when pulling the FK from branch `master`, since it depended on `.title4` etc.

# What?
- Re-added the removed enum cases, this time with a deprecation warning.

# Show me
_No UI changes._


bstien:
Ref. PR #378

PR #383, opened 2019-03-27
Tweak cell spacing to use multiplies of 4 or 8

robinsalehjan:
# Why?

- To be consistent with spacing and sizing of views.

# What?

- Tweak spacing of elements in the cell to preferably use `4` or `8`
- Change image size for active and inactive cells to be a multiple of `8`
- Adjust the `CreateNewAd` height and width anchor to use a multiplier of `0.65` 
- Adjust the `ViewAllAds` height and with anchor to use a multiplier of `1` 

# Show me

### Before

old-iphone

old-ipad

### After

new-iphone

new-ipad


Sources/Recycling/ListViews/UserAds/Cell/UserAdsListViewSeeAllAdsCell.swift
```diff
@@ -43,8 +43,8 @@ public class UserAdsListViewSeeAllAdsCell: UITableViewCell {
         addSubview(seeAllAdsButton)
 
         NSLayoutConstraint.activate([
-            seeAllAdsButton.heightAnchor.constraint(equalTo: heightAnchor, multiplier: 0.6),
-            seeAllAdsButton.widthAnchor.constraint(equalTo: widthAnchor, multiplier: 0.8),
+            seeAllAdsButton.heightAnchor.constraint(equalTo: heightAnchor, multiplier: 1),
```
nickfalk: Guess the multiplier is not really needed here anymore. :-)

robinsalehjan: Nice catch, fixed :)

PR #385, opened 2019-03-28
removes unnecessary extension

shredlocker:
# Why?

After updating to Xcode 10.2 there was an error with `AVAudioSession.setCategory()`. It was previously a bug in Swift 4.2 which is now fixed with Swift 5.0

# What?

- Removed previously created extensions to work around this problem.

# Show me

No UI


pimms:
Does this mean that we need to use Xcode 10.2 to build FinniversKit?

bstien:
Seems like Travis at least needs that version😕

shredlocker:
Things doesn't look good xD not sure what to do about this.

osanoj:
@shredlocker Maybe downgrade to Xcode 10.1 for now, and then we can look at it more next week...

bstien:
Guess this one is ready to be closed now, ref. #406

PR #387, opened 2019-03-29
Improve support for people who use dynamic font size

robinsalehjan:
# Why?

- Inconsistent font sizes among labels and not a pleasant experience

# What?

- Adjust constraints

# Show me

| Before  | After |
| ------- | ------ |
| before - iPhone | after - iPhone  |


robinsalehjan:
It's not perfect, but at least it's better. Need to rewrite the constraints.

pimms:
Most of the labels are not fully readable now – would it make sense to allow for multiple-line labels? The descriptive labels are maybe not as important, but it's hard to understand what the button _"Lag en…nnonse"_ does.. 🙂

PR #397, opened 2019-04-04
Add awaiting payment status

robinsalehjan:
# Why?

- The status `Venter på betaling` did not exist in the enum.
- `QualityGate` was complaining about the long `function` signature.

# What?

- Add the `Venter på betaling` case.
- Shorten the `parameter` name in the `function` signature.

# Show me
|     Before                   |            After               |
| before | after |


ag12:
B 4 Bonus.

PR #398, opened 2019-04-05
Feature/native adverts

pimms:
# Why?

I didn't want to add the native content-ad view into the app itself, and I moved the existing native advert view into FinniversKit as well.

# What?

- `NativeAdvertView` _(moved from `ios-app`)_, displays regular native ads
- `NativeContentAdvertView` _(new)_, displays content-ads ("annonsørinnhold")
- The two views share no ancestry, but **do** share a view-model.
  - The view-model shares 3/4 properties with `ANNativeAdResponse`, but `sponsoredText` is unique to `NativeAdvertViewModel`.
- The new cogwheel-button required too many adjustments from the existing `CogWheelButton` to comfortably shoe-horn as customizable attributes so I created a new one.

The idea is that when we receive an `ANNativeAdResponse` in the app, we will check the type of the ad and instantiate the correct view _(`NativeAdvertView` or `NativeContentAdvertView`)_ and instantiate a view-model with the correct `sponsoredText` attribute.

- In the app, the views are always 250px tall, so the demo reflects this.
- The `NativeAdvertView` had a hard width-cap of `400px` in the app, this is retained
- The `NativeContentAdvertView` has a hard width-cap of `360px` as the main image is scaled to fill, and such a wide view would look silly.

**NOTE!** The design & layout are not accepted by UX yet, I'll do this on Monday. Any changes after this should be super trivial.

# Show me

### iPhone 7
native-ads-fk-iphone

### iPad (5th)
native-ads-fk-ipad


pimms:
Ran through the design with Julie and made some adjustments. Nothing major, but most notably we enforce the massively popular aspect-ratio `400:209` on the primary image. Images with other aspect-ratios will be content-filled.

We also had to adjust the paddings around the title-label slightly in order to guarantee a fit for 50 characters. If the provided title is too long, the label will auto-scale down to 50% font-size.

Screen Shot 2019-04-08 at 14 18 46
Screen Shot 2019-04-08 at 14 18 18

Sources/Components/NativeAdvert/NativeContentAdvertView.swift
```diff
@@ -0,0 +1,136 @@
...
+        let view = UIImageView(withAutoLayout: true)
+        view.contentMode = .scaleAspectFit
+        return view
+    }()
+
+    private lazy var titleLabel: Label = {
+        let label = Label(style: .title2)
+        label.translatesAutoresizingMaskIntoConstraints = false
+        label.numberOfLines = 0
+        label.minimumScaleFactor = 0.5
```
bstien: I guess such a low scale-factor is is set `just in case the title is loooooooooooo…`?

pimms: Yeah – or contains a lot of wide characters. We need to support 50 characters, period – I interpreted that to mean 50 *wide* characters, such as `M` as well 🙂

Sources/Components/NativeAdvert/NativeAdvertView.swift
```diff
@@ -0,0 +1,250 @@
...
+
+            logoImageView.widthAnchor.constraint(equalToConstant: logoSize),
+            logoImageView.heightAnchor.constraint(equalToConstant: logoSize),
+            logoImageView.centerYAnchor.constraint(equalTo: bottomContainerView.centerYAnchor),
+            logoImageView.trailingAnchor.constraint(equalTo: bottomContainerView.trailingAnchor)
+            ])
+        setupFonts(largeTitles: largeTitles)
+    }
+
+    func build(viewModel: NativeAdvertViewModel) {
```
vadymmarkov: We usually call methods like this `configure(with viewModel:)`.

Sources/Components/NativeAdvert/NativeAdvertView.swift
```diff
@@ -0,0 +1,250 @@
...
+        logoImageView.image = nil
+        if let imageURL = viewModel.iconImageURL {
+            imageDelegate?.nativeAdvertView(setImageWithURL: imageURL, onImageView: logoImageView)
+        }
+
+        titleLabel.text = viewModel.title
+        sponsoredByLabel.text = viewModel.sponsoredText
+    }
+
+    func setupFonts(largeTitles: Bool) {
```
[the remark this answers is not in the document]
pimms: Yup, redid this part.

PR #399, opened 2019-04-08
Implement StatisticsItemEmptyView for AdManagement

robinsalehjan:
# Why?

- We were missing a `EmptyView` for the statistics section in the `AdManagement` screen.

# What?

- Implement the `EmptyView`

# Show me
|     Before              |            After               |
| before | after |


Sources/Recycling/ListViews/AdManagement/Cells/UserAdManagementStatisticsCell.swift
```diff
@@ -3,15 +3,28 @@
...
+    // MARK: - Public
+
     public var itemModels = [StatisticsItemModel]() {
         didSet { updateStackViewContent() }
     }
+
+    // MARK: - Private
+
+    private lazy var emptyView: StatisticsItemEmptyView = {
+        let view = StatisticsItemEmptyView(title: "Følg med på effekten", description: "Etter at du har publisert annonsen din kan du se statistikk for hvor mange som har sett annonsen din, favorisert den og som har fått tips om den.")
```
pimms: Should we have inline texts like this? I don't think it necessarily makes sense to inject "static" texts like these from the main app, but I also think it doesn't look right to have them hard-coded inline.

Would it make sense to have a `Localizable.strings` inside FinniversKit?

Sources/Components/StatisticsView/StatisticsItemEmptyView.swift
```diff
@@ -0,0 +1,73 @@
...
+        let view = UIImageView(image: UIImage(named: .statsEmpty))
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.contentMode = .scaleAspectFit
+        view.tintColor = .ice
+        return view
+    }()
+
+    private lazy var titleLabel: UILabel = {
+        let label = UILabel()
+        label.translatesAutoresizingMaskIntoConstraints = false
```
bstien: You could use `UILabel(withAutoLayout: true)` instead

Sources/Components/StatisticsView/StatisticsItemEmptyView.swift
```diff
@@ -0,0 +1,73 @@
...
+        label.font = UIFont.bodyStrong
+        label.textColor = .licorice
+        label.textAlignment = .center
+        label.setContentHuggingPriority(.defaultHigh, for: .vertical)
+        return label
+    }()
+
+    private lazy var descriptionLabel: UILabel = {
+        let label = UILabel()
+        label.translatesAutoresizingMaskIntoConstraints = false
```
bstien: Same here

Sources/Recycling/ListViews/AdManagement/Cells/UserAdManagementStatisticsCell.swift
robinsalehjan: This isn't correct, brain fart.

The text is there for demo purposes, but in the actual app they will be in the `Localizable.strings` file.

I'll remove them from the constructor and make them part of the model or a separate model.

Sources/Recycling/ListViews/AdManagement/Cells/UserAdManagementStatisticsCell.swift
```diff
@@ -33,17 +46,22 @@ public class UserAdManagementStatisticsCell: UITableViewCell {
...
+
+            for (index, model) in itemModels.enumerated() {
+                let itemView = StatisticsItemView(model: model)
+                itemView.shouldShowLeftSeparator = index > 0 && index < lastIndex
+                itemView.shouldShowRightSeparator = index > 0 && index < lastIndex
+                // Slight overengineering, as there are currently no plans to receive any other number than three items
+                if lastIndex == 1 { itemView.shouldShowLeftSeparator = index == lastIndex }
+                if lastIndex > 2 { itemView.shouldShowRightSeparator = index == lastIndex-1 }
+                stackView.addArrangedSubview(itemView)
+                itemView.setupConstraints()
```
bstien: Could you move this call to one of the lifecycle methods on the view itself? I.e. `viewDidLoad()`?

bstien: I see now that you didn't add this loc and that it's added to a stackView. Don't mind me :)

Demo/Recycling/ListViews/AdManagement/AdManagementDemoView.swift
```diff
@@ -59,32 +65,40 @@ public class AdManagementDemoView: UIView {
...
 
     public func numberOfSections(in tableView: UITableView) -> Int {
-        return 1 + actionCellModels.count
+        return actionCellModels.count + 1
     }
 
     public func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
         if indexPath.section == 0 {
-            if indexPath.row == 0 {
+            if indexPath.row == 0 && statisticsCellModels.isEmpty == true {
```
nickfalk: I think this if/else is uncalled for. :-)

PR #402, opened 2019-04-10
Bump to 10

ag12:
# Why?
The main app doesn't use 9. 

# What?
Bumped to iOS 10


Sources/Components/Feedback/FeedbackView.swift
```diff
@@ -145,16 +145,12 @@ public class FeedbackView: UIView {
...
-        }
+        UIImpactFeedbackGenerator(style: .light).impactOccurred()
         delegate?.feedbackView(self, didSelectButtonOfType: .positive, forState: state)
     }
 
     @objc private func negativeButtonTapped() {
-        if #available(iOS 10.0, *) {
-            UIImpactFeedbackGenerator(style: .light).impactOccurred()
-        }
+        UIImpactFeedbackGenerator(style: .light).impactOccurred()
```
bstien: Maybe we could pull `UIImpactFeedbackGenerator` into an instance var now? Not sure if it has any impact (pun intended) on performance, but it makes sense to not create it each time we need it.

ag12: Yeah, good idea! Should keep this PR as is.

PR #410, opened 2019-04-26
Fading in frontpage recommendation images

pimms:
# Why?

In high-latency situations, the instant fade-into-existence is quite eye-gnawing... 🙂

# What?

Fading in images for frontpage recommendations. Note that I also don't clear the background color when there is no image, I haven't spoken to Thuy or anyone else about this yet. I think it looks kickass FWIW 🔥

# Show me

### Before
frontpage-highlatency-no-fade

### After
frontpage-highlatency-fade

### DIFF
--- a/Sources/Recycling/GridViews/Ads/Cell/AdsGridViewCell.swift
+++ b/Sources/Recycling/GridViews/Ads/Cell/AdsGridViewCell.swift
@@ -27,10 +27,17 @@ public class AdsGridViewCell: UICollectionViewCell {
     private static let imageDescriptionHeight: CGFloat = 35.0
     private static let iconSize: CGFloat = 23.0
 
+    private lazy var imageBackgroundView: UIView = {
+        let view = UIView()
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.layer.cornerRadius = AdsGridViewCell.cornerRadius
+        view.layer.masksToBounds = true
+        return view
+    }()
+
     private lazy var imageView: UIImageView = {
         let imageView = UIImageView()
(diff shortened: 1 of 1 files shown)


Sources/Recycling/GridViews/Ads/Cell/AdsGridViewCell.swift
```diff
@@ -27,10 +27,17 @@ public class AdsGridViewCell: UICollectionViewCell {
     private static let imageDescriptionHeight: CGFloat = 35.0
     private static let iconSize: CGFloat = 23.0
 
+    private lazy var imageBackgroundView: UIView = {
+        let view = UIView()
```
bstien: You could use `UIView(withAutoLayout: true)` 💥

Sources/Recycling/GridViews/Ads/Cell/AdsGridViewCell.swift
```diff
@@ -235,28 +249,35 @@ public class AdsGridViewCell: UICollectionViewCell {
...
 
+    // MARK: - Private
+
+    private func setImage(_ image: UIImage?) {
+        imageView.alpha = 0.0
+        imageView.image = image
+
+        UIView.animate(withDuration: 0.2, delay: 0.0, options: [.curveEaseOut], animations: {
+            self.imageView.alpha = 1.0
+        }, completion: nil)
```
bstien: You can remove `completion: nil`, since `nil` is already the default argument

osanoj:
Maybe get that colorful no image background approved before merging too 😄

PR #412, opened 2019-04-30
Resolved accidentally introduced frontpage flickering

pimms:
# Why?

After integrating PR #410 into the FINN-app I noticed that I introduced a lot of flickering because we call `AdsGridView.reloadData()` whenever we load additional recommendations.

Also, for transparent primary images & unusual aspect ratios, the background color of the `AdsGridViewCell`s was visible.

# What?

- Simulating the case that caused the flickering in the FINN-app in the FinniversKit demo.

- The `AdsGridView` caches all images _(weakly)_ after fetching.
- `AdsGridViewCell` will ask its' delegate for a cached image before asking the delegate to download the image. If the cached image exists, the cell will **not** animate the transition _(the animation is what caused the flickering)_.

- After setting **any non-default image**, the background color of the `AdsGridViewCell` is set to `.clear`.

# Show me

### Before

frontpage-reload-flicker

### After

frontpage-reload-noflicker
ios


Demo/Fullscreen/FrontPage/FrontPageViewDemoView.swift
```diff
@@ -44,7 +45,17 @@ extension FrontpageViewDemoView: FrontPageViewDelegate {
 }
 
 extension FrontpageViewDemoView: AdsGridViewDelegate {
-    public func adsGridView(_ adsGridView: AdsGridView, willDisplayItemAtIndex index: Int) {}
+    public func adsGridView(_ adsGridView: AdsGridView, willDisplayItemAtIndex index: Int) {
+        if index >= visibleItems - 10 {
+            visibleItems += 10
+
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.5, execute: { [weak self] in
+                self!.frontPageView.reloadAds()
```
bstien: Why not `self?`?

pimms: Ah snap, I was testing something out and forgot to clean up after myself. Will fix!!!

Sources/Recycling/GridViews/Ads/Model/AdsGridImageCache.swift
```diff
@@ -0,0 +1,34 @@
...
+private class WeakImage {
+    private(set) weak var value: UIImage?
+
+    init(value: UIImage) {
+        self.value = value
+    }
+}
+
+class AdsGridImageCache {
+    private var cache: [String: WeakImage] = [:]
```
vadymmarkov: Might be worth using `NSCache` instead of dictionary here. Then you can set some count and cost limit and objects will be automatically removed from the cache if the device is running low on memory.

Sources/Recycling/GridViews/Ads/GridView/AdsGridView.swift
```diff
@@ -183,8 +184,22 @@ extension AdsGridView: UICollectionViewDataSource {
...
+        }
+
+        return imageCache.image(forKey: imagePath)
+    }
+
     public func adsGridViewCell(_ adsGridViewCell: AdsGridViewCell, loadImageForModel model: AdsGridViewModel, imageWidth: CGFloat, completion: @escaping ((AdsGridViewModel, UIImage?) -> Void)) {
-        dataSource?.adsGridView(self, loadImageForModel: model, imageWidth: imageWidth, completion: completion)
+        dataSource?.adsGridView(self, loadImageForModel: model, imageWidth: imageWidth, completion: { [weak self] model, image in
+            if let image = image, let imagePath = model.imagePath {
+                self?.imageCache.add(image, forKey: imagePath)
```
vadymmarkov: Don't we have some sort of image caching in the app?

Sources/Recycling/GridViews/Ads/Cell/AdsGridViewCell.swift
```diff
@@ -250,28 +252,47 @@ public class AdsGridViewCell: UICollectionViewCell {
...
+        let animations = { [weak self] in
+            self?.imageView.alpha = 1.0
+            self?.imageBackgroundView.backgroundColor = .clear
+        }
+
+        if animated {
+            imageView.alpha = 0.0
+            UIView.animate(withDuration: 0.2, delay: 0.0, options: [.curveEaseOut], animations: animations)
+        } else {
+            animations()
```
osanoj: If not `animated` then do `animations`... 🙃 

Maybe `animations` is not the best name for it? Not sure what is better though, but even `work` would probably be better

Sources/Recycling/GridViews/Ads/GridView/AdsGridView.swift
osanoj: Yes, I hope this would already be cached when used in the app

pimms: The network layer already caches the images if that's what you're thinking about. The problem I'm really solving here is that the cell needs to know whether the image was already loaded _(and as such we don't need to animate)_, or whether the image was fetched from the internet _(and was therefore probably not previously visible)_.

Sources/Recycling/GridViews/Ads/Cell/AdsGridViewCell.swift
pimms: Will fix!

Sources/Recycling/GridViews/Ads/GridView/AdsGridView.swift
osanoj: Aha, now I understand. So basically you want a way to check if the image should be set with animation... 🤔 

And you use a weak reference to the image so if it is expelled from other cache it would be nilled here too, so maybe all good then 🤔  

What do you think @vadymmarkov?

vadymmarkov: Yeah, we should be good I think. Then there's no need for using `NSCache` instead of dictionary @pimms

PR #415, opened 2019-05-08
Review control/button for AdManagement

robinsalehjan:
# Why?

- To improve the experience on the `AdManagement` screen

# What?

- A `UIControl` similar in look and feel to a `UIButton`
- A show/disappear animation
- Light haptic feedback on press

# Show me

Screenshot 2019-05-08 at 16 12 46


Sources/Components/Control/ReviewButtonControl/ReviewButtonControl.swift
```diff
@@ -0,0 +1,93 @@
...
+public final class ReviewButtonControl: UIControl {
+    // MARK: - Public
+
+    public weak var delegate: ReviewButtonControlDelegate?
+    public var text: String = "" {
+        didSet {
+            titleLabel.text = text
+        }
+    }
+    public static let cornerRadius: CGFloat = 8
```
nickfalk: Doesn't really need to be neither public nor static, right? 
(Probably doesn't need to be a  property either, since it seems to be used once, and only in the setup).

Sources/Components/Control/ReviewButtonControl/ReviewButtonControl.swift
```diff
@@ -0,0 +1,93 @@
...
+    public required init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+        setup()
+    }
+
+    // MARK: - Override
+
+    public override func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent?) {
+        super.touchesBegan(touches, with: event)
+        UIView.animate(withDuration: 0.0, animations: {
```
nickfalk: Setting the duration to 0.0 makes this a non-animating-animation™. Maybe you should just set the values and not bother with the `.animate` stuff(?)
(same thing in the `touchesEnded`, I guess). :-)

Sources/Components/ReviewButtonView/ReviewButtonView.swift
```diff
@@ -0,0 +1,100 @@
...
+
+public protocol ReviewButtonViewDelegate: AnyObject {
+    func reviewButtonView(_ reviewButtonView: ReviewButtonView, giveReviewWasTapped startReview: Bool)
+}
+
+public final class ReviewButtonView: UIView {
+    // MARK: - Public
+
+    public weak var delegate: ReviewButtonViewDelegate?
+    public var text: String = "" {
```
nickfalk: Since this class is `ReviewButtonView` and not `ReviewButton` it might make sense to rename this to `buttonText` or `buttonTitle`. I think it would make the intent of the api clearer.

Sources/Components/ReviewButtonView/ReviewButtonView.swift
```diff
@@ -0,0 +1,100 @@
...
+        didSet {
+            reviewButtonControl.text = text
+        }
+    }
+
+    // MARK: - Private
+
+    private let reviewButtonControlHeight: CGFloat = 48
+    private let viewHeight: CGFloat = 64
```
nickfalk: Wouldn't just `height` read better(?) since I guess `self` is the view in question. (Makes the constraint setup in `addToView` more readable, I believe.

Sources/Components/ReviewButtonView/ReviewButtonView.swift
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+        setup()
+    }
+
+    // MARK: - Public methods
+
+    public func addToView(_ view: UIView) {
```
nickfalk: Would it be possible to simply use `didlMoveToSuperview` instead?

robinsalehjan: That's a nice one, will fix it.

Sources/Components/ReviewButtonView/ReviewButtonView.swift
```diff
@@ -0,0 +1,106 @@
...
+
+    required init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+        setup()
+    }
+
+    // MARK: - Overrides
+
+    public override func didMoveToSuperview() {
+        super.didMoveToSuperview()
```
nickfalk: Very minor nitpick, but you don't really need to call the super-class' implementation, as the default implementation does nothing.

robinsalehjan: Noted and fixed👍

PR #416, opened 2019-05-08
Remove warnings

robinsalehjan:
# Why?

- Had a lot of ` public modifier is redundant in public extensions` warnings

# What?

- Remove `public` modifier where applicable
- Update `generate_image_assets_symbols` build phase script to not output a `public` modifier

# Show me

|     Before              |            After               |
| before | after |


robinsalehjan:
:pepesnuggles:

PR #418, opened 2019-05-26
Add contact form fullscreen view

vadymmarkov:
# Why?

This is a new UI component we're going to use in the FINN app.

# What?

- Add `ContactFormView` fullscreen view
- Support new text field type for phone numbers
- Adjust scroll view when the keyboard is shown

# Show me

contact


Sources/Components/TextField/TextField.swift
```diff
@@ -297,6 +303,10 @@ public class TextField: UIView {
         return evaluate("[A-Z0-9a-z._%+-]+@[A-Za-z0-9.-]+\\.[A-Za-z]{2,64}", with: emailAdress)
     }
 
+    fileprivate func isValidPhoneNumber(_ phoneNumber: String) -> Bool {
+        return evaluate("^[0-9]{8,11}$", with: phoneNumber)
```
bstien: Not sure how users might input their phonenumbers, but this doesn't seem to validate numbers prefixed with country code or with spacing in between.

Like:
- `+47xxxxxxxx`  ← 11 characters, but `+` isn't allowed according to regex.
- `0047xxxxxxxx` ← 12 characters, all digits.
- `xxx xx xxx`   ← The "usual" way of formatting mobile numbers.
- `xx xx xx xx`  ← The "usual" way of formatting landline numbers.

osanoj: Good observation, maybe this should be configurable too? Because I guess different backends might want this sent in different ways, maybe that we do the conversion to that format after, but if backend doesn't support `+` we should perhaps not allowed it as input either 🤔 

We should maybe ask what is allowed in this first use case.

vadymmarkov: I took `[0-9]{8,11}` from the webform this view is based on. But I think it's a good idea to make it configurable.

osanoj:
Maybe out of scope for this PR, but we should add support for setting `textContentType` on our TextField class... If we know it is password/email/phone number it is nice to get help filling it in 🙂

vadymmarkov:
@bstien @osanoj Made a few changes:
1) Set `textContentType` to enable auto suggest.
2) Changed phone number regex to allow whitespaces because they are usually present in auto suggest.
3) Added helper texts to show when email or phone number is not correct
4) Fix content offsets for iPhone 5

Simulator Screen Shot - iPhone 5s - 2019-05-27 at 14 10 23

PR #422, opened 2019-05-27
Fix/icon collection view

bstien:
# Why?
There were some issues with `IconCollectionView` when trying to use it on our objectPage. 

# What?
- Only use collectionView's width to calculate width of cells on iPhone.
- Realistic constraints/text for objectPage.


Sources/Components/IconCollection/IconCollectionView.swift
```diff
@@ -80,8 +80,8 @@ extension IconCollectionView: UICollectionViewDelegateFlowLayout {
             let width = max(collectionView.frame.width / CGFloat(viewModels.count), height * 2)
             return CGSize(width: width, height: height)
         } else {
-            let width = min(collectionView.frame.width, collectionView.frame.height)
```
vadymmarkov: I thought it was a smart thing to do for supporting landscape on phones. But now I see it wasn't 🙈

PR #423, opened 2019-05-27
YAMF™ -- aka. Yet Another Minor Fix (for IconCollectionView)

bstien:
# Why?
Seems like some of the ads can have _even more_ text placed into the property type than I originally thought.

# What?
- Set `minimumScaleFactor` to 0.75 for the cells' `titleLabel`.


vadymmarkov:
@bstien How does it look with minimumScaleFactor btw?

vadymmarkov:
I see on web they have two columns:

Screenshot 2019-05-27 at 15 05 52

bstien:
It would look like this:
Screenshot 2019-05-27 at 15 07 30

bstien:
Yeah, we should probably make it two lines, if it's needed. It looks much better than shrinking the font.

vadymmarkov:
We can merge this and I can add support for dynamic height in a separate PR if you like.

bstien:
I'm not sure how to do it, but figured I'd give it a try. Already working on it :)

bstien:
But I'll merge this one so we can have a separate PR

PR #424, opened 2019-05-27
 Icon collection view: support dynamic height

vadymmarkov:
# Why?

Because we want to support longer titles on cells.

# What?

- Support dynamic height for collection view cells to fit the content
- Top align collection view cells using custom collection view layout
- Resize `UICollectionView` to fit its content size because it shouldn't really be scrollable. Now there is no need to set the height for `IconCollectionView`

@bstien Feel free to commit to this PR if you see that I've missed anything 😉 

# Show me

Simulator Screen Shot - iPhone X - 2019-05-27 at 17 08 31


Sources/Components/IconCollection/IconCollectionViewFlowLayout.swift
```diff
@@ -0,0 +1,38 @@
...
+
+            for attribute in attributes {
+                let dy = (maxHeightY ?? attribute.frame.origin.y) - attribute.frame.origin.y
+                attribute.frame = attribute.frame.offsetBy(dx: 0, dy: dy)
+            }
+        }
+
+        return layoutAttributes
+    }
+}
```
pimms: I found this class to be especially well done 👍

PR #426, opened 2019-05-28
Fix/icon collection view system layout size fitting

bstien:
# Why?
We need the collectionView to calculate its own height before it's added to the view hierarchy. This is because of how we calculate the size of each view on our objectPage.

# What?
- Override method `systemLayoutSizeFitting(_:withHorizontalFittingPriority:verticalFittingPriority:)`

# Show me
_No UI._

### DIFF
--- a/Sources/Components/IconCollection/IconCollectionView.swift
+++ b/Sources/Components/IconCollection/IconCollectionView.swift
@@ -51,6 +51,19 @@ public final class IconCollectionView: UIView {
         addSubview(collectionView)
         collectionView.fillInSuperview()
     }
+
+    // MARK: - Overrides
+
+    // This override exists because of how we calculate view sizes in our objectPage.
+    // The objectPage needs to know the size of this view before it's added to the view hierarchy, aka. before
+    // the collectionView itself knows it's own contentSize, so we need to calculate the total height of the view manually.
+    //
+    // All we're given to answer this question is the width attribute in `targetSize`.
+    //
+    // This implementation may not work for any place other than the objectPage, because:
+    //   - it assumes `targetSize` contains an accurate targetWidth for this view.
+    //   - it ignores any potential targetHeight.
+    //   - it ignores both horizontal and vertical fitting priority.
     public override func systemLayoutSizeFitting(_ targetSize: CGSize, withHorizontalFittingPriority horizontalFittingPriority: UILayoutPriority, verticalFittingPriority: UILayoutPriority) -> CGSize {
         let targetWidth = targetSize.width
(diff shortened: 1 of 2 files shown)


Sources/Components/IconCollection/IconCollectionView.swift
```diff
@@ -51,6 +51,35 @@ public final class IconCollectionView: UIView {
...
+            let height = IconCollectionViewCell.height(for: viewModel, withWidth: width)
+            return CGSize(width: width, height: height)
+        }
+
+        guard let firstItem = cellSizes.first else { return targetSize }
+        let numberOfCellsInRow = Int(floor(targetWidth / firstItem.width))
+        let cellRows = cellSizes.chunked(into: numberOfCellsInRow)
+        let totalHeight = cellRows.compactMap { $0.max(by: { $0.height < $1.height }) }.reduce(0, { $0 + $1.height })
+
+        return CGSize(width: targetWidth, height: totalHeight + .mediumLargeSpacing)
```
vadymmarkov: Why do we need `.mediumLargeSpacing` here?

Sources/Components/IconCollection/IconCollectionView.swift
```diff
@@ -98,3 +132,11 @@ private final class CollectionView: UICollectionView {
         }
     }
 }
+
+private extension Array {
+    func chunked(into size: Int) -> [[Element]] {
+        return stride(from: 0, to: count, by: size).map {
+            Array(self[$0 ..< Swift.min($0 + size, count)])
```
vadymmarkov: Is there some sort of conflicting `min` func in `Array`?

Sources/Components/IconCollection/IconCollectionView.swift
bstien: Just to add some spacing to the bottom. A few cases I noticed had the collectionView "too close" to the ad's description.

Sources/Components/IconCollection/IconCollectionView.swift
bstien: I got this error. Couldn't figure out how to get around it, other than to use `Swift`.

> Use of 'min' nearly matches global function 'min' in module 'simd' rather than instance method 'min(by:)'

bstien:
Found a couple of new things we've overseen. We didn't account for the label's leading/trailing margin when calculating the cells height.

Also, I've added 8 px, `.mediumSpacing`, extra spacing per row.

| Before | After |
| --- | --- |
| Screenshot 2019-05-28 at 15 13 12 | Screenshot 2019-05-28 at 15 53 27 |

vadymmarkov:
Awesome 👍

PR #428, opened 2019-06-05
ContactFormView keyboard inset

bstien:
# Why?
The class `ContactFormView` handles insetting the content within its scrollView once a keyboard is presented. This worked on iPhones and iPads in fullscreen, but not on iPads when presented with `UIModalPresentationStyle.formSheet`.

The issue is the same as we had in our main app for the messaging popup on iPad, which was solved in PR#1383.

This solution also copies the `KeyboardInfo` struct that was included in PR#1437 into FinniversKit, but modifies it to be able to calculate the inset based on the view's frame within the window.

By using `KeyboardNotificationInfo` we should also be able to fix some of the issues we're having if the user has connected an external keyboard to his/hers iPad.

## Example
**Check if keyboard is presented or hidden.**
```swift
guard let keyboardInfo = KeyboardNotificationInfo(notification) else { return }

switch keyboardInfo.action {
case .willShow:
	print("Showing keyboard")
case .willHide:
	print("Hiding keyboard")
}
```

**Get the intersection between the keyboard and some view.**
```swift
guard let keyboardInfo = KeyboardNotificationInfo(notification) else { return }

let intersection = keyboardInfo.keyboardFrameEndIntersectHeight(inView: someView)
```

**Animate something alongside the keyboard.**
```swift
guard let keyboardInfo = KeyboardNotificationInfo(notification) else { return }

let intersection = keyboardInfo.keyboardFrameEndIntersectHeight(inView: someView)

UIView.animateAlongsideKeyboard(keyboardInfo: keyboardInfo) {
	switch keyboardInfo.action {
	case .willShow:
		someView.transform = CGAffineTransform(translationX: 0, y: -intersection)
	case .willHide:
		someView.transform = .identity
	}
}
```

# What?
- Added struct `KeyboardNotificationInfo` that handles keyboard notifications.
- Added extensions to `UIView` for animating alongside a keyboard animation.
- Updated `ContactFormView` to use these two things.

# Show me
| Before | After |
| --- | --- |
| contactform-ipad-before | contactform-ipad-after |

### DIFF
--- a/Sources/Extensions/UIKit/UIViewExtensions.swift
+++ b/Sources/Extensions/UIKit/UIViewExtensions.swift
@@ -43,3 +43,18 @@ public extension UIView {
         layer.rasterizationScale = UIScreen.main.scale
     }
 }
+
+// MARK: - Animate alongside keyboard
+
+public extension UIView {
+    class func animateAlongsideKeyboard(keyboardInfo: KeyboardNotificationInfo, animations: @escaping () -> Void, completion: ((Bool) -> Void)? = nil) {
+        let animationDuration = keyboardInfo.animationDuration
+        let animationOptions = keyboardInfo.animationOptions
+
+        UIView.animate(withDuration: animationDuration, delay: 0, options: animationOptions, animations: animations, completion: completion)
+    }
+
+    class func animateAlongsideKeyboard(keyboardInfo: KeyboardNotificationInfo, animations: @escaping () -> Void) {
+        UIView.animateAlongsideKeyboard(keyboardInfo: keyboardInfo, animations: animations, completion: nil)
+    }
+}
--- a/Sources/Extensions/UIKit/UIViewExtensions.swift
+++ b/Sources/Extensions/UIKit/UIViewExtensions.swift
@@ -51,10 +51,10 @@ public extension UIView {
         let animationDuration = keyboardInfo.animationDuration
         let animationOptions = keyboardInfo.animationOptions
 
-        UIView.animate(withDuration: animationDuration, delay: 0, options: animationOptions, animations: animations, completion: completion)
+        animate(withDuration: animationDuration, delay: 0, options: animationOptions, animations: animations, completion: completion)
     }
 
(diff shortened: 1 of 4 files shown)


Sources/Util/KeyboardNotificationInfo.swift
```diff
@@ -0,0 +1,57 @@
...
+    /// Calculate the total intersection between the keyboard and a view. The value returned
+    /// will be between 0 and whatever intersection occurs. This is in case the user has
+    /// an iPad with an external keyboard connected, which would've returned a negative value.
+    public func keyboardFrameEndIntersectHeight(inView view: UIView) -> CGFloat {
+        guard let windowHeight = view.window?.frame.size.height, let frameEnd = frameEnd else { return 0 }
+        let keyboardHeight = frameEnd.size.height
+        let frameInWindow = view.convert(view.bounds, to: nil)
+
+        // Calculate the intersection between the keyboard and this view.
+        let intersection = keyboardHeight - (windowHeight - frameInWindow.origin.y - frameInWindow.size.height)
```
shredlocker: CGRect has this function `intersection(_:)` which is quite useful.

bstien: Didn't think of that one😂 Thanks!

Sources/Extensions/UIKit/UIViewExtensions.swift
```diff
@@ -43,3 +43,18 @@ public extension UIView {
...
 }
+
+// MARK: - Animate alongside keyboard
+
+public extension UIView {
+    class func animateAlongsideKeyboard(keyboardInfo: KeyboardNotificationInfo, animations: @escaping () -> Void, completion: ((Bool) -> Void)? = nil) {
+        let animationDuration = keyboardInfo.animationDuration
+        let animationOptions = keyboardInfo.animationOptions
+
+        UIView.animate(withDuration: animationDuration, delay: 0, options: animationOptions, animations: animations, completion: completion)
```
osanoj: I guess you could skip `UIView.` here, but no biggie

bstien: You're right, I'll change it👌

PR #429, opened 2019-06-14
Empty view: stop motion manager when view is not visible

vadymmarkov:
# Why?

Because we don't want to monitor accelerometer and device motion updates while the empty view is not visible.

# What?

- Stop motion manager and remove all behaviors from animator when view is not visible
- Re-start motion manager and add necessary behaviors when view becomes visible

# Show me

No UI changes.


vadymmarkov:
Yes, I think it's what @osanoj and @bstien discovered while looking at battery usage.

PR #433, opened 2019-06-17
Interest form disclaimer view

bstien:
# Why?
We need a disclaimer at the bottom of the contact form for `realestate-planned`, which also should contain a button that lets the user read more about the T&C.

# What?
- Added `DisclaimerView` + demo.
- Added `DisclaimerView` to `ContactFormView`.
- Removed label `accessoryText` from both the view and the viewModel.

# Show me
| DisclaimerView |
| --- |
| Screenshot 2019-06-17 at 14 33 36 |

| Before | After |
| --- | --- |
| Screenshot 2019-06-17 at 15 35 30 | Screenshot 2019-06-17 at 14 16 21 |


Demo/Components/Disclaimer/DisclaimerDemoView.swift
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public required init?(coder aDecoder: NSCoder) { fatalError() }
+
+    private func setup() {
+        addSubview(disclaimerView)
+
+        NSLayoutConstraint.activate([
+            disclaimerView.topAnchor.constraint(equalTo: topAnchor, constant: .largeSpacing),
+            disclaimerView.centerXAnchor.constraint(equalTo: centerXAnchor),
```
vadymmarkov: I guess `centerXAnchor` is not necessary here, is it?

PR #436, opened 2019-06-26
The BottomSheetDelegate can now prevent the sheet from being dismissed

pimms:
# Why?

I'm redoing the contact-form _(the first message sent from the buyer to the seller)_ in a `BottomSheet`, and need a way to prevent to the user from accidentally dismissing the `BottomSheet` when they are in the middle of writing a message.

# What?

The `BottomSheetPresentationControllerDelegate` and `BottomSheetDelegate` has two new methods — `...CanDismiss() -> Bool` and `...DidAttemptToDismiss()`.

When the user performs a dismissive action, the `BottomSheetPresentationController` will consult its delegate to check if this is supported. The idea is that the delegate will return `false` if the presented content contains unsaved/uncommitted changes.

### Minor QoL improvements

 - Added proxy-property `isOn` to `SwitchView`
 - Added the struct `SwitchViewDefaultViewModel` for easier usage of `SwitchView`

### Other

The `BottomSheetDynamics` demo contains a `SwitchView` toggling whether dismissals initiated by the `BottomSheet` must be confirmed by the user.

# Show me

**Confirmation requirement on**
confirm-dismiss-on

**Confirmation requirement off** _(default behavior)_
confirm-dismiss-off


Demo/Components/BottomSheet/BottomSheetMechanicsDemoViewController.swift
```diff
@@ -246,6 +259,27 @@ extension BottomSheetMechanicsDemoViewController: RootViewControllerDelegate {
 }
 
 extension BottomSheetMechanicsDemoViewController: BottomSheetDelegate {
+    func bottomSheetCanDismiss(_ bottomSheet: BottomSheet) -> Bool {
```
shredlocker: Maybe change this to `bottomSheetShouldDismiss`? I think that is Apple's convention 🍎

Demo/Components/BottomSheet/BottomSheetMechanicsDemoViewController.swift
```diff
@@ -246,6 +259,27 @@ extension BottomSheetMechanicsDemoViewController: RootViewControllerDelegate {
...
+        return !requireConfirmationOnDragSwitch.isOn
+    }
+
+    func bottomSheetDidAttemptToDismiss(_ bottomSheet: BottomSheet) {
```
shredlocker: And maybe this to `bottomSheetDidCancelDismiss`?

pimms:
@shredlocker @vadymmarkov Should I add default implementations for the new `BottomSheetDelegate` methods? Otherwise we'll end up in an awkward unbuildable state in Charcoal and FINN for a while.

vadymmarkov:
Not a big fan of default implementations for protocols. I think it would be better if we update `Charcoal` first.

pimms:
All right, I'll spin up a PR for this later today 👍

PR #437, opened 2019-06-26
Bumped version to 1.1.2

pimms:
# Why?

I forgot to bump the version in my last PR.


osanoj:
Isn't this actually a breaking change? Or at least a new feature?

pimms:
Maybe so yeah — I'll bump to 1.2.0 instead. It feels way to minor to bump to 2.0.0, even though it's technically slightly breaking.

pimms:
Not gonna bother waiting for the build :P

PR #438, opened 2019-06-28
Carousel View

shredlocker:
# Why?

We want to have the ability to scroll from the last image in the gallery to the first, as well as from the first to the last.

# What?

This PR introduces the `CarouselView`, which is a wrapper around the `UICollectionView`. It works by adding two more items to the collection. This means that the items are shifted and the index paths of the collection view doesn't match up the data array. The overridden methods corrects the index paths and the content offset.

carousel

When ever the collection view is about to scroll to the first item on the left, it adjusts the scroll view content offset and the target content offset to the other end of the collection. This creates a smooth animation and gives the illusion of an infinite collection view. The GIF below contains the scroll indicator to make this clear.

# Show me

carousel-view


FinniversKit.xcodeproj/project.pbxproj
```diff
@@ -2030,18 +2037,19 @@
 		CF8F021421DCC5B2004C4042 /* Components */ = {
 			isa = PBXGroup;
 			children = (
-				CFB61377229ACB6200DE283D /* IconCollection */,
 				083349EA21EF24AE000E68D3 /* BottomSheet */,
 				CF8F021D21DCC691004C4042 /* Broadcast */,
 				CF8F021E21DCC6A0004C4042 /* Button */,
 				CFCDDA5D22731D4D005E1F4A /* Callout */,
+				DAE738C522C4C84C001C69E3 /* CarouselVIew */,
```
osanoj: That capital `I` is a bit out of place there I think 😄

Demo/Components/CarouselVIew/CarouselViewDemoCell.swift
```diff
@@ -0,0 +1,48 @@
...
+    var text: String? {
+        get { return label.text }
+        set { label.text = newValue }
+    }
+
+    private lazy var label: UILabel = {
+        let label = UILabel(frame: .zero)
+        label.font = .systemFont(ofSize: 48, weight: .bold)
+        label.textColor = .white
+        label.backgroundColor = UIColor(hue: 0.6, saturation: 0.3, brightness: 0.6, alpha: 1.0)
```
vadymmarkov: Why don't you just use some system or FINN color here? Just asking 😄

Sources/Components/CarouselView/CarouselView.swift
```diff
@@ -0,0 +1,192 @@
...
+public typealias CarouselViewCell = UICollectionViewCell
+
+public protocol CarouselViewDataSource: AnyObject {
+    func numberOfItems(in carouselView: CarouselView) -> Int
+    func carouselView(_ carouselView: CarouselView, cellForItemAt indexPath: IndexPath) -> CarouselViewCell
+}
+
+public protocol CarouselViewDelegate: AnyObject {
+    func carouselView(_ carouselView: CarouselView, didSelectItemAt indexPath: IndexPath)
+    func carouselViewDidEndDecelerating(_ carouselView: CarouselView)
```
vadymmarkov: Should this function be required in protocol implementations?

Sources/Components/CarouselView/CarouselView.swift
```diff
@@ -0,0 +1,192 @@
...
+        didSet {
+            collectionView.reloadData()
+        }
+    }
+
+    public weak var delegate: CarouselViewDelegate?
+
+    // MARK: - Private Properties
+
+    private var initialLayout = true
```
vadymmarkov: Should we change this to more "bool-friendly" name, like for example `shouldSetInitialLayout` or something like that?

Sources/Components/CarouselView/CarouselView.swift
```diff
@@ -0,0 +1,192 @@
...
+    func register(_ cellClass: CarouselViewCell.Type) {
+        collectionView.register(cellClass)
+    }
+
+    func dequeue<T>(_ cellType: T.Type, for indexPath: IndexPath) -> T where T: CarouselViewCell {
+        return collectionView.dequeue(cellType, for: indexPath)
+    }
+
+    func scrollToItem(at indexPath: IndexPath, at scrollPosition: UICollectionView.ScrollPosition, animated: Bool) {
+        collectionView.scrollToItem(at: IndexPath(item: indexPath.item + 1, section: 0), at: scrollPosition, animated: animated)
```
vadymmarkov: Since this function is public, I guess it's possible to crash the app when attempting to scroll to invalid index path. Perhaps some index path validation is needed here.

FinniversKit.xcodeproj/project.pbxproj
shredlocker: Haha, I fix

Sources/Components/CarouselView/CarouselView.swift
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    public func scrollViewDidEndDecelerating(_ scrollView: UIScrollView) {
+        delegate?.carouselViewDidEndDecelerating(self)
+    }
+
+    public func scrollViewWillEndDragging(_ scrollView: UIScrollView, withVelocity velocity: CGPoint, targetContentOffset: UnsafeMutablePointer<CGPoint>) {
+        let numberOfItems = collectionView.numberOfItems(inSection: 0)
+
+        if scrollView.isPagingEnabled {
```
vadymmarkov: We could use `guard` here for early return.

Sources/Components/CarouselView/CarouselView.swift
```diff
@@ -0,0 +1,192 @@
...
+                )
+            }
+        }
+    }
+}
+
+// MARK: - Private Functions
+
+private extension CarouselView {
+    func indexPath(forItem item: Int) -> IndexPath {
```
vadymmarkov: What would happen if the number of items is 0?

Sources/Components/CarouselView/CarouselView.swift
```diff
@@ -0,0 +1,192 @@
...
+        guard let dataSource = dataSource else {
+            return 0
+        }
+
+        return dataSource.numberOfItems(in: self) + 2
+    }
+
+    public func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
+        guard let dataSource = dataSource else {
+            preconditionFailure("Data source is not available")
```
vadymmarkov: We could avoid doing this by having `dataSource` as a parameter in `init` method.

Demo/Components/CarouselVIew/CarouselViewDemoCell.swift
shredlocker: No, reason. I just like the colors with saturation 0.3 and brightness 0.6. I can use a FINN color here 😉

Sources/Components/CarouselView/CarouselView.swift
shredlocker: We do use this method in the `AdPageGalleryView`, that is why I added it. Should I make an empty extension?

Sources/Components/CarouselView/CarouselView.swift
shredlocker: I like it!

Sources/Components/CarouselView/CarouselView.swift
shredlocker: That is probably a good idea.

Sources/Components/CarouselView/CarouselView.swift
shredlocker: You are right

Sources/Components/CarouselView/CarouselView.swift
shredlocker: Something would be definitely break 😅 This carousel view doesn't make sense unless there are more than one data entry so I can return the same item is number of items is not sufficient.

Sources/Components/CarouselView/CarouselView.swift
shredlocker: `dataSource` would still be optional though. What if we do `collectionView.dataSource = self // or nil` in the `didSet` method of `dataSource`? Then this would never occur.

Sources/Components/CarouselView/CarouselView.swift
vadymmarkov: No, I think it's better to keep it then.

Sources/Components/CarouselView/CarouselView.swift
vadymmarkov: What if we make it `private weak var dataSource: CarouselViewDataSource?`, so you can inject it only during initialization of `CarouselView`.

shredlocker: Yep, that sounds good

Sources/Components/CarouselView/CarouselView.swift
```diff
@@ -0,0 +1,203 @@
...
+                y: scrollView.contentOffset.y
+            )
+        }
+    }
+}
+
+// MARK: - Private Functions
+
+private extension CarouselView {
+    func indexPath(forItem item: Int) -> IndexPath {
```
pimms: I don't understand what this method is doing — is `item` 1-indexed? And why is the returned value called `next`?

pimms: Oh, now I understand — it translates the Index Paths from the underlying `UICollectionView` to the logical indices used by the carousel view. Could we rename the method to better reflect this?

```
func translate(indexPath: IndexPath) -> IndexPath
func convertToCarouselIndexPath(_: IndexPath) -> IndexPath
func translateToCarousel(indexPath: IndexPath) -> IndexPath
```

Sources/Components/CarouselView/CarouselView.swift
```diff
@@ -0,0 +1,203 @@
...
+        let numberOfItems = collectionView.numberOfItems(inSection: 0)
+
+        guard numberOfItems > 1 else {
+            return IndexPath(item: item, section: 0)
+        }
+
+        let next: Int
```
pimms: Ref comments above, I think we should use a different name for this guy — `converted`, `translated`, `carouselIndex` for instance.

Sources/Components/CarouselView/CarouselView.swift
shredlocker: I like the `translate(_ indexPath: IndexPath) -> IndexPath`. Changing it!

Sources/Components/CarouselView/CarouselView.swift
shredlocker: I'll change it

PR #440, opened 2019-06-28
AdsGridViewDataSource and MarketsGridViewDataSource can return nil model

osanoj:
# Why?
Trying to fix some crashes which I suspect come from timing issues related to updating the data models.

# What?
Add option to return nil instead of model object for AdsGridViewDataSource and MarketsGridViewDataSource. Better to not show anything or a non-configured cell for a short time (hopefully) than to crash.


shredlocker:
I must say that I'm not a big fan of using the `[safe:]` subscript in these situations. What this tells us is that there is something wrong with the code somewhere, but instead of solving the problem we simply hide it. If there are timing issues we should find them and fix them.

shredlocker:
I see that we reload recommendations in the `reloadMarkets()` and vice versa in `reloadAds()` in `FrontPageView`. Maybe this can cause the timing issues?

osanoj:
I of course agree that the best is to fix the source of issues. But something that is "impossible" to reproduce is hard to fix (and not always worth spending a lot of time on).

The issue is that the data objects in the datasource can change while the collectionview is doing "something". I think this issue most likely is due to reloads because user logging out or such things, so not directly related to this view.

pimms:
Has either of you investigated this further? I agree with Henrik that we shouldn't "wipe the error under the rug" — unless we confirm that there's a ridiculously complicated race goin on.... 🙂 I think we should have good enough control over our timings to at least be able to identify potential races.

Then again, this issue is making the app crash **now** 🙃

osanoj:
The issue has been making the app crash for at least a year I think. But it is not many crashes so it is not something worth spending weeks on trying to reproduce and debug...

PR #441, opened 2019-07-01
Message Form redesign + predefined message templates

pimms:
# Why?

We're going to run an experiment with message-templates _(pre-defined messages)_ in the contact form. The message templates required a little UI rework, and in case the experiment is a dud, we'll have a refreshed & redesigned contact form 🎊 🍾 

_(Note: while we refer to the initial message view as "contact form", that name was taken in FinniversKit, and I named this view `MessageForm`)_

# What?

 - The Message Form has been rewritten from scratch in FinniversKit
 - An optional keyboard extension showing pre-defined messages can be configured by the presenter
 - The Message Form is presented as a bottom-sheet
   - If the input field is dirty, the sheet can't be dismissed without explicit user interaction

The form takes _(in addition to localizables & templates)_ one configuration; whether or not to show the template toolbar.

On submit, the form will notify its' delegate of the entered message & the template state — whether the user is sending:
 - a message template, without modification _(`template`)_
 - a message template, with addition _(`modifiedTemplate`)_
 - a custom message _(or subtracted template)_ _(`custom`)_

The `MessageFormBottomSheet` is stand-alone, and requires no integration. It must be handled by an object whose lifecycle supersedes its own _(see `MessageFormDemoPresenter` for an example)_ which feels awkward, but this should integrate without issue in the FINN app.

# Show me

|                | Expanded | Collapsed |
|----------------|----------|-----------|
| iPad Pro 2018  |Screen Shot 2019-07-01 at 10 00 26      | Screen Shot 2019-07-01 at 10 00 31    |
| iPad (classic) |Screen Shot 2019-07-01 at 10 00 43|Screen Shot 2019-07-01 at 10 00 46|
| iPhone 8       |Screen Shot 2019-07-01 at 10 00 49|Screen Shot 2019-07-01 at 10 00 51|
| iPhone X       |Screen Shot 2019-07-01 at 10 00 55|Screen Shot 2019-07-01 at 10 00 58|
| iPhone XS Max  |Screen Shot 2019-07-01 at 10 01 02|Screen Shot 2019-07-01 at 10 01 05|

| Expanded (without templates) | Collapsed (without templates) |
|-------------------------------|-------------------------------|
|Screen Shot 2019-07-01 at 10 32 31|Screen Shot 2019-07-01 at 10 32 34|

### Dismissal Handling

The `MessageFormBottomSheet` refuses to be dismissed unless the text-input form is empty.

dismiss-iphone

### Replace Handling

Selecting a template prompts the user if they really want to replace the content that's already entered into the input field. The exception is if the text-field only contains another template.

replace-iphone

### Template States

templateStates


Sources/Fullscreen/MessageForm/MessageFormBottomSheet.swift
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// MARK: - Private extensions
+
+private extension BottomSheet.Height {
+    static var messageFormHeight: BottomSheet.Height {
+        let screenSize = UIScreen.main.bounds.size
+
+        if screenSize.height <= 568 {
+            return BottomSheet.Height(compact: 510, expanded: 510)
```
osanoj: Why this number?

Sources/Fullscreen/MessageForm/MessageFormView.swift
```diff
@@ -0,0 +1,105 @@
...
+        setup()
+    }
+
+    private func setup() {
+        addSubview(textView)
+        addSubview(transparencyLabel)
+
+        // Adding a low-priority "infinite" height constraint to the TextView
+        // makes sure it gets chosen to fill the available vertical space.
+        let textViewHeightConstraint = textView.heightAnchor.constraint(lessThanOrEqualToConstant: 3000)
```
osanoj: Can you try content hugging etc instead?

pimms: `label.setContentHuggingPriority(.defaultHigh, for: .vertical)` worked wonders. I must've done the layout for this class while sleeping, because I was using another constraint in addition to this to enforce the constraints.

osanoj: Looks a lot nicer with the new commit, good job! 😄

PR #443, opened 2019-07-03
Implement QuestionnaireView with basic styling

robinsalehjan:
# Why?

- We would like to have a non-intrusive way of asking our users if they would like to participate in a questionnaire.

# What?

- Implement the `QuestionnaireView` and basic styling.

# Show me
Simulator Screen Shot - iPhone X - 2019-07-03 at 09 56 33


osanoj:
Would it be possible to extend the styles of the `Infobox` component to also meet the requirements for this?

robinsalehjan:
It is possible, but then the secondary button and the close button has to be optional in the `InfoBox` component.

I decided to make into a separate component due to the fact that the `QuestionnaireView` should always be cancellable, but the `InfoBox` doesn't necessarily need that.

osanoj:
Hmm, OK, good that you were aware it was there. I'm slightly concerned that we are getting a lot of almost identical views...

Don't necessarily think it would be a problem for `Infobox` to have optional close and secondary buttons. How to name the different styles might be a bigger issue (for instance that title text is smaller than detail text here). It would be a breaking change to update the model class, but that shouldn't be a big problem.

robinsalehjan:
I agree with you and see your point. For this specific case they may look identical, however I do think that it's fair to separate the `InfoView` and the `QuestionnaireView` apart

What if we would like to present a larger `QuestionnaireView` where the title and detail are inverted in sizes or if we would like to present a couple of questions within the app instead of linking out to an external form?

osanoj:
Let's come back to issue of merging the "boxes" if we get more examples 👍

PR #447, opened 2019-07-05
ToastView now covers the safeAreaInsets

pimms:
# Why?

The `ToastView` was anchored to the safe bottom-anchor, which caused an unwanted gap when presented from the bottom on a device with rounded corners.

# What?

The `ToastView` now animated relative to the view's `bottomAnchor`. In order to expand to the same height as it used to, it adds a view to cover the `safeAreaInsets.bottom` _(if the inset is greater than zero)_.

The demo now requires user interaction to trigger the animated toasts, and demonstrates the animation with and without the cover-view.

# Show me

### Before

before

### After

after

In app:

Screen Shot 2019-07-05 at 12 37 33


shredlocker:
Does it work to constrain the `messageTitle` to `compatibleBottomAnchor` instead? Then you don't need the cover view.

pimms:
Literally just had the same thought, this was a stupid way of achieving something simple :P Will fix.

osanoj:
I think you would need to have additional constraints on the height or bottomAnchor of the ToastView then, to make sure it covers all the way to the real bottom considering it now uses a transform based on the height to animate into view. Probably the animation would need to redone to use constraints instead.

pimms:
Hmm yeah, I'm not sure if the solution I went with was as stupid as I thought after all. The solution I've done so far is clear and explicit, and changing constraint constants and/or animating based on constraints would probably not be as obvious.

I'm kind of torn — what do you think @osanoj ?

pimms:
Discussed with Jonas, merging as-is as we're under the impression that doing this with constraints & without the extra view will make the existing logic more complicated & _"denser"_ than it needs to be.

PR #448, opened 2019-07-05
Add favorite folder list view

vadymmarkov:
# Why?

There is a new design for the list of favorite folders that we want to implement for our iOS app.

# What?

- Implement `FavoritesListView` according to design. The remaining tasks are to show an empty view when there are no hits on search and the edit mode for folder deletion

- Improve `RemoteImageTableViewCell` by setting images from cache if it's present

- Don't constraint views in demo to the bottom safe area when it's needed, e.g. when we want to show a list in the bottom sheet.

@bstien Feel free to fix comments and merge this PR when it's approved 😉 

# Show me
folderlist_changes


Sources/Recycling/ListViews/FavoriteFolders/FavoriteFoldersListView.swift
```diff
@@ -0,0 +1,296 @@
...
+    public func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        tableView.deselectRow(at: indexPath, animated: false)
+
+        guard let section = Section(rawValue: indexPath.section) else { return }
+
+        switch section {
+        case .addButton:
+            delegate?.favoriteFoldersListViewDidSelectAddButton(self)
+        case .folders:
+            delegate?.favoriteFoldersListView(self, didSelectItemAtIndex: indexPath.row)
```
bstien: Note to self _(since I'm supposed to make changes to this PR)_:

We should probably send the view model instead, seeing as the list of folders may be filtered, which the delegate probably doesn't know about.

Sources/Recycling/ListViews/FavoriteFolders/FavoriteFoldersListView.swift
```diff
@@ -0,0 +1,296 @@
...
+            footerView.heightAnchor.constraint(equalToConstant: footerHeight)
+        ])
+    }
+}
+
+// MARK: - UITableViewDataSource
+
+extension FavoriteFoldersListView: UITableViewDataSource {
+    public func numberOfSections(in tableView: UITableView) -> Int {
+        return Section.allCases.count
```
pimms: Never thought of handling multiple sections in a Table/Collection view this way before, that's really clever 👍 👍

PR #450, opened 2019-07-09
KeyboardNotificationInfo adjusts for portion of view that's out of bounds

pimms:
# Why?

The method `KeyboardNotificationInfo.keyboardFrameEndIntersectHeight(inView:)` did not take into account that the `view` might extend below the bottom of the screen. This turned out to be an issue with `BottomSheet`, which does not shrink vertically when the `BottomSheet` is initially presented with the device in portrait mode and then rotated to landscape. The opposite is true — when initially presented in landscape, the `BottomSheet` expands when the device is rotated to portrait.

# What?

- Added unit tests to prove the presence & subsequent resolve of this bug
- Adjusting the intersection height to account for the portion of the view that's out of bounds
- Bumped version to 1.4.1

# Show me

### Before

before

### After

after

### DIFF
--- a/FinniversKit.podspec
+++ b/FinniversKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = 'FinniversKit'
-  s.version      = '1.4.0'
+  s.version      = '1.4.1'
   s.summary      = "FINN's iOS Components"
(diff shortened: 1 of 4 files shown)


bstien:
Awesome! Great job 🙌

PR #451, opened 2019-07-09
Added property to enable/disable text input in MessageForm

pimms:
# Why?

When showing a `LoadingView` while sending a message in the FINN app, the keyboard can still be interacted with.

# What?

Added a property to `MessageFormView` _(proxied through `MessageFormViewController` and `MessageFormBottomSheet`)_ that handles the first responder status and disables user-interaction on the `TextView`.

Bumped version to 1.4.2.

# Show me

Integrated in the FINN app:

### After

Internet is disconnected for the first couple of attempts, then enabled when the message gets sent, then I disabled it again for the last attempt.

iphone8-failure-2

### DIFF
--- a/Sources/Fullscreen/MessageForm/MessageFormView.swift
+++ b/Sources/Fullscreen/MessageForm/MessageFormView.swift
@@ -41,6 +41,16 @@ class MessageFormView: UIView {
         }
     }
 
+    var inputEnabled: Bool = true {
+        didSet {
+            if inputEnabled {
+                _ = becomeFirstResponder()
+            } else {
+                _ = resignFirstResponder()
+            }
+        }
+    }
+
     // MARK: - Private properties
 
     private let viewModel: MessageFormViewModel
@@ -77,6 +87,10 @@ class MessageFormView: UIView {
 
     public override func becomeFirstResponder() -> Bool {
         return textView.becomeFirstResponder()
+    }
+
+    public override func resignFirstResponder() -> Bool {
+        return textView.resignFirstResponder()
     }
 }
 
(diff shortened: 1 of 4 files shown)


Sources/Fullscreen/MessageForm/MessageFormView.swift
```diff
@@ -41,6 +41,17 @@ class MessageFormView: UIView {
...
     }
 
+    var inputEnabled: Bool = true {
+        didSet {
+            textView.isUserInteractionEnabled = inputEnabled
+            if inputEnabled {
+                _ = becomeFirstResponder()
+            } else {
+                _ = resignFirstResponder()
+            }
```
bstien: Just to get rid of one of the `_`s:

```swift
inputEnabled ? becomeFirstResponder() : resignFirstResponder()
```

Or do you get a complaint from Xcode/SwiftLint if you drop it?

pimms: Seems to go through without warning

bstien: Awesome 👍
Otherwise, just add `@discardableResult` to both methods.

bstien: You should probably add the annotation anyways since they're public methods. I don't think we usually care about the result.

pimms: Ya, kind of prefer this style:
```
            if inputEnabled {
                becomeFirstResponder()
            } else {
                resignFirstResponder()
            }
```

The ternary operator implies to me that we're _first and foremost_ producing a result, but we really aren't in this particular case.

PR #452, opened 2019-07-09
Additions to favorite folder list view

bstien:
**This PR is against @vadymmarkov's branch in #448.**

The changes within this PR is done in collaboration with Thuy and it is ready to be merged in to the other branch.

# What?
- Remove the cancel button within the searchBar.
- Dismiss keyboard on tableView drag.
- Don't hide footerView.
- Scroll tableView to top before reloading it.
- Resign keyboard on search button press.
- Always show footerView once a search is active.
- Make footerView move 1.5 times as fast when scrolling the list.
- Removed delegate method.
- Remove top padding in searchBar, making it 36pt height + 8pt bottom padding. 44pt total.
- Don't show footerView when the tableView's content is smaller than its frame plus the default cellHeight and a search isn't active.
- Make footerView have same shadow as the searchBar.

# Show me

| Before | After |
| --- | :---: |
| folderlist_dismissKeyboard_before | folderlist_changes |


bstien:
Had a discussion with Thuy and got this approved. Added some more changes and updated the PR description.

Sources/Recycling/ListViews/FavoriteFolders/FavoriteFoldersListView.swift
```diff
@@ -274,23 +275,61 @@ extension FavoriteFoldersListView: UIScrollViewDelegate {
...
 }
+
+// MARK: - UISearchBar extension
+
+private extension UISearchBar {
+
+    /// The cancel button is automatically enabled/disabled based on the searchBar's firstResponder status.
+    /// We want to enable it when it's shown, no matter if the searchBar has focus or not.
+    func enableCancelButton() {
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.001, execute: { [weak self] in
```
pimms: Why is this scheduled 1ms in the future?

bstien: Yesterday the button was disabled if I only used `DispatchQueue.main.async`, but it worked if I gave it some delay.

Was going to make a gif showing this, but somehow it works now without delay. No clue why 🤔

Will remove the delay 👍

Sources/Recycling/ListViews/FavoriteFolders/FavoriteFoldersListView.swift
```diff
@@ -274,23 +275,61 @@ extension FavoriteFoldersListView: UIScrollViewDelegate {
...
+            guard let self = self else { return }
+            for view in self.subviews {
+                for subview in view.subviews {
+                    if let button = subview as? UIButton {
+                        button.isEnabled = true
```
osanoj: This will enable any button inside the search bar to be enabled? I think this is quite dangerous since apple might add another button at any time, or move the cancel button inside something else. 

Is this really something we have to have? Because we don't have an API for doing this so I would really like to avoid it...

Sources/Recycling/ListViews/FavoriteFolders/FavoriteFoldersListView.swift
pimms: Sounds to me like a fugly potential race, does it **always** ™️ work?

bstien: Nevermind this one. We won't be showing the cancel button in the searchBar at all.

Sources/Recycling/ListViews/FavoriteFolders/FavoriteFoldersListView.swift
pimms: I think that if we're going to do this, we should do it fully recursively in case the `UIButton` is moved one lever deeper by  at some point:

https://stackoverflow.com/a/46697075
```
    func enableButtons(_ view:UIView) {
        for subView in view.subviews {
            enableButtons(subView)
        }
        if let buttonView = view as? UIButton {
            buttonView.isEnabled = true
        }
    }
```

pimms: Also, still not digging the need for doing this asynchronously — I'm willing to bet that we'll encounter a race before issues with view hierarchy changes... 🙂

bstien:
Added a few more commits and updated the PR description 👌

bstien:
I've removed the `WIP` label and updated the description/gif. I think it's ready to be merged into the other branch 🙌

PR #453, opened 2019-07-11
Add new login entry view

fespinoza:
**Update**: the design is confirmed with thuy and the test was added. This PR is ready, but i would wait to merge after I work on the integration too, to ensure there is no unforeseen extra view requirements

# Why?

New design and text that better explain why the user should log in on each of the pages you are asked to log in.

# What?

Added a new full screen page and its demo view. It's my first PR doing an addition to the project so please check if I'm following the conventions correctly and so on.

I made sure the screen looks great on iPad too.

# Show me

**Note**: the "before" screen is not in FinniversKit

| Before | After |
| --- | :---: |
| IMG_4173 | IMG_4172 |

# Extra

- ~~I'm not sure about the 2: login and register buttons, as SPiD has a single UI for both flows. I need to double check with Thuy.~~
- ~~I will add a screenshot test for this view too, once I make sure the final design of the UI.~~
- ~~Even though this is a "WIP" PR, i would like a review for it.~~


Demo/Fullscreen/LoginEntry/LoginEntryViewDemoView.swift
```diff
@@ -0,0 +1,27 @@
+//
+//  Copyright © 2019 FINN AS. All rights reserved.
+//
+
+import FinniversKit
+
+public class LoginEntryViewDemoView: UIView {
+    private lazy var loginEntryView: LoginEntryView = {
+        let loginEntryView = LoginEntryView()
+        loginEntryView.translatesAutoresizingMaskIntoConstraints = false
```
bstien: You could use the FinniversKit initializer instead for `UIView`'s. That saves you the annoying line `x.translatesAutoresizingMaskIntoConstraints = false` 🎉

I think it applies to every view we have in FK, except for `Label` and `Button` which has their own init.

I.e.:
```swift
let loginEntryView = LoginEntryView(withAutoLayout: true)
```

fespinoza: 👍

fespinoza:
This is ready @bstien, can you review it? (as it was approved before the last changes)

Sources/Fullscreen/LoginEntryView/LoginEntryDialogueView.swift
```diff
@@ -0,0 +1,129 @@
+//
+//  Copyright © 2019 FINN AS. All rights reserved.
+//
+
+import UIKit
+
+public class LoginEntryDialogueView: UIView {
```
bstien: Does this class need to be public?

fespinoza: I guess it doesn't really have to be. I can fix that

PR #454, opened 2019-07-12
Feature/hud

3lvis:
# Why?

Because we are lacking a simple HUD-like component for use cases such as displaying a confirmation that a URL was copied to the clipboard.

# What?

Adds a different display type to LoadingView that shows a HUD-like component.

# Show me

loading4


ag12:
Thanks for your contributions.
I was wondering do you have a link to Sketch files for this one? Trello? Couldn't find it my self. Need to verify the design.

3lvis:
@ag12 nope, just discussed with Thuy, design is verified.

ag12:
Ok. Where is this going to be used? The "Why" on your PR is not sufficient enough.

3lvis:
I can expand on the Why :)

We're redesigning the way we see the map when you tap in an address inside the object page. Here there's a copy address button, after you press this button we want to notify the user that the URL was copied, hence the added functionality.

IMG_7017

ag12:
Thanks for clarifying this.

3lvis:
@finn-no/ios Any input on this?

bstien:
Does this mean we have to set the `displayType` before and after we show the loading view to reset it?

```swift
LoadingView.shared.displayType = .boxed
LoadingView.showSuccess(withMessage: "You got it")
LoadingView.shared.displayType = .fullScreen
```

3lvis:
Good question, I think the display type should be part of the `showSuccess` method. So we choose what to use in each case.

3lvis:
Improved to provide the HUD.

3lvis:
I've tweaked the methods to provide a way to say which type of display type it should use, it defaults to fullscreen. Also, added a segment control to the demo to switch between types.

Simulator Screen Shot - iPhone Xʀ - 2019-07-16 at 13 45 21

Sources/Fullscreen/LoadingView/LoadingView.swift
```diff
@@ -12,35 +12,43 @@ import UIKit
...
-    private lazy var newLoadingIndicator: LoadingIndicatorView = {
-        let view = LoadingIndicatorView()
-        view.translatesAutoresizingMaskIntoConstraints = false
-        view.transform = loadingIndicatorInitialTransform
-        return view
-    }()
+    private var successImageViewCenterY: NSLayoutConstraint?
+    private var loadingIndicatorCenterY: NSLayoutConstraint?
+
+    private var loadingConstraints = [NSLayoutConstraint]()
```
bstien: This doesn't seem to be used anywhere. Is it needed?

Sources/Fullscreen/LoadingView/LoadingView.swift
```diff
@@ -119,55 +127,93 @@ import UIKit
...
+                messageLabel.textColor = .milk
+                backgroundColor = UIColor.black.withAlphaComponent(0.8)
+                layer.cornerRadius = 16
+
+                NSLayoutConstraint.activate([
+                    heightAnchor.constraint(equalToConstant: boxedSize),
+                    widthAnchor.constraint(equalToConstant: boxedSize),
+                    centerXAnchor.constraint(equalTo: window.centerXAnchor),
+                    centerYAnchor.constraint(equalTo: window.centerYAnchor)
+                    ])            }
```
bstien: A newline instead of spaces would be nice 😄

Sources/Fullscreen/LoadingView/LoadingView.swift
3lvis: I was using it but then I figured I didn't need it. It's removed now :)

Sources/Fullscreen/LoadingView/LoadingView.swift
3lvis: Done!

PR #456, opened 2019-07-16
Update style for the settings view

fespinoza:
# Why?

In order to standardize the settings styles making them more close to the Settings app in iOS.

And also because this view will be used for the non-logged user settings

# What?

Update the styles according what was defined on [abstract for the non-logged user settings](https://share.goabstract.com/a07b5458-f9e1-4fd5-b517-04b2ef38b0c5?collectionLayerId=fdb63fa6-aaa6-4a35-a496-97af0a0a697f&mode=build)

# Show me

| Before | After |
| --- | :---: |
| img1 | img2 |


Sources/Recycling/ListViews/Settings/Cell/SettingsViewCell.swift
```diff
@@ -86,13 +86,16 @@ private extension SettingsViewCell {
...
+            arrowView.widthAnchor.constraint(equalToConstant: 8),
 
             hairline.leadingAnchor.constraint(equalTo: titleLabel.leadingAnchor),
             hairline.trailingAnchor.constraint(equalTo: contentView.trailingAnchor),
             hairline.bottomAnchor.constraint(equalTo: contentView.bottomAnchor),
             hairline.heightAnchor.constraint(equalToConstant: 1 / UIScreen.main.scale),
 
-            contentView.bottomAnchor.constraint(equalTo: titleLabel.bottomAnchor, constant: labelInset)
+            contentView.bottomAnchor.constraint(equalTo: titleLabel.bottomAnchor, constant: labelInset),
+            contentView.heightAnchor.constraint(greaterThanOrEqualToConstant: 52)
```
bstien: Could we pull `52` into a static var instead?

Sources/Recycling/ListViews/Settings/ListView/SettingsView.swift
```diff
@@ -19,12 +19,12 @@ public class SettingsView: UIView {
...
         view.translatesAutoresizingMaskIntoConstraints = false
         view.contentInset = UIEdgeInsets(top: .mediumLargeSpacing, leading: 0, bottom: 0, trailing: 0)
-        view.backgroundColor = .milk
+        view.backgroundColor = .marble
         view.separatorStyle = .none
         if #available(iOS 11, *) {
         } else {
-            view.estimatedRowHeight = 44
-            view.estimatedSectionHeaderHeight = 44
+            view.estimatedRowHeight = 52
```
bstien: Ref. the comment above 👆 Here we could use the static var. I.e. `SettingsViewCell.estimatedRowHeight`.

PR #464, opened 2019-07-18
Feature/address view

3lvis:
# Why?

In our process of cleaning up old screens in our app we started working on the screen that gets presented when you tap on an address. We not only improved the screen adding auto layout and iPad support but we also added extra features such as copy address, get directions and center map in the focused area.

# What?

This PR includes the relevant UI elements to make this feature work in our app.

# Show me

| iPhone Pin | iPhone Area |
| --- | :---: |
| iPhone Pin | iPhone Area |

| iPad Portrait | iPad Landscape |
| --- | :---: |
| iPad Portrait | iPad Landscape |


Sources/Fullscreen/AddressView/AddressCardView.swift
```diff
@@ -0,0 +1,103 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import UIKit
+import MapKit
+
+protocol AddressCardViewDelegate: class {
```
bstien: This should be `AnyObject` instead of `class`

Sources/Fullscreen/AddressView/AddressCardView.swift
```diff
@@ -0,0 +1,103 @@
...
+    func addressCardViewDidSelectCopyButton(_ addressCardView: AddressCardView)
+    func addressCardViewDidSelectGetDirectionsButton(_ addressCardView: AddressCardView)
+}
+
+class AddressCardView: UIView {
+    weak var delegate: AddressCardViewDelegate?
```
bstien: Should probably be made `public`

Sources/Fullscreen/AddressView/AddressCardView.swift
```diff
@@ -0,0 +1,103 @@
...
+    }()
+
+    lazy var copyButton: Button = {
+        let button = Button(style: .default, size: .small)
+        button.translatesAutoresizingMaskIntoConstraints = false
+        button.addTarget(self, action: #selector(copyAction), for: .touchUpInside)
+        return button
+    }()
+
+    lazy var getDirectionsButton: Button = {
```
bstien: Could this property, and the ones above, be made `private`?

Sources/Fullscreen/AddressView/AddressCardView.swift
```diff
@@ -0,0 +1,103 @@
...
+            getDirectionsButton.leadingAnchor.constraint(equalTo: leadingAnchor, constant: .mediumLargeSpacing),
+            getDirectionsButton.bottomAnchor.constraint(equalTo: bottomAnchor, constant: -.mediumLargeSpacing)
+            ])
+    }
+
+    @objc func getDirectionsAction() {
+        delegate?.addressCardViewDidSelectGetDirectionsButton(self)
+    }
+
+    @objc func copyAction() {
```
bstien: Guess this method and the one above could also be `private`

Sources/Fullscreen/AddressView/AddressView.swift
```diff
@@ -0,0 +1,222 @@
...
+import MapKit
+
+public protocol AddressViewDelegate: class {
+    func addressViewDidSelectCopyButton(_ addressView: AddressView)
+    func addressViewDidSelectGetDirectionsButton(_ addressView: AddressView)
+    func addressViewDidSelectCenterMapButton(_ addressView: AddressView)
+    func addressView(_ addressView: AddressView, didSelectMapTypeAtIndex index: Int)
+}
+
+@objc public class AddressView: UIView {
```
bstien: Do we need to expose this class to objc?

Sources/Fullscreen/AddressView/AddressView.swift
```diff
@@ -0,0 +1,222 @@
...
+        let control = UISegmentedControl(items: [])
+        control.translatesAutoresizingMaskIntoConstraints = false
+        control.addTarget(self, action: #selector(mapTypeChanged), for: .valueChanged)
+        control.tintColor = .primaryBlue
+        return control
+    }()
+
+    private lazy var segmentContainer: UIView = {
+        let segmentContainer = UIView()
+        segmentContainer.translatesAutoresizingMaskIntoConstraints = false
```
bstien: You could rather use our FK init, `let segmentContainer = UIView(withAutoLayout: true)` to get rid of this line 🙌

Sources/Fullscreen/AddressView/AddressCardView.swift
3lvis: All the AddressCardView is internal use since it's not meant to be used from outside of the framework. It's this little card

Screenshot 2019-07-18 at 14 17 10

Sources/Fullscreen/AddressView/AddressCardView.swift
bstien: Yes, but not for new stuff 😄

Sources/Fullscreen/AddressView/AddressCardView.swift
bstien: Ah, I see 👍

Sources/Fullscreen/AddressView/AddressView.swift
```diff
@@ -0,0 +1,222 @@
...
+        let view = MKMapView()
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.delegate = self
+        return view
+    }()
+
+    private lazy var addressCardView: AddressCardView = {
+        let view = AddressCardView()
+        view.delegate = self
+        view.translatesAutoresizingMaskIntoConstraints = false
```
bstien: Same for this one, and `mapView` above, regarding `.init(withAutoLayout: true)`

Sources/Fullscreen/AddressView/AddressView.swift
```diff
@@ -0,0 +1,222 @@
...
+    public override init(frame: CGRect) {
+        super.init(frame: frame)
+        setup()
+    }
+
+    public required init?(coder aDecoder: NSCoder) {
+        fatalError()
+    }
+
+    public var model: AddressViewModel? {
```
bstien: Could you place this property at the top with the delegate? Just to have them grouped together

Sources/Fullscreen/AddressView/AddressView.swift
3lvis: We don't, fixing :)

Sources/Fullscreen/AddressView/AddressView.swift
```diff
@@ -0,0 +1,222 @@
...
+
+    public func changeMapType(mapType: MKMapType) {
+        mapView.mapType = mapType
+    }
+}
+
+// MARK: - Private methods
+
+private extension AddressView {
+    private func setup() {
```
bstien: Think you can remove the private modifier, since it's already in a private extension

Sources/Fullscreen/AddressView/AddressView.swift
3lvis: Updated all over the place. It seems like our components by default don't have this, like buttons and labels, I think it should be default since we do everything with AutoLayout. What do you think?

Sources/Fullscreen/AddressView/AddressView.swift
3lvis: Moved the delegate bellow so all the publics are together

Sources/Fullscreen/AddressView/AddressView.swift
3lvis: Fixed!

Sources/Fullscreen/AddressView/AddressView.swift
bstien: Yup, agree. I don't think we have a single scenario where we use our `Button` or `Label` classes without setting `translatesAutoresizingMaskIntoConstraints` to `false`.

3lvis:
PR updated following comments 💯

3lvis:
Thanks, Bastian 👑

PR #467, opened 2019-07-19
Campaign: Klimabrølet

fespinoza:
# Why?

There will be a campaign like Earth Hour by the 30th of August called [Klimabrølet](https://klimabrolet.no) and FINN is participating.

# What?

Add the view to show a pop-up given a date range and user's distance to the events.

# Show me

img


bstien:
Not sure if I'm correct, but it seems the `g` in `August` gets clipped just a tiny bit.

Screenshot 2019-07-19 at 14 43 26

Demo/Components/Klimabrolet/KlimabroletEventsDemoViewController.swift
```diff
@@ -0,0 +1,96 @@
...
+        Event(title: "Klimabrølet Tønsberg"),
+        Event(title: "Klimabrølet Stavanger"),
+        Event(title: "Klimabrølet Bergen"),
+        Event(title: "Klimabrølet Røros"),
+        Event(title: "Klimabrølet Trondheim"),
+        Event(title: "Klimabrølet Henningsvær"),
+        Event(title: "Klimabrølet Svalbard"),
+    ]
+
+    var delegate: KlimabroletEventsDemoViewControllerDelegate?
```
bstien: Should be `weak`, just to be safe

Sources/Components/Klimabrolet/KlimabroletView.swift
```diff
@@ -0,0 +1,177 @@
...
+            return
+        }
+
+        if scrollView.contentOffset.y + scrollView.frame.height < contentSize.height {
+            actionsView.layer.shadowOpacity = 0.2
+        }
+    }
+}
+
+// MARK: - KlimabroletView UIScrollViewDelegate
```
bstien: Guess `KlimabroletView` isn't necessary here, since it's the protocol we're interested in. Same for the other ones below.

Sources/Components/Klimabrolet/Views/KlimabroletActionsView.swift
```diff
@@ -0,0 +1,76 @@
...
+
+import Foundation
+
+protocol KlimabroletActionsViewDelegate: AnyObject {
+    func klimabroletViewDidSelectPrimaryButton(_ view: KlimabroletActionsView)
+    func klimabroletViewDidSelectSecondaryButton(_ view: KlimabroletActionsView)
+}
+
+class KlimabroletActionsView: UIView {
+    public var delegate: KlimabroletActionsViewDelegate?
```
bstien: Should also be `weak`

Sources/Components/Klimabrolet/Views/KlimabroletActionsView.swift
```diff
@@ -0,0 +1,76 @@
...
+
+    private(set) lazy var primaryButton: UIButton = {
+        let button = Button(style: .callToAction)
+        button.translatesAutoresizingMaskIntoConstraints = false
+        button.titleLabel?.numberOfLines = 0
+        button.addTarget(self, action: #selector(handleTapOnPrimaryButton), for: .touchUpInside)
+        return button
+    }()
+
+    private(set) lazy var secondaryButton: UIButton = {
```
bstien: Do we need to expose these two buttons, or could the be `private`?

Sources/Components/Klimabrolet/Views/KlimabroletActionsView.swift
```diff
@@ -0,0 +1,76 @@
...
+            secondaryButton.trailingAnchor.constraint(equalTo: layoutMarginsGuide.trailingAnchor),
+            secondaryButton.bottomAnchor.constraint(equalTo: layoutMarginsGuide.bottomAnchor),
+        ])
+    }
+
+    @objc func handleTapOnPrimaryButton() {
+        delegate?.klimabroletViewDidSelectPrimaryButton(self)
+    }
+
+    @objc func handleTapOnSecondaryButton() {
```
bstien: These two `@objc` methods can be `private`

Sources/Components/Klimabrolet/Views/KlimabroletActionsView.swift
bstien: Ah, I see now why they're `private(set)`.

Could you instead make a method called `configure(primaryButtonTitle: String, secondaryButtonTitle: String)` to use from the `KlimabroletView.configure(…)` method?

Sources/Components/Klimabrolet/Views/KlimabroletContentView.swift
```diff
@@ -0,0 +1,97 @@
...
+
+    private(set) lazy var bodyTextLabel: UILabel = {
+        let label = Label(style: .body)
+        label.translatesAutoresizingMaskIntoConstraints = false
+        label.numberOfLines = 0
+        label.textAlignment = .center
+        return label
+    }()
+
+    private(set) lazy var accessoryButton: UIButton = {
```
bstien: Same for these `private(set)` properties regarding a `configure(…)` method

Sources/Resources/FinniversImageAsset.swift
```diff
@@ -90,6 +91,7 @@ public enum FinniversImageAsset: String {
     case more
     case moreImg
     case moteplassen
+    case newClose
```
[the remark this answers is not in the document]
fespinoza: sure.. do you have any suggestions? this will be part of the 24x24 icons moving forward, which right now are 28x28

fespinoza:
@bstien this is ready to ship (or review at least)

I fixed all your comments + checked some iOS 10 potential issues.

regarding the comment about the label being cut, i guess it's just the font as i checked in the view debugger

Screenshot 2019-07-29 at 10 33 00

pimms:
It seems like it's actually being clipped, I tried to change the `bound`'s height in Reveal, and it revealed (eheheh) the lower part of the "g".

Screen Shot 2019-07-30 at 09 43 15

fespinoza:
thanks @pimms, i checked in more detail and yeah, it was being clipped

Sources/Components/Klimabrolet/KlimabroletView.swift
```diff
@@ -0,0 +1,177 @@
...
+        backgroundColor = .milk
+        layer.cornerRadius = 20
+        clipsToBounds = true
+
+        scrollView.addSubview(contentView)
+
+        NSLayoutConstraint.activate([
+            contentView.leadingAnchor.constraint(equalTo: scrollView.leadingAnchor),
+            contentView.trailingAnchor.constraint(equalTo: scrollView.trailingAnchor),
+            contentView.widthAnchor.constraint(equalTo: scrollView.widthAnchor),
```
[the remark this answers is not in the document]
bstien: Is it possible to use `contentView.fillInSuperview()` or do you need to set the width constraint as well?

Sources/Components/Klimabrolet/Views/KlimabroletActionsView.swift
```diff
@@ -0,0 +1,77 @@
...
+
+    private(set) lazy var primaryButton: UIButton = {
+        let button = Button(style: .callToAction)
+        button.translatesAutoresizingMaskIntoConstraints = false
+        button.titleLabel?.numberOfLines = 0
+        button.addTarget(self, action: #selector(handleTapOnPrimaryButton), for: .touchUpInside)
+        return button
+    }()
+
+    private(set) lazy var secondaryButton: UIButton = {
```
bstien: I guess both this and `primaryButton` can be `private` now

fespinoza: 👍

Sources/Components/Klimabrolet/KlimabroletView.swift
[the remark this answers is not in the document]
fespinoza: sorry i didn't address your comment before.

I do need the width anchor constraint, otherwise the content within the scroll view would look like:

Screenshot 2019-07-31 at 10 09 39

I guess it's because a scrollview and given the `intrinsicContentSize` of the content view, then it can still be wider than the scrollview even though is constrained to leading and trailing

But the rest of the constraints can be defined with `content.fillInSuperView`.

PR #468, opened 2019-07-19
Feature: Tweakable

3lvis:
# Why?

When working with components (more specific: fullscreen components), often we find ourselves in need of providing different data to show different variations of our UI, or we might want to mutate the state of our component. The way we have solved this so far is by adding several components to our demos, each with a different setup, unfortunately, this doesn't translate well to fullscreen components where you don't have this freedom since each component covers the whole screen.

# What?

This PR adds the Tweakable functionality, the way it works is by adding the Tweakable protocol to your DemoView, it will ask you to provide an array of TweakableOptions which are plain objects with a title and an action. When you navigate to your demo now you'll find a floating button that after being pressed it will show all your options without interfering with your UI.

## Implementation example

```swift
public class TweakableDemoView: UIView, Tweakable {
    lazy var tweakingOptions: [TweakingOption] = {
        var options = [TweakingOption]()

        options.append(TweakingOption(title: "Option 1", description: nil, action: { parentViewController in
            self.titleLabel.text = "Choosen Option 1!\n\nYou can drag the button too :D"
            parentViewController?.dismiss(animated: true)
        }))

        options.append(TweakingOption(title: "Option 2", description: nil, action: { parentViewController in
            self.titleLabel.text = "Choosen Option 2!\n\nYou can drag the button too :D"
            parentViewController?.dismiss(animated: true)
        }))

        return options
    }()

    ///...
    ///...
    ///...
}
```

# Show me

tweakable


Demo/Tweakable/CornerAnchoringViewController.swift
```diff
@@ -0,0 +1,172 @@
...
+protocol CornerAnchoringViewDelegate: AnyObject {
+    func cornerAnchoringViewDidSelectTweakButton(_ cornerAnchoringView: CornerAnchoringView)
+}
+
+class CornerAnchoringView: UIView {
+    weak var delegate: CornerAnchoringViewDelegate?
+
+    private lazy var anchoredView: EasterEggButton = {
+        let button = EasterEggButton()
+        button.translatesAutoresizingMaskIntoConstraints = false
```
bstien: We could use `let button = EasterEggButton(withAutoLayout: true)` instead

Demo/Tweakable/CornerAnchoringViewController.swift
```diff
@@ -0,0 +1,172 @@
...
+        button.tintColor = .primaryBlue
+        button.setImage(UIImage(named: .wrench).withRenderingMode(.alwaysTemplate), for: .normal)
+        button.isUserInteractionEnabled = true
```
bstien: This property is `true` by default

Demo/Tweakable/CornerAnchoringViewController.swift
```diff
@@ -0,0 +1,172 @@
...
+        super.layoutSubviews()
+
+        anchoredView.center = anchorPositions.last ?? .zero
+    }
+
+    private func addAnchorAreaView() -> UIView {
+        let view = UIView()
+        addSubview(view)
+        anchorAreaViews.append(view)
+        view.translatesAutoresizingMaskIntoConstraints = false
```
bstien: Same here regarding `UIView(withAutoLayout: true)`

Demo/Tweakable/TweakableDemoView.swift
```diff
@@ -0,0 +1,44 @@
...
+
+import FinniversKit
+
+public class TweakableDemoView: UIView, Tweakable {
+    lazy var tweakingOptions: [TweakingOption] = {
+        var options = [TweakingOption]()
+
+        options.append(TweakingOption(title: "Option 1", description: nil, action: { parentViewController in
+            self.titleLabel.text = "Choosen Option 1!\n\nYou can drag the button too :D"
+            parentViewController?.dismiss(animated: true)
```
bstien: Shouldn't it be the responsibility of `TweakingOptionsTableViewController` to dismiss the presented viewController/bottomSheet?

Demo/Tweakable/TweakableDemoView.swift
```diff
@@ -0,0 +1,44 @@
...
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import FinniversKit
+
+public class TweakableDemoView: UIView, Tweakable {
+    lazy var tweakingOptions: [TweakingOption] = {
+        var options = [TweakingOption]()
+
+        options.append(TweakingOption(title: "Option 1", description: nil, action: { parentViewController in
```
bstien: Should probably add `[weak self]`, just to be on the safe side

Demo/Tweakable/TweakingOptionCell.swift
```diff
@@ -0,0 +1,57 @@
+import FinniversKit
+
+class TweakingOptionCell: UITableViewCell {
+    var option: TweakingOption? {
+        didSet {
+            titleLabel.text = option?.title
```
bstien: We tend to create/use a method named something along the lines of `configure(withOption option: TweakingOption)` for configuring cells. Then this property can stay private

Demo/Tweakable/TweakingOptionCell.swift
```diff
@@ -0,0 +1,57 @@
...
+        super.init(style: style, reuseIdentifier: reuseIdentifier)
+        setup()
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+        setup()
+    }
+
+    var descriptionLabelBottomAnchor: NSLayoutConstraint?
```
bstien: This can be `private`. Should also probably be placed together with the rest of the private properties.

Demo/Tweakable/TweakingOptionsTableViewController.swift
```diff
@@ -0,0 +1,55 @@
+import FinniversKit
+
+class TweakingOptionsTableViewController: UIViewController {
+    let options: [TweakingOption]
```
bstien: Can this be private, since the `options` are passed through the constructor?

3lvis:
Fixed the dismissing animation as well as added some state tracking to remember the last position of the wrench 🔧.

PR #471, opened 2019-07-23
Clean up folder structure

3lvis:
# Why?

The TableViewCells section was added to FinniversKit some time ago, when this was done it was done inside the Recycling section which now is more focused to UITableViews and UICollectionViews. So I thought it made sense to separate this folder that represents two sections in the demo into separate folders.

Finally, in the list of components, we go from smaller to biggest so it makes sense that the cells come before the recycling and before fullscreen, leaving the sorting of folders/items to:

- Protocols, Extensions, Others
- DNA
- Components
- Cells
- Recycling
- Fullscreen

# What?

This PR separates cells into its own folder outside of Recycling and changes the order of the sections so Cells is placed before Recycling instead of at the end of everything (after fullscreen).

# Show me

| Before | After |
| --- | --- |
|  |  |


3lvis:
😊 😊 😊

PR #475, opened 2019-07-24
Use categories for Demo app

3lvis:
# Why?

Our Demo app has been growing more and more and we’ve come to the point where it’s hard to navigate between components. This is the first PR that aims to improve the structure of our Demo app.

# What?

Changes the normal sectioned table view for a single section per screen, you can select the top of the screen to jump into a different section. Also, it adds some ideas of what a backward-compatible dark mode could be for us.

### Before

screens-before

### After

screens


Demo/Demo/DemoViewsTableViewController.swift
```diff
@@ -7,16 +7,26 @@ import FinniversKit
...
-    }
 
-    required init?(coder aDecoder: NSCoder) { fatalError("") }
+        NotificationCenter.default.addObserver(self, selector: #selector(userInterfaceStyleDidChange(_:)), name: .DidChangeUserInterfaceStyle, object: nil)
+    }
 
-    override var prefersStatusBarHidden: Bool {
-        return true
+    deinit {
+        NotificationCenter.default.removeObserver(self, name: .DidChangeUserInterfaceStyle, object: nil)
```
pimms: I don't think this is needed since iOS 9

Sources/DNA/Color/Color.swift
```diff
@@ -65,10 +65,42 @@ import UIKit
...
+
     // swiftlint:disable:next identifier_name
     convenience init?(r: CGFloat, g: CGFloat, b: CGFloat, a: CGFloat = 1.0) {
         self.init(red: r / 255.0, green: g / 255.0, blue: b / 255.0, alpha: a)
     }
+
+    /// Base initializer, it creates an instance of `UIColor` using an HEX string.
+    ///
+    /// - Parameter hex: The base HEX string to create the color.
+    convenience init(hex: String) {
```
pimms: Why use `String` when we can use `Int`?

Demo/Demo/DemoViewsTableViewController.swift
3lvis: You are right! For anyone feeling curious check "Automatic Removal of Finalized Key-Value Observers When Running Garbage-Collected (New since November seed)".

3lvis: https://developer.apple.com/library/archive/releasenotes/Foundation/RN-FoundationOlderNotes/index.html#X10_11Notes

Sources/DNA/Color/Color.swift
3lvis: Hex strings are in the form of "1D1D26" or "#1D1D26" so Int wouldn't be a good fit.

Demo/Demo/DemoHelpers.swift
```diff
@@ -218,3 +218,21 @@ extension Array {
         return indices.contains(index) ? self[index] : .none
     }
 }
+
+extension Foundation.Notification.Name {
+    static let DidChangeUserInterfaceStyle = Foundation.Notification.Name("DidChangeUserInterfaceStyle")
```
vadymmarkov: I think we usually don't use capitalized names for constants. Also, most of the notification names from Apple are not capitalized, e.g. https://developer.apple.com/documentation/appkit/nsaccessibility/notification/1527225-applicationactivated

3lvis: I was thinking about that, fixed on upcoming PR

vadymmarkov:
Btw, is there any value in backward-compatible dark mode? I think having dark mode as iOS13 only feature will encourage users to upgrade.

3lvis:
@vadymmarkov One of the benefits is that it simplifies the development since we don't have to have a separate branch for Xcode 11 / iOS 13. In that sense, we can probably drop the compatibility mode once we all start using Xcode 11.

I can imagine that for the most part of 2019 enabling dark mode is gonna be quite bad for users in general since many apps won't be optimized for it. So having the choice of enabling dark mode only for the FINN app seems like a nice thing to do for our users.

PR #480, opened 2019-07-25
RemoteImageView and selectable cell

bstien:
# Why?
We want to have a reusable imageView that is capable of loading remote images through a datasource. Currently we have this functionality in `RemoteImageTableViewCell`, but the functionality should be pulled out into a separate class/view.

The functionality for setting the default background color on our cells should be pulled out into an extension on `UITableViewCell` so we may set it for cells that need it, instead of duplicating the logic.

# What?
- Added view `RemoteImageView`, a subclass of `UIImageView`.
- Added `UITableViewCell` extension with method `setDefaultSelectedBackgound`.
- Make `BasicTableViewCell` call `setDefaultSelectedBackgound` in `setup()`.


vadymmarkov:
As for the naming part, how about `SelectedBackgroundTableViewCell` or just `TableViewCell`?

Sources/Cells/SelectableTableViewCell.swift
```diff
@@ -0,0 +1,25 @@
...
+        super.init(style: style, reuseIdentifier: reuseIdentifier)
+        setup()
+    }
+
+    required public init?(coder aDecoder: NSCoder) { fatalError() }
+
+    // MARK: - Setup
+
+    private func setup() {
+        let selectedBackgroundView = UIView()
```
vadymmarkov: How about moving this function to extension:

```swift
extension UITableViewCell {
    func setDefaultSelectedBackgound() {
        let selectedBackgroundView = UIView()
        selectedBackgroundView.backgroundColor = .defaultCellSelectedBackgroundColor
	self.selectedBackgroundView = selectedBackgroundView
    }
}
```

`BasicTableViewCell` would call this function during its setup, and we can also call it from any other table view cell subclass.

PR #481, opened 2019-07-26
AddressView support for custom tiles

3lvis:
# Why?

Helps to implement support for injecting FINN tiles to our AddressView. Also includes other minor UI tweaks.

# What?

Adds a method to inject custom tiles.

# Show me

Example from the FINN app where this is implemented.

| Before | After |
| --- | --- |
| Old | New |


Sources/Fullscreen/AddressView/AddressView.swift
```diff
@@ -110,16 +110,36 @@ public class AddressView: UIView {
         mapView.setRegion(region, animated: animated)
     }
 
+    private var tileOverlay: MKTileOverlay?
+
+    public func addTileOverlay(tileOverlay: MKTileOverlay) {
```
bstien: Could this be named `updateTileOverlay(_ tileOverlay: MKTileOverlay)`? Since we're not adding, but removing and setting a new overlay.

bstien: Or perhaps `configure(tileOverlay: MKTileOverlay)`. Then we could use the same naming conventions on the other `addX`-methods.

3lvis:
Updated with PRs comments

PR #482, opened 2019-07-26
Support edit mode in the list of favorite folders

vadymmarkov:
# Why?

Because it should be possible to select favorite folders for editing.

# What?

- Implement edit mode in the list of favorite folders according to design. Here we extend the default `UITableView` edit mode by using custom checkmark icons and additional transition animations. There are two new methods `setEditing(true)` and `setEditing(false)` on `FavoriteFoldersListView` that should be called to enter/exit the edit mode.

- Make it possible to reload specific rows in the list of favorite folders by calling new `reloadRow(at:)` method. This fixes the issue with "jumping" table view during selection.

- Don't use the bottom sheet in the favorite folders demo. There are multiple ways of presenting this view, therefore the demo should be more generic.

- Use new `Tweakable` protocol is the favorite folders demo for switching between toggle and edit modes.

# Show me

**Toggle mode**

| Before | After |
| --- | --- |
| toggle_before | toggle_after |

**Edit mode**

edit_mode


vadymmarkov:
It seems like `testAddressView` randomly fails. Any idea why this is happening?

bstien:
Could it be because we potentially load the map from the network? If we can get the failed diff image from CircleCI we'd know, but that's my guess.

vadymmarkov:
Yeah, if there is a map in the address view that could be it.

Sources/Recycling/ListViews/FavoriteFolders/Subviews/FavoriteFolderSelectableViewCell.swift
```diff
@@ -5,21 +5,24 @@
...
+
+    private lazy var leftCheckmarkView: CheckmarkView = {
+        let view = CheckmarkView(withAutoLayout: true)
+        view.isBordered = true
+        return view
+    }()
+
+    private lazy var editModeView: UIView = {
+        let view = UIView(withAutoLayout: true)
+        view.backgroundColor = .white
```
bstien: This should probably be `.milk` so it's consistent with what we set in `setSelected(_:)` and `setHighlighted(_:)`

PR #485, opened 2019-07-29
Fix - RemoteImageView image width

bstien:
# Why?
We need `RemoteImageView` to pass the wanted image width to its data source.

# What?
- Refactored `RemoteImageView`, `FavoriteFoldersListView` and demo.

# Show me
_No UI._


3lvis:
Why not passing the model instead?

bstien:
Because `RemoteImageView` is now a reusable component and not directly tied to `RemoteImageTableViewCell` like it was

PR #486, opened 2019-07-30
Update error and success icons

3lvis:
# Why?

Designer request, previous icons were 18x18, new ones are 16x16.

# What?

Updates the error and success icons.

# Show me

Screenshot 2019-07-30 at 10 17 17


pimms:
They seem to be rendered at the same size, is that intentional? Also, aren't the PDFs embedding SVGs that are arbitrarily scalable?

bstien:
We could probably wait with this one. The color on the checkmark icon is using the old `.pea` color, not the new one, so Thuy is supposed to make new ones.

The new `.pea` color is used on the button on the right

osanoj:
@pimms The PDF is used to generate PNGs, unless you explicitly say you want it to remain resizable (which gives worse performance).

PR #487, opened 2019-07-30
Feature: FavoriteAdCell

bstien:
# Why?
We need a new cell type for our redesign. This will be used when listing favorited ads within a folder.

The icon for the "more" button is probably not the correct one, so I'll keep that in mind until Thuy is back from 🏖🍹

I'm not sure of the access level set on the design spec and won't post it here, so just ask if you want it🤓

# What?
- Added cell `FavoriteAdTableViewCell`.
- Make it possible to set a `RibbonView`'s style after creating it.
- Added demo + snapshot reference images.

# Show me
Screenshot 2019-07-30 at 14 04 38


Demo/Cells/FavoriteAdTableViewCell/FavoriteAdCellDemoView.swift
```diff
@@ -0,0 +1,185 @@
...
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+// MARK: - UITableViewDelegate
+
+extension FavoriteAdCellDemoView: UITableViewDelegate {
+    func tableView(_ tableView: UITableView, willDisplay cell: UITableViewCell, forRowAt indexPath: IndexPath) {
+        let isLastCell = indexPath.row == (viewModels.count - 1)
```
robinsalehjan: Nice one 👍

Sources/Cells/FavoriteAdTableViewCell/FavoriteAdTableViewCell.swift
```diff
@@ -0,0 +1,197 @@
...
+        let button = UIButton(withAutoLayout: true)
+        let image = UIImage(named: .more).withRenderingMode(.alwaysTemplate)
+        button.setImage(image, for: .normal)
+        button.imageView?.contentMode = .scaleAspectFit
+        button.tintColor = .stone
+        button.addTarget(self, action: #selector(moreButtonTapped), for: .touchUpInside)
+        return button
+    }()
+
+    private lazy var statusRibbon: RibbonView = {
```
vadymmarkov: How about `statusRibbon = RibbonView(withAutoLayout: true)`?

bstien: I'll see if I can do that easily for `RibbonView`👍  I have to make some changes there for it to work

bstien:
We're also missing labels to display distance or last updated when sorting the list. These were not in the design, so I'll have to add them later.

bstien:
And I see now that we should probably expand the touch area for the more button. It's only 24x24pts

PR #488, opened 2019-07-30
Fix: FavoriteAdCell touch area for more button

bstien:
# Why?
As I mentioned in the last comment in #487 I noticed the touchable area for the more button was only 24x24pts.

I've set the content inset for the button and removed some constraint constants so we keep the spacing between elements as in the design spec. The touchable area is now 40x44pts.

# What?
- Change touchable area for button from 24x24pts to 40x44pts.

# Show me

| Before | After |
| --- | --- |
| Screenshot 2019-07-30 at 15 00 33 | Screenshot 2019-07-30 at 14 59 15 |


fespinoza:
the trial ends in 8 days! 😨

PR #493, opened 2019-08-02
Favorite folders: add refresh control

vadymmarkov:
# Why?

Because it should be possible to refresh data by user request.

# What?

- Add custom refresh control to the list of favorite folders
- Hide refresh control in the edit mode and during search
- Add new delegate method to call when refresh control begins refreshing

# Show me

refresh_control


bstien:
This wouldn't block any functionality when we present it within a `BottomSheet`, right?

vadymmarkov:
Refresh control starts refreshing on scroll within the table view, so I guess it shouldn't conflict with the bottom sheet.

PR #494, opened 2019-08-02
Feature: SearchResultMapView and MapSettingsButton

hakloev:
# Why?

We're refactoring the view controller in the FINN-app for showing search results on a map. Thus, the map view and buttons for controlling the map are added to FinniversKit

# What?

* Added `SearchResultMapView`
* Added `MapSettingsButton`
* Added demo view
* Added `.distanceOutlined` as an asset 
* Replaced the `.info` asset with a vector based asset to allow setting the tint color

# TODO

- [x] Add snapshot reference image 
- [x] Rebase with `master`

# Show me

### FinniversKit Demo

Simulator Screen Shot - iPhone Xs - 2019-08-02 at 14 44 10

### FINN-app

Simulator Screen Shot - iPhone Xs - 2019-08-02 at 14 43 11
Simulator Screen Shot - iPad Air - 2019-08-02 at 14 42 20


Sources/Fullscreen/SearchResultMapView/SearchResultMapView.swift
```diff
@@ -0,0 +1,186 @@
...
+        mapSettingsButton.delegate = self
+        return mapSettingsButton
+    }()
+
+    // MARK: - Public
+
+    public override func layoutSubviews() {
+        super.layoutSubviews()
+
+        if !didSetupView {
```
3lvis: Why doing the setup in a custom layoutSubviews instead of in an init? Doing it in the init will remove the need to have `didSetupView` or am I missing something? 🤔 :)

Sources/Fullscreen/SearchResultMapView/SearchResultMapView.swift
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        return MKOverlayRenderer(overlay: overlay)
+    }
+
+    public func mapView(_ mapView: MKMapView, viewFor annotation: MKAnnotation) -> MKAnnotationView? {
+        guard !(annotation is MKUserLocation) else { return nil }
+
+        if let annotation = annotation as? SearchResultMapViewAnnotation {
+            let marker = MKAnnotationView(annotation: annotation, reuseIdentifier: annotation.isCluster ? "clusterPOI" : "POI")
```
3lvis: There's something strange about having the annotations reuse identifier as hardcoded strings here, maybe we can make a String enum or something safely typed?

Sources/Fullscreen/SearchResultMapView/SearchResultMapView.swift
```diff
@@ -0,0 +1,186 @@
...
+            marker.image = annotation.image
+            marker.accessibilityIdentifier = "annotationPOI"
```
3lvis: What's this needed for? Is this what Siri will read in VoiceOver?

Demo/Components/Button/MapSettings/MapSettingsButtonDemoView.swift
```diff
@@ -0,0 +1,34 @@
+//
+//  Copyright © 2019 FINN AS. All rights reserved.
+//
+
+import FinniversKit
+
+public class MapSettingsButtonDemoView: UIView {
+
+    private lazy var mapSettingsButton: MapSettingsButton = {
```
vadymmarkov: Could be just `private lazy var mapSettingsButton = MapSettingsButton(withAutoLayout: true)`

Sources/Fullscreen/SearchResultMapView/SearchResultMapView.swift
hakloev: You're right! This is probably the consequence of me looking at a similar view and trying to adapt the same patterns. Will change it to an `init` ☺️

Demo/Fullscreen/SearchResultMapView/SearchResultMapViewDemoViewController.swift
```diff
@@ -0,0 +1,114 @@
...
+        searchResultMapView.clearMapOverlay(SearchResultMapViewAnnotationFactory.tileOverlay)
+    }
+
+}
+
+// MARK: - SearchResultMapViewDelegate
+
+extension SearchResultMapViewDemoViewController: SearchResultMapViewDelegate {
+
+    public func searchResultMapViewDidSelectChangeMapTypeButton(_ view: SearchResultMapView, mapSettingsButtonView: MapSettingsButton) {
```
shredlocker: We usually write these delegate methods in a neat and swifty way like
`searchResultMapView(_ mapView: SearchResultMapView, didSelect mapSettingsButton: MapSettingsButton)`.

Sources/Fullscreen/SearchResultMapView/SearchResultMapView.swift
hakloev: Absolutely agree! I'll extract them.

Demo/Fullscreen/SearchResultMapView/SearchResultMapViewDemoViewControllerHelpers.swift
```diff
@@ -0,0 +1,47 @@
+//
+//  Copyright © 2019 FINN AS. All rights reserved.
+//
+
+import MapKit
+import FinniversKit
+
+public class SearchResultMapViewDemoAnnotation: NSObject, SearchResultMapViewAnnotation {
+
+    public var isCluster: Bool
```
vadymmarkov: Do these properties need to be mutable?

Demo/Fullscreen/SearchResultMapView/SearchResultMapViewDemoViewControllerHelpers.swift
```diff
@@ -0,0 +1,47 @@
...
+
+    public static var tileOverlay: MKTileOverlay = {
+        let tile = MKTileOverlay(urlTemplate: "https://tile.openstreetmap.org/{z}/{x}/{y}.png")
+        tile.canReplaceMapContent = true
+        return tile
+    }()
+
+    public static func create(numberOfAnnotations: Int) -> [SearchResultMapViewDemoAnnotation] {
+        return (0 ..< numberOfAnnotations).map { _ in
+            let isCluster = Double.random(in: 0...1) >= 0.8
```
shredlocker: To me it makes a bit more sense to say `< 0.2`. I think that makes it more clear that it is 20 %.

Sources/Components/Button/MapSettings/MapSettingsButton.swift
```diff
@@ -0,0 +1,100 @@
...
+        layer.backgroundColor = UIColor.milk.withAlphaComponent(0.8).cgColor
+    }
+
+    private func setup() {
+        addSubview(mapTypeButton)
+        addSubview(divider)
+        addSubview(centerMapButton)
+
+        NSLayoutConstraint.activate([
+            mapTypeButton.topAnchor.constraint(equalTo: topAnchor, constant: 8),
```
vadymmarkov: You can also use `.mediumSpacing` instead of 8 here.

Sources/Components/Button/MapSettings/MapSettingsButton.swift
```diff
@@ -0,0 +1,100 @@
...
+            mapTypeButton.widthAnchor.constraint(equalToConstant: 42),
+            mapTypeButton.heightAnchor.constraint(equalToConstant: 32),
```
vadymmarkov: 32 can be replaced with `.largeSpacing`. Check `Spacing.swift` file, there are more constants we usually use for defining constraints.

Sources/Components/Button/MapSettings/MapSettingsButton.swift
```diff
@@ -0,0 +1,100 @@
+//
+//  Copyright © 2019 FINN.no. All rights reserved.
+//
+
+import UIKit
+
+public protocol MapSettingsButtonDelegate: AnyObject {
```
shredlocker: You could maybe have one delegate method here and define action enums for each button? Or maybe a better word than action..

`mapSettingsButton(_ mapButton: MapSettingsButton, didSelect action: MapSettingsAction)`

Sources/Fullscreen/SearchResultMapView/SearchResultMapView.swift
hakloev: It can probably be removed, since I most likely just added this line since it was present in the old code. As far as I can understand, this is used to identify an annotation element in the  existing UITests.

Sources/Components/Button/MapSettings/MapSettingsButton.swift
```diff
@@ -0,0 +1,100 @@
...
+            centerMapButton.leadingAnchor.constraint(equalTo: leadingAnchor),
+            centerMapButton.trailingAnchor.constraint(equalTo: trailingAnchor),
+            centerMapButton.widthAnchor.constraint(equalToConstant: 42),
+            centerMapButton.heightAnchor.constraint(equalToConstant: 32),
+            ])
+    }
+
+    // MARK: - Actions
+
+    @objc func mapTypeButtonTapped() {
```
vadymmarkov: These two methods can be `private`.

Sources/Components/Button/MapSettings/MapSettingsButton.swift
```diff
@@ -0,0 +1,100 @@
...
+//
+
+import UIKit
+
+public protocol MapSettingsButtonDelegate: AnyObject {
+    func mapSettingsButtonDidSelectChangeMapTypeButton(_ view: MapSettingsButton)
+    func mapSettingsButtonDidSelectCenterMapButton(_ view: MapSettingsButton)
+}
+
+public final class MapSettingsButton: UIView {
```
shredlocker: Could you rename this to MapSettingsView? Button confused me a bit..

Sources/Fullscreen/SearchResultMapView/SearchResultMapView.swift
```diff
@@ -0,0 +1,186 @@
+//
+//  Copyright © 2019 FINN AS. All rights reserved.
+//
+
+import MapKit
+
+public protocol SearchResultMapViewDelegate: AnyObject {
+    func searchResultMapViewDidSelectChangeMapTypeButton(_ view: SearchResultMapView, mapSettingsButtonView: MapSettingsButton)
```
vadymmarkov: How about `searchResultMapView(_ view: SearchResultMapView, didSelectMapSettingsButton button: MapSettingsButton)`

Sources/Fullscreen/SearchResultMapView/SearchResultMapView.swift
```diff
@@ -0,0 +1,186 @@
...
+    func searchResultMapViewDidSelectCenterMapButton(_ view: SearchResultMapView)
+    func searchResultMapViewDidSelectAnnotationView(_ view: SearchResultMapView, annotationView: MKAnnotationView)
```
vadymmarkov: How about `searchResultMapView(_ view: SearchResultMapView, didSelectAnnotationView view: MKAnnotationView)`

Sources/Components/Button/MapSettings/MapSettingsButton.swift
hakloev: Yes, I can do that. I named it `MapSettingsButton` since I thought it was a convention due to all the other buttons within `Components/Button` being named `<name>Button.swift`

Demo/Fullscreen/SearchResultMapView/SearchResultMapViewDemoViewController.swift
shredlocker: This makes more sense if we have those action enums I mention later. So maybe:

`searchResultMapView(_ mapView: SearchResultMapView, didSelect mapSettingsAction: MapSettingsAction, in view: MapSettingsView)`

Or it just makes it more confusing 😆

hakloev:
I've addressed the comments now 👍 Please have another look on the changes.

Demo/Fullscreen/SearchResultMapView/SearchResultMapViewDemoViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+}
+
+// MARK: - SearchResultMapViewDelegate
+
+extension SearchResultMapViewDemoViewController: SearchResultMapViewDelegate {
+
+    public func searchResultMapView(_ view: SearchResultMapView, didSelect mapSettingsAction: MapSettingsButton.Actions, in button: MapSettingsButton) {
+        switch mapSettingsAction {
```
shredlocker: I think it is enough to call this `action`, its type is enough to know what it is in these types of delegate methods.

Sources/Components/Button/MapSettings/MapSettingsButton.swift
```diff
@@ -0,0 +1,104 @@
...
+
+import UIKit
+
+public protocol MapSettingsButtonDelegate: AnyObject {
+    func mapSettingsButton(_ button: MapSettingsButton, didSelect action: MapSettingsButton.Actions)
+}
+
+public final class MapSettingsButton: UIView {
+
+    public enum Actions {
```
shredlocker: Not a big deal, but I would make this `Action` in singular form.

Sources/Components/Button/MapSettings/MapSettingsButton.swift
shredlocker: I didn't realise you put this in the `Button` folder. Since this view is a bit more complicated than a single button I would rename it and put it under `Components` ;)

3lvis:
I wouldn't make MapSettingsButton into a separate component since it's not meant to be used outside of SearchResultMapView. I think a good rule of thumb is to add this as components when they are meant to be used in multiple places or if they can have multiple purposes. At the moment I don't see this button been used anywhere else but I might be wrong.

shredlocker:
Put it in the same folder as `SeachResultMapView`, that way it is clear that it is only used by that view ;)

PR #507, opened 2019-08-08
Neighborhood profile: add paging on iPhone

vadymmarkov:
# Why?

We decided to enable paging for the collection view on iPhone and keep regular scrolling behaviour on iPad.

# What?

- Implement custom collection view flow layout with paging in mind. Each cells except for the first and last ones will be centered

- Add page control

- Update layout and collection view height calculations in `NeighborhoodProfileView`

# Show me

np


vadymmarkov:
Made a couple of changes to keep page control more in sync with scrolling.

np

PR #513, opened 2019-08-12
Neighborhood profile: support dynamic fonts

vadymmarkov:
# Why?

Because the layout was broken with scaled fonts.

# What?

- Support dynamic fonts in `NeighborhoodProfileView`
- Fix calculations in `systemLayoutSizeFitting`
- Set header data in `configure` method

# Show me

IMG_6604


Sources/Recycling/GridViews/NeighborhoodProfile/Cells/NeighborhoodProfileInfoViewCell.swift
```diff
@@ -119,12 +121,12 @@ final class NeighborhoodProfileInfoViewCell: NeighborhoodProfileViewCell {
...
             linkButtonToStackViewConstraint,
             linkButton.leadingAnchor.constraint(equalTo: titleLabel.leadingAnchor),
-            linkButton.trailingAnchor.constraint(equalTo: iconImageView.leadingAnchor),
+            linkButton.trailingAnchor.constraint(equalTo: titleLabel.trailingAnchor),
 
             iconImageView.trailingAnchor.constraint(equalTo: titleLabel.trailingAnchor),
             iconImageView.widthAnchor.constraint(equalToConstant: NeighborhoodProfileInfoViewCell.iconSize),
             iconImageView.heightAnchor.constraint(equalTo: iconImageView.widthAnchor),
-            iconImageView.bottomAnchor.constraint(equalTo: bottomAnchor, constant: -.mediumLargeSpacing)
+            iconImageView.bottomAnchor.constraint(equalTo: contentView.bottomAnchor, constant: -.mediumLargeSpacing)
```
pimms: The `iconImageView` is added to `self`, does this constraint make sense? Alternatively, shouldn't all subs be added to `contentView`?

vadymmarkov: Nice catch, I'll add all subviews to `contentView` 😉

PR #516, opened 2019-08-13
[WIP] Button component and AddressView in dark mode

3lvis:
# Why?

Brainstorming approaches to supporting dark mode into our app.

# What?

AddressView required buttons to support dark mode so I updated the control as well. This is a rough draft and WIP.

# Show me

|  🌝   | 🌚 |
| --- | --- |
|  |  |
|  |  |


3lvis:
`UserInterfaceUpdatable` is required since only the view controller receives the notification on when to switch to dark mode via 

```
override public func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
```

Also, only the UIViewController has the UITraitCollection which contains the UIUserInterfaceStyle that's required to know if the app uses Dark mode or not.

shredlocker:
What about using `UITraitEnvironment`? `UIView` has that property and you can access the `traitCollection` from there. Then you can use the main `UIWindow` as the trait environment.

Sources/Components/Button/Button/Button.swift
```diff
@@ -10,6 +10,7 @@ public class Button: UIButton {
     private let cornerRadius: CGFloat = 8.0
     private var titleHeight: CGFloat?
     private var titleWidth: CGFloat?
+    var userInterfaceStyle: UserInterfaceStyle?
```
shredlocker: I feel like this is not necessary. I was imagining a static theme with all the defined colors. 

Something like:
```
struct Theme {
    static var current: Theme {
        // dark or light
    }

    var background: UIColor
    var foreground: UIColor
    var link: UIColor
    // ... etc
}
```

Then use it like this maybe:

```
self.backgroundColor = Theme.current.background
```

3lvis: How we would check the current user style, tho? Should we keep a global state? Grab it from the UIWindow? 

Haven't come up with a proper solution other than passing it from the view.

3lvis:
> What about using `UITraitEnvironment`? `UIView` has that property and you can access the `traitCollection` from there. Then you can use the main `UIWindow` as the trait environment.

Good point! Updated to use the `traitCollection` from the UIView.

Sources/Components/Button/Button/Button.swift
shredlocker: I don't know how it is going to work, but I would guess we could grab it from UIWindow when we create it in AppDelegate.

shredlocker: Or! We might want to start using the appearance api. Implement this method in all the components: `appearance(for trait: UITraitCollection)`

3lvis: It's tricky since the only one that knows the trait collection is the view and view controllers. The UIWindow has no knowledge of the UIUserInterfaceStyle as far as I know. 

Also, when iOS switches to dark or light mode, only UIViewControllers are informed of this change, the UIWindow or the AppDelegate don't know about it.

osanoj: Isn't UIWindow a UIView? 😄

osanoj: One question is also if we want to support that users can override dark or light in our app or if we should only follow the iOS setting?

shredlocker: UIWindow is a UITraitEnvironment 😉

3lvis: > Isn't UIWindow a UIView? 😄

Haha, true! Doesn't change much since it still doesn't get informed for changes in the UITraitCollection.

> One question is also if we want to support that users can override dark or light in our app or if we should only follow the iOS setting?

I think the proper implementation is to react to changes. Only following the iOS setting could be a viable option as well if we don't mind the lack of responsiveness.

I think we need to decide what we want to do.

3lvis: I'll do an iteration without listening to changes.

3lvis: This is the alternative https://github.com/finn-no/FinniversKit/pull/517

3lvis: Whether we do live updates in the style or if we set the style when the app launches it's something that should be decided by the designer, PM, and TDE with feedback from the developers.

And we should decide soon so we can continue the progress with this task :)

osanoj:
Close this for #590?

ag12:
Seams like this is a redundant PR and should be closed.

PR #517, opened 2019-08-13
Dark mode easy mode

3lvis:
# Why?

A simpler approach to supporting dark mode reduces implementation complexity.

# What?

Stores the current style and reuses that before setting colors. It doesn't have the same dramatic effect but it does the job.

# Show me

Simulator Screen Shot - iPhone X - 2019-08-13 at 15 57 25


osanoj:
Close this for #590?

PR #518, opened 2019-08-13
Quick implementation for using semantic dynamic colors on iOS 13

osanoj:
Note: This is not a PR to master, it is an example of how #517 might be improved.

# Why?
Wanted to try using `init(dynamicProvider: @escaping (UITraitCollection) -> UIColor)` to see if we could avoid listening to `traitCollectionDidChange` and still get our colors to change "live".

# What?
Changed implementation for iOS 13 to have dynamic colors based on dark/light mode. 

Only tested on iOS 13 too, so it might break everything else. And needs cleanup.

# Show me
dark-demo

### DIFF
--- a/Sources/DNA/Color/UserInterfaceStyle.swift
+++ b/Sources/DNA/Color/UserInterfaceStyle.swift
@@ -24,8 +24,16 @@ public class Theme {
         }
     }
 
+    /// Needs to be called from main thread on iOS 13
     public static func setCurrentUserInterfaceStyle(_ userInterfaceStyle: UserInterfaceStyle) {
-        UserDefaults.standard.set(userInterfaceStyle.rawValue, forKey: currentUserInterfaceStyleKey)
-        UserDefaults.standard.synchronize()
+        if #available(iOS 13.0, *) {
+            // Untested, this might change mode live (but not remember the setting)
+            let uiUserInterfaceStyle: UIUserInterfaceStyle = userInterfaceStyle == .dark ? .dark : .light
+            let updatedTraits = UITraitCollection(traitsFrom: [UITraitCollection.current, UITraitCollection(userInterfaceStyle: uiUserInterfaceStyle)])
+            UITraitCollection.current = updatedTraits
+        } else {
+            UserDefaults.standard.set(userInterfaceStyle.rawValue, forKey: currentUserInterfaceStyleKey)
+            UserDefaults.standard.synchronize()
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


osanoj:
This only builds on the Xcode beta I see 😄

3lvis:
👏

osanoj:
Fixed build and added possibility to use moon button on iOS 13 (but that isn't used/loaded when app restarts).

dark-mode-live-switch

Now you can do what you want with this PR @3lvis 😄

3lvis:
🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥

Sources/DNA/Color/UserInterfaceStyle.swift
```diff
@@ -24,7 +24,16 @@ public class Theme {
...
 
-    public static func setCurrentUserInterfaceStyle(_ userInterfaceStyle: UserInterfaceStyle) {
+    /// Needs to be called from main thread on iOS 13
+    public static func setCurrentUserInterfaceStyle(_ userInterfaceStyle: UserInterfaceStyle, in window: UIWindow?) {
+        if #available(iOS 13.0, *) {
+            #if swift(>=5.1)
+            let uiUserInterfaceStyle: UIUserInterfaceStyle = userInterfaceStyle == .dark ? .dark : .light
+            let updatedTraits = UITraitCollection(traitsFrom: [UITraitCollection.current, UITraitCollection(userInterfaceStyle: uiUserInterfaceStyle)])
+            UITraitCollection.current = updatedTraits
+            window?.overrideUserInterfaceStyle = uiUserInterfaceStyle
```
vadymmarkov: Nice, is it actually possible to override user interface style?

vadymmarkov:
BTW, I don't know if you have considered this already, but with the minimum target set to iOS 11, we can also use asset catalog colors to support dark mode. 

Screenshot 2019-08-14 at 10 29 59

3lvis:
Jonas mentioned that as another possibility. I like the usage of `UIColor(dynamicProvider: { (traitCollection` tho

osanoj:
I think with the asset catalog colors it will be easier to have dark mode on pre iOS 13 devices. If we want to support that.

Sources/DNA/Color/UserInterfaceStyle.swift
osanoj: Yes, it seems to work well, you can do it for individual views/viewcontrollers too.

Originally I thought that `UITraitCollection.current = updatedTraits` would be enough but it wasn't (so now I'm not 100% that part is actually needed).

3lvis:
I really like this implementation, I have found that sometimes the navigation bar doesn't update, tho. 🤔

osanoj:
Strange, on iOS 13 or earlier version?

3lvis:
On iOS 13, trying again with Xcode 11 Beta 6

osanoj:
Closing this since we have #590 now

PR #519, opened 2019-08-14
Custom templates

pimms:
This is a semi-big one, let me know if you want me to give you a quick F2F walkthrough.

# Why?

The pre-defined messages introduced in July has been quite successful, so we want to iterate on the solution and provide users with a way of adding their own custom pre-defined messages.

This is still quite experimental, so to iterate quickly and experiment wildly all data is intended to be stored on device. This does however not affect the implementation in FinniversKit.

# What?

When explicitly enabled and configured, a _"customize"_ button is shown on the `MessageFormToolbar`. The overview-view displays the existing user-defined message templates, and from this view you are able to create new ones.

Notes:
 - Templates are always shown in the order they were added
 - If either the message-view or edit template-view has uncommitted changes, dismissive actions on the `BottomSheet` will be interrupted 
 - We distinguish between "default templates" and "custom templates" — default templates cannot be deleted or edited
 - This changeset breaks the existing integration from the FINN-app because of the new properties added in `MessageFormViewModel`

All UX and design has been iterated upon and verified by Ingrid.

# Show me

### Add a new template
add-template

### Editing a custom template
edit-template

### Cancelling a new template
add-template-cancel

### Delete all custom templates
delete-all

### Delete via 'edit' buton
delete-edit

### Delete via swipe-from-right
delete-right

### Dismissal handling when editing a template
dismiss-on-edittemplate

### Dismissal handling when adding a new template
dismiss-on-newtemplate

### Dismissal handling on overview
dismiss-on-overview


Sources/Fullscreen/MessageForm/CustomTemplates/MessageTemplateOverviewViewController.swift
```diff
@@ -0,0 +1,319 @@
...
+        case 1:
+            return templateStore.customTemplates.count
+        default:
+            return 0
+        }
+    }
+
+    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        switch indexPath.section {
+        case 0:
```
vadymmarkov: We could also create `enum Section` and give sections some names.

Sources/Fullscreen/MessageForm/CustomTemplates/MessageTemplateOverviewViewController.swift
```diff
@@ -0,0 +1,319 @@
...
+            return UITableViewCell()
+        }
+    }
+
+    func tableView(_ tableView: UITableView, canEditRowAt indexPath: IndexPath) -> Bool {
+        return indexPath.section == 1
+    }
+
+    public func tableView(_ tableView: UITableView, titleForFooterInSection section: Int) -> String? {
+        if section == 0 && templateStore.customTemplates.count == 0 {
```
vadymmarkov: Could also be `templateStore.customTemplates.isEmpty`

Sources/Fullscreen/MessageForm/CustomTemplates/MessageTemplateEditViewController.swift
```diff
@@ -0,0 +1,122 @@
...
+        let saveButton = UIBarButtonItem(title: viewModel.saveButtonTitle, style: .done, target: self, action: #selector(saveButtonTapped))
+        saveButton.isEnabled = false
+        navigationItem.rightBarButtonItem = saveButton
+
+        if let existingTemplate = viewModel.existingTemplate {
+            messageInputTextView.text = existingTemplate.text
+        }
+    }
+
+    private func registerForNotifications() {
```
vadymmarkov: Should we unregister as well?

pimms:
Thanks! Yeah, it's another "too-large-for-comfort"-PR from me, I find it really difficult to combine continuous integration with UX & UI iterations. I'm working on it though! 😅 At least the messaging-template feature as a whole was divided in two — with and without custom templates... 🙂

PR #524, opened 2019-08-16
Possibility to set size of HorizontalSlideTransition

hakloev:
# Why?

When enabling Charcoal on iPad there arose a need to be able to adjust the width of the `HorizontalSlideTransition` since 60% was to wide. 

# What?

Created a new initializer for `HorizontalSlideController` that accepts a container percentage. Also added an override for the default initializer to still set the default percentage. Furthermore, it's added a convenience initializer to `HorizontalSlideTransition` to set the container percentage that should be passed to the `HorizontalSlideController`.

Is this an appropriate way of passing this optional argument? 

# Show me

| Default initializer | Convenience initializer with 40% |
| --- | --- |
| Simulator Screen Shot - iPhone Xs - 2019-08-16 at 08 42 53 | Simulator Screen Shot - iPhone Xs - 2019-08-16 at 08 42 35 |


Sources/Components/HorizontalSlide/HorizontalSlideController.swift
```diff
@@ -28,6 +28,24 @@ class HorizontalSlideController: UIPresentationController {
...
+    ///   - presentedViewController: The view controller being presented modally.
+    ///   - presentingViewController: The view controller whose content represents the starting point of the transition.
+    ///   - containerPercentage: The width of the container for the view controller being presented modally
+    init(presentedViewController: UIViewController, presenting presentingViewController: UIViewController?, containerPercentage: CGFloat) {
+        self.containerPercentage = containerPercentage
+        super.init(presentedViewController: presentedViewController, presenting: presentingViewController)
+    }
+
+    override init(presentedViewController: UIViewController, presenting presentingViewController: UIViewController?) {
+        self.containerPercentage = UIDevice.current.userInterfaceIdiom == .pad ? 0.60 : 0.85
```
shredlocker: What if you put this in the `HorizontalSlideTransition` class instead as a default value? Then you don't really need this init and you can remove the optional below.

hakloev: That sounds like a much better idea!

Sources/Components/HorizontalSlide/HorizontalSlideTransition.swift
```diff
@@ -5,10 +5,32 @@ import UIKit
 }
 
 public class HorizontalSlideTransition: NSObject, UIViewControllerTransitioningDelegate {
+
     @objc public weak var delegate: HorizontalSlideTransitionDelegate?
 
+    private let containerPercentage: () -> CGFloat
```
shredlocker: I think a would have defined a struct instead. Something like:

```swift
struct ContainerSize {
    let portrait: CGFloat
    let landscape: CGFloat?

    static var `default`: ContainerSize {
        if iPad {
            return ContainerSize(portrait: 0.4, landscape: 0.3)
        } else {
            return ContainerSize(portrait: 0.6)
        }
}
```
And then with an init
`public init(width size: ContainerSize)`

hakloev: Didn't even think of this as a possibility! This looks way better 👍

Sources/Components/HorizontalSlide/HorizontalSlideTransition.swift
```diff
@@ -5,10 +5,50 @@ import UIKit
...
+        }
+    }
+
     @objc public weak var delegate: HorizontalSlideTransitionDelegate?
 
+    private let containerSize: ContainerSize
+
+    // MARK: - Init
+
+    public init(size containerSize: ContainerSize) {
```
shredlocker: I would keep only `size` or `containerSize` here, to me it doesn't really matter which one it is ;)

shredlocker:
Looks good! Nice work 😄

PR #525, opened 2019-08-16
Fullscreen gallery carousel

shredlocker:
# Why?

People have requested to be able to loop the gallery.

# What?

After trying to get the `CarouselView` in to the `FullscreenGalleryViewController`, the problems kept piling up.

This approach seems to work fine and it is of course very little code.

# Show me
 
simple-carousel-view


shredlocker:
Might not be a point to keep the `CarouselView` in `FinniversKit` since we're not reusing it here, or what do you think @osanoj ? It is not a lot of code needed to implement that carousel on the object page anyway.

Sources/Fullscreen/FullscreenGallery/FullscreenGalleryViewController.swift
```diff
@@ -179,24 +181,37 @@ public class FullscreenGalleryViewController: UIPageViewController {
     private func currentImageViewController() -> FullscreenImageViewController? {
         return viewControllers?.first as? FullscreenImageViewController
     }
+
+    // Find remainder of both positiv and negative numbers
+    private func remainder(_ number: Int, divider: Int) -> Int {
```
osanoj: Is it really the remainder you are calculating here? This is a quite confusing thing when dealing with negative number, quick googling didn't help me much either. Except that Swift `%` "modulo operand" actually is a remainder operation, but yeah, then it is the negative thing too 🤕 

Could we maybe just rename it to what you actually use it for instead? I guess something like `imageIndexForPage`?

shredlocker: I might not have to think about the remainder of negative number after all, I'll check it out.

shredlocker: What do you think about it now @osanoj ?

osanoj: Much better, now I don't have to scratch my head and think 😉

PR #526, opened 2019-08-16
Favorite folders: improve edit functionality

vadymmarkov:
# Why?

We decided to use the default edit checkmarks and make transition animations smoother. 

# What?

- Use default edit checkmarks
- Tweak animations when transitioning to/from edit mode
- Add an option to enable "swipe to delete" for table view cells 
- Show detail text when it's present
- Add new method to select/deselect all cells

# Show me

| Before | After |
| --- | --- |
| finnivers_before | finnivers_after |


Sources/Recycling/ListViews/FavoriteFolders/Subviews/FavoriteFolderSelectableViewCell.swift
```diff
@@ -41,147 +48,75 @@ public class FavoriteFolderSelectableViewCell: RemoteImageTableViewCell {
...
         bringSubviewToFront(editModeView)
-        setCheckmarksHidden(true)
+
+        contentView.alpha = state == .showingEditControl && !isEditable ? 0.5 : 1
+        editModeView.isHidden = isEditable
+        titleLabel.font = titleLabelDefaultFont
+        checkmarkImageView.isHidden = true
     }
 
     public override func didTransition(to state: UITableViewCell.StateMask) {
```
bstien: Guess this method can be removed now, since it only calls `super`

PR #529, opened 2019-08-21
Add directions to CalloutView

hakloev:
# Why?

When enabling Charcoal on iPad there arose a need to be able to add a `CalloutView` with the arrow pointing downwards. 

# What?

Added an enum `Direction` and applies a transformation to the arrow according to the direction. Also updates the constraints accordingly. For now, `down` is the only new direction. 

# Show me

| Before | After |
| --- | --- |
| Simulator Screen Shot - iPhone Xs - 2019-08-21 at 10 24 54 | Simulator Screen Shot - iPhone Xs - 2019-08-21 at 10 20 15 |


Sources/Components/Callout/CalloutView.swift
```diff
@@ -5,6 +5,23 @@
...
+        case up
+        case down
+    }
+
+    private let direction: Direction
+
+    private var arrowRotationTransformation: CGAffineTransform? {
+        switch direction {
+        case .up:
+            return nil
```
bstien: How about returning `.identity` here? So we don't have to check for `nil`-values

3lvis:
😍

PR #530, opened 2019-08-21
Feature: save search view

fespinoza:
# Why?

_This view was previously part of the finn app_
In order to add support to allow email notifications for saved searches when creating them.

# What?

- Add a new `SaveSearchView` withs is view model, delegate and demo pages
- Add an internal `LabelSwitchView` which looks like a `UITableViewCell` but is just a plain view when using a `UITableView` with 2 static rows would be an overkill

# Show me

Screenshot 2019-08-21 at 10 48 29
Screenshot 2019-08-21 at 10 49 14


Sources/Components/SaveSearchView/LabelSwitchView.swift
```diff
@@ -0,0 +1,88 @@
+//
+//  Copyright © 2019 FINN AS. All rights reserved.
+//
+
+protocol LabelSwitchViewDelegate: AnyObject {
+    func labelSwitchViewValueChanged(_ labelSwitchView: LabelSwitchView)
+}
+
+class LabelSwitchView: UIView {
```
osanoj: Is it possible to use `SwitchView` instead of making a new one?

fespinoza: I will try it out, i didn't see it around

UnitTests/ComponentViewTests.swift
```diff
@@ -17,9 +17,9 @@ class ComponentViewTests: FBSnapshotTestCase {
     override class func tearDown() {
         super.tearDown()
 
-        if ComponentViewTests.allViews.count > 0 {
-            fatalError("Not all elements were implemented, missing: \(ComponentViewTests.allViews.map { $0.rawValue }.joined(separator: ", "))")
-        }
+//        if ComponentViewTests.allViews.count > 0 {
+//            fatalError("Not all elements were implemented, missing: \(ComponentViewTests.allViews.map { $0.rawValue }.joined(separator: ", "))")
+//        }
```
osanoj: Why remove this? 🤔

fespinoza: It was just when writing the test, i will restore it :P

Sources/Components/SaveSearchView/LabelSwitchView.swift
fespinoza: So, it wasn't used anywhere relevant (both in FinniversKit and the finn app) just in `BottomSheetMechanicsDemoViewController` within FinniversKit

Then i suggest to replace the current SwitchView with this one

osanoj: Sounds good to me 👍

fespinoza:
I did the change, by replacing the implementation of `SwitchView`, but keeping the name, which makes it in strict sense a breaking change (although the view is not used in the FINN app)

PR #531, opened 2019-08-21
Drop iOS 10

fespinoza:
# Why?

Because being modern and cool (also iOS 10 is from 2016).
Setting iOS 11.2 as minimum target, following the Finn app

# What?

- update the deployment target of the Demo app
- update the deployment target of the FinniversKit target
- update the podspec


fespinoza:
Is there any potential issue with this @3lvis and @osanoj ?

3lvis:
Looks like it passed now 🤔

3lvis:
https://github.com/finn-no/FinniversKit/pull/532

fespinoza:
> Looks like it passed now 🤔

I had forgotten to update the deployment target for the tests

PR #534, opened 2019-08-22
Modernize implementation of LoginEntryView

fespinoza:
# Why?

As an example of setting constraints towards the `layoutMarginsGuide`

# What?

- remove `#if available(iOS 11)` from `LoginEntryView`
- set constraints for `LoginEntryDialogueView` with `layoutMarginsGuide`

# Show me

No changes


Sources/Fullscreen/LoginEntryView/LoginEntryDialogueView.swift
```diff
@@ -85,35 +85,39 @@ class LoginEntryDialogueView: UIView {
 
     private func setup() {
         layer.cornerRadius = 33
+        layoutMargins = UIEdgeInsets(
+            top: .veryLargeSpacing, leading: .largeSpacing, bottom: .largeSpacing, trailing: .largeSpacing
+        )
```
pimms: The docs specify to use `directionalLayoutMargins` for iOS 11+ though, should we be using `layoutMargins`?

pimms: https://developer.apple.com/documentation/uikit/uiview/1622566-layoutmargins

Sources/Fullscreen/LoginEntryView/LoginEntryDialogueView.swift
```diff
@@ -85,35 +85,39 @@ class LoginEntryDialogueView: UIView {
...
-            titleLabel.leadingAnchor.constraint(equalTo: leadingAnchor, constant: .largeSpacing),
-            titleLabel.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -.largeSpacing),
+            titleLabel.leadingAnchor.constraint(equalTo: margins.leadingAnchor),
+            titleLabel.trailingAnchor.constraint(equalTo: margins.trailingAnchor),
 
             detailLabel.topAnchor.constraint(equalTo: titleLabel.bottomAnchor, constant: .mediumSpacing),
-            detailLabel.leadingAnchor.constraint(equalTo: leadingAnchor, constant: .mediumSpacing + .largeSpacing),
-            detailLabel.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -.mediumSpacing - .largeSpacing),
+            detailLabel.leadingAnchor.constraint(equalTo: margins.leadingAnchor, constant: .mediumSpacing),
+            detailLabel.trailingAnchor.constraint(equalTo: margins.trailingAnchor, constant: -.mediumSpacing),
```
pimms: I think it's less obvious what the actual inset is when we're anchoring to the layout-margins — which already has an inset, and then adding additional insets onto that. For the other views which anchor directly to the margins, I love it _(I didn't know this even existed tbh)_.

Sources/Fullscreen/LoginEntryView/LoginEntryDialogueView.swift
fespinoza: that's a good point

Sources/Fullscreen/LoginEntryView/LoginEntryDialogueView.swift
fespinoza: We discussed a bit this during the iOS meeting today.

I think setting the constraint to the margin (with extra constants if needed) is a cleaner way to use this overall than specify the full inset.

This is for views in a case like:

Screenshot 2019-08-22 at 13 36 35

Where all the content of a given view is with a given inset (margins), then changing that given inset for the full view is easier than updating each single constraint.

For the case of views that require an extra inset I think the margins apply too and the presence of the constant means an extra inset. Then on-and-all I think the exact amount of the inset for normal and extra inset views is not that relevant, as the view controls the amount for the subviews through ~~`layoutMargins`~~ `directionalLayoutMargins`

pimms: Sounds reasonable! 🙂

PR #535, opened 2019-08-22
Added IdentityView

pimms:
# Why?

We're going to gradually transition the views from the T&R SDK into FinniversKit for unified code style & consistent architecture. This is the first view being moved into FK.

# What?

Added the `IdentityView`. This view will be shown on the object- and public-profile-pages.

# Show me

### iPad
iPadL
iPadP

### iPhone SE
iSE

### iPhone 8
i8


Sources/Components/Profile/IdentityView.swift
```diff
@@ -0,0 +1,152 @@
...
+            subtitleLabel.topAnchor.constraint(equalTo: profileNameLabel.bottomAnchor, constant: .verySmallSpacing),
+            subtitleLabel.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -.mediumLargeSpacing),
+        ])
+
+        if !descriptionLabel.isHidden {
+            addSubview(descriptionLabel)
+            NSLayoutConstraint.activate([
+                descriptionLabel.leadingAnchor.constraint(equalTo: leadingAnchor, constant: .mediumLargeSpacing),
+                descriptionLabel.topAnchor.constraint(greaterThanOrEqualTo: profileImageView.bottomAnchor, constant: .mediumLargeSpacing),
+                descriptionLabel.topAnchor.constraint(greaterThanOrEqualTo: subtitleLabel.bottomAnchor, constant: .mediumLargeSpacing),
```
pimms: These two constraints exists to make sure that we have the correct spacing between the `subtitleLabel` and `descriptionLabel` in case the device is using a larger than default font-size, as the `maxY` of the `subtitleLabel` would then exceed that of `profileImageView`.

If it's too cryptic, I'll add a comment about this 🙂

3lvis:
Wait, this wasn't a table view?!

pimms:
No, the identity & public profile pages were not table views. The feedback listings used to be collection views, but are currently table views within the SDK.

pimms:
Oops, forgot snapshot tests & the fact that we should download the image through the delegate. Fixing.

pimms:
I think I found a good solution to handling how the profile image is loaded — if it's _aaaalmost_ immediately available, we don't want to flicker the default image and the loaded one, so the default profile image will be set:
 - Immediately, if no `profileImageUrl` is defined
 - On `profileImageUrl` download error
 - Faded in after 200ms if `profileImageView.image` is still nil at that point

The fade-in _(performed in `setupDefaultProfileImage()`)_ is thread-safe because the check `profileImageView.image == nil` is performed on the main thread.

pimms:
I had to make quite a few changes, as the SDK will _(in almost all cases)_ not have the `IdentityViewModel` ready when the `IdentityView`-instances has been created and added to the view hierarchy.

Sources/Components/Profile/IdentityView.swift
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    // MARK: - UI properties
+
+    private let profileImageSize: CGFloat = 40.0
+    private lazy var defaultProfileImage = UIImage(named: FinniversImageAsset.avatar)
+
+    private lazy var profileImageView: UIImageView = {
+        let imageView = UIImageView()
+        imageView.translatesAutoresizingMaskIntoConstraints = false
```
bstien: Could rather be `let imageView = UIImageView(withAutoLayout: true)`

Sources/Components/Profile/IdentityView.swift
```diff
@@ -0,0 +1,213 @@
...
+    public required init(viewModel: IdentityViewModel?) {
+        self.viewModel = viewModel
+        super.init(frame: .zero)
+        setupSubviews()
+        addTapListener()
+        setupDefaultProfileImage()
+        viewModelChanged()
+    }
+
+    private func setupSubviews() {
```
bstien: This should be named `setup()`

Sources/Components/Profile/IdentityView.swift
```diff
@@ -0,0 +1,213 @@
...
+        fatalError("\(#function) not implemented")
+    }
+
+    public required init(viewModel: IdentityViewModel?) {
+        self.viewModel = viewModel
+        super.init(frame: .zero)
+        setupSubviews()
+        addTapListener()
+        setupDefaultProfileImage()
+        viewModelChanged()
```
bstien: Could the calls to these methods be moved to `setup()` instead?

Demo/Components/Profile/IdentityDemoView.swift
```diff
@@ -0,0 +1,106 @@
...
+    }
+}
+
+extension IdentityDemoView: IdentityViewDelegate {
+    func identityViewWasTapped(_ identityView: IdentityView) {
+        print("Identity view of '\(identityView.viewModel?.displayName ?? "<nil>")' was tapped")
+    }
+
+    public func identityView(_ identityView: IdentityView, loadImageWithUrl url: URL, completionHandler: @escaping (UIImage?) -> Void) {
+        let task = URLSession.shared.dataTask(with: url) { data, _, _ in
```
bstien: Just in case, we should add a delay to this fetch so we don't mess up the snapshot tests

PR #539, opened 2019-08-23
Remove conditional code `if #available(iOS 11.0)`

fespinoza:
# Why?

They are not necessary and cleaning code is nice 😊

# What?

As that is the only code path available right now.

Then also removed:
- `compatibleTopAnchor`
- `compatibleBottomAnchor`
- `safeLayoutGuide`

# Show me

No UI changes

### DIFF
--- a/Sources/Components/BottomSheet/Controller/BottomSheet.swift
+++ b/Sources/Components/BottomSheet/Controller/BottomSheet.swift
@@ -129,30 +129,12 @@ public class BottomSheet: UIViewController {
         fatalError("init(coder:) has not been implemented")
     }
 
-    // Only necessary if iOS < 11.0
-    // ------------
-    public override func viewDidLayoutSubviews() {
-        super.viewDidLayoutSubviews()
-        if #available(iOS 11.0, *) {
-            return
-        } else {
-            let path = UIBezierPath(roundedRect: view.bounds,
-                                    byRoundingCorners: [.topLeft, .topRight],
-                                    cornerRadii: CGSize(width: cornerRadius, height: cornerRadius)).cgPath
-            maskLayer.path = path
-            view.layer.mask = maskLayer
-        }
-    }
-    // ------------
-
     public override func viewDidLoad() {
         super.viewDidLoad()
         view.backgroundColor = rootViewController.view.backgroundColor ?? .milk
         view.clipsToBounds = true
-        if #available(iOS 11.0, *) {
-            view.layer.cornerRadius = cornerRadius
-            view.layer.maskedCorners = [.layerMinXMinYCorner, .layerMaxXMinYCorner]
-        }
+        view.layer.cornerRadius = cornerRadius
+        view.layer.maskedCorners = [.layerMinXMinYCorner, .layerMaxXMinYCorner]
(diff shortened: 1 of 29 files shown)


Sources/Components/BottomSheet/Controller/BottomSheet.swift
```diff
@@ -129,30 +129,12 @@ public class BottomSheet: UIViewController {
...
-    // ------------
-    public override func viewDidLayoutSubviews() {
-        super.viewDidLayoutSubviews()
-        if #available(iOS 11.0, *) {
-            return
-        } else {
-            let path = UIBezierPath(roundedRect: view.bounds,
-                                    byRoundingCorners: [.topLeft, .topRight],
-                                    cornerRadii: CGSize(width: cornerRadius, height: cornerRadius)).cgPath
-            maskLayer.path = path
```
shredlocker: You can remove the `maskLayer` property as well ;)

Sources/Extensions/UIKit/UIViewExtensions.swift
```diff
@@ -10,30 +10,6 @@ public extension UIView {
...
-
-    var compatibleBottomAnchor: NSLayoutYAxisAnchor {
-        if #available(iOS 11.0, *) {
-            return safeAreaLayoutGuide.bottomAnchor
-        } else {
-            return bottomAnchor
-        }
-    }
-
-    var safeLayoutGuide: UILayoutGuide {
```
vadymmarkov: It's probably a breaking change, we use it in the app and in Charcoal I think.

fespinoza: I see, i opted to remove it as I could have kept it and just return the `safeLayoutAreaGuide`, but that (together with the other stuff) mean that we use custom properties instead of the vanilla iOS properties.

I guess we need to update charcoal and any other dependency to FinniversKit to drop iOS 10 as well.

We can potentially restore this if it's a major issue for fixing the app release

shredlocker: But we are not including this in the current release right? So the current release is not broken

fespinoza: yeah, this is not part of the current release

osanoj: Yes, we just need to update main app and Charcoal so they will not break next time someone wants to update FinniverKit

fespinoza: the main app is on iOS 11+, Now to compile it thought we need Charcoal, Finnbox to be updated to iOS 11+.

Then I guess we should release FinniversKit, the unreleased changes right now are:

```
## Changes since 3.2.0 to master (squash)

- #538: Extend empty view with image support
- #540: Handle page control value changes
- #539: Remove conditional code `if #available(iOS 11.0)`
- #536: Integration with `pr_changelog`
- #531: Drop iOS 10
```

Would that mean we go for `4.0`?

osanoj: Yes 👍 

And then update main app and Charcoal to use 4.0

PR #542, opened 2019-08-23
Embed `SaveSearchView` content in a UIScrollView

fespinoza:
# Why?

So for the iPhone SE when presenting the bottom sheet containing the save search view, then you can scroll in its content as the keyboard blocks some of it.

# What?

- Embed `SaveSearchView` content in a UIScrollView
- Replace `SaveSearchViewDemoView` for a view controller
- Present the save search demo view controller with a `UINavigationController` in a `BottomSheet`
- Adjust the content size for the scroll view when the phone keyboard appears/dissapears

# Show me

iPhone SE

Aug-23-2019 14-09-18


Sources/Components/SaveSearchView/SaveSearchView.swift
```diff
@@ -75,6 +75,19 @@ public class SaveSearchView: UIView {
         return line
     }()
 
+    private lazy var contentView: UIView = {
```
vadymmarkov: Could be simply `private lazy var contentView = UIView(withAutoLayout: true)`

PR #543, opened 2019-08-23
Feature: add favorite action sheet

vadymmarkov:
# Why?

It should be possible to perform actions on favorite ads, such us delete and add notes.

# What?

- Add `FavoriteActionView`, which can be used both separately or within `FavoriteActionSheet` view controller

- Increate bottom margin for all bottom sheets used in favorites

# Show me

| View | Bottom sheet |
| --- | --- |
| Simulator Screen Shot - iPhone Xs - 2019-08-23 at 15 52 52 | Simulator Screen Shot - iPhone Xs - 2019-08-23 at 15 52 41 |


Sources/Recycling/ListViews/FavoriteActions/Views/FavoriteActionHeaderView.swift
```diff
@@ -0,0 +1,130 @@
...
+    // MARK: - Setup
+
+    func configure(withImage image: UIImage?, detailText: String, accessoryText: String) {
+        imageView.image = image
+        detailTextLabel.text = detailText
+        accessoryLabel.text = accessoryText
+    }
+
+    private func setup() {
+        backgroundColor = .milk
```
bstien: Shouldn't the background be a light shade of gray?

vadymmarkov: Apparently it should be a blur view 😄 I'll fix it.

vadymmarkov:
This is how it looks like with a blur view:

Simulator Screen Shot - iPhone Xs - 2019-08-26 at 10 44 55

bstien:
⛴⛴⛴

PR #548, opened 2019-08-27
Fix: `FavoriteAdTableViewCell` constraints and `FavoriteAdViewModel` refactoring

bstien:
# Why?
We weren't taking deleted ads into account when we created `FavoriteAdTableViewCell` in the sense that those ads doesn't contain any info at all. This would make the cells look silly because of the `stackView`'s bottomConstraint.

We also decided to refactor `FavoriteAdViewModel` from a protocol and into a struct.

# What?
- Set correct constraints against `contentView.bottomAnchor`.
- Refactor `FavoriteAdViewModel` into a struct.
  - Added properties `lastUpdated`, `addedToFolderDate` and `titleColor`.

# Show me

| Before | After |
| --- | --- |
| Screenshot 2019-08-27 at 09 51 16 | Screenshot 2019-08-27 at 14 49 56 |


Demo/Fullscreen/FavoriteAdsListView/FavoriteAdsListDemoViewHelper.swift
```diff
@@ -4,35 +4,47 @@
...
-                       titleText: title,
-                       descriptionPrimaryText: primaryDescriptions[index],
-                       descriptionSecondaryText: secondaryDescriptions[index],
-                       imagePath: imagePaths[index],
-                       ribbonStyle: ribbonStyles[index].style,
-                       ribbonTitle: ribbonStyles[index].title,
-                       addedToFolderDate: addedToFolderDates[index])
+            guard let title = title else {
+                return FavoriteAdViewModel.deletedAd(addedToFolderDate: addedToFolderDates[index],
+                                            ribbonStyle: ribbonStyles[index].style,
```
vadymmarkov: Something is wrong with indentation here.

PR #550, opened 2019-08-28
Use closures for Toast actions

3lvis:
# Why?

Dealing with actions in our current toasts is quite cumbersome, at the moment we are using delegates which gives resembles the old UIAlertView way of doing things.

# What?

This PR makes use of ToastAction which resembles the UIAlertController way of doing things. This makes actions easier to implement since the action is where the toast is created.

# Show me

#### Before

```swift
let successButtonToast = ToastView(style: .successButton)
successButtonToast.buttonText = "Action success"
successButtonToast.delegate = ToastClass()

class ToastClass: NSObject, ToastViewDelegate {
    func didTap(toastView: ToastView) {
        print("Toast view tapped \(toastView.style)")
    }
}
```

#### After

```swift
let successButtonToast = ToastView(style: .successButton)
successButtonToast.text = "Action success"
successButtonToast.action = ToastAction(title: "Action") {
    print("Action button tapped")
}
```

### DIFF
--- a/Demo/Components/Toast/ToastDemoView.swift
+++ b/Demo/Components/Toast/ToastDemoView.swift
@@ -4,16 +4,6 @@
 
 import FinniversKit
 
-class ToastClass: NSObject, ToastViewDelegate {
-    func didTapActionButton(button: UIButton, in toastView: ToastView) {
-        print("Button tapped \(toastView.style)")
-    }
-
-    func didTap(toastView: ToastView) {
-        print("Toast view tapped \(toastView.style)")
-    }
-}
-
 public class ToastDemoView: UIView {
     private lazy var bottomToastButton: Button = {
         let button = Button(style: .callToAction)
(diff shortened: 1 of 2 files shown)


bstien:
There is a distinct difference between tapping the toast and tapping the button, especially if we're also removing the swipe-to-dismiss gesture on it.

We should probably add an action for swiping/tapping the Toast so it can dismissed if the user wants to ignore the action suggested by the button.

3lvis:
We discussed adding back the pull down to dismiss and drop the tap to dismiss since it's error-prone especially when we are already handling taps for actions.

3lvis:
Funny thing that I found in our code base is that we are presenting the toast with different time outs: 3, 4, 5, 10 seconds

osanoj:
> Funny thing that I found in our code base is that we are presenting the toast with different time outs: 3, 4, 5, 10 seconds

It sort of makes sense to have some time differences, but 3, 4 and 5 seconds seems a bit much 😂 

Maybe we can have an enum with `short` (3 or 4), `medium` (5 or 6), `long` (10)? 

Or perhaps `normal`, `long` and `custom(seconds: Int)` 🤔

3lvis:
That's interesting, using an enum for this makes a lot more sense than using hardcoded numbers

3lvis:
Re-introduced swipe down to dismiss :)

PR #552, opened 2019-08-29
Support hiding the description on IdentityView

pimms:
# UPDATE

I also need the option to override the profile image directly from an `UIImage`-instance, so I added a field `profileImage: UIImage?` to the `IdentityViewModel`. If present, the `IdentityView` will not trigger a network call when setting up the profile image.

# Why?

I forgot about this detail — whether or not we should show the profile description should be configurable to make the existing view easier to replace. The existing implementation supports it in this way.

Screen Shot 2019-08-29 at 11 30 35

### DIFF
--- a/Demo/Components/Profile/IdentityDemoView.swift
+++ b/Demo/Components/Profile/IdentityDemoView.swift
@@ -8,7 +8,7 @@ class IdentityDemoView: UIView, Tweakable {
 
     // MARK: - UI properties
 
-    private var identityViews: [(IdentityView, IdentityViewModel)] = []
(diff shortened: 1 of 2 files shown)


pimms:
The integration from the FINN app does things way differently than what I expected from the T&R demo, so I had to add the option to override the profile image through a `UIImage` _(instead of `URL` which is the current only option)_. This is for the use-case where the user wants to change their profile image and we don't have (or want) a URL.

The intended usage of this would be:
```
func uploadProfileImage(_ image: UIImage) {
    uploadProfileImage(image) {
        self.viewModel.profileImage = image
        self.identityView.viewModel = self.viewModel
    }
}
```

PR #557, opened 2019-09-05
Week 36 Feature: Large Ads in Frontpage

3lvis:
# Why?

We are running an experiment to track the increase in the effect of using large ads for promoted ads.

# What?

Ads support for displaying "nicer" ads, they will show in one column in iPad for now but they take full width in iPhones.

# Show me

| iPhone | iPad |
| --- | --- |
| Simulator Screen Shot - iPhone X - 2019-09-05 at 07 52 29 | Simulator Screen Shot - iPad Air 2 - 2019-09-05 at 07 56 29 |

### DIFF
--- a/Sources/Recycling/GridViews/Ads/Cell/AdsGridPromotedViewCell.swift
+++ b/Sources/Recycling/GridViews/Ads/Cell/AdsGridPromotedViewCell.swift
@@ -1,0 +1,106 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import UIKit
+
+public class AdsGridPromotedViewCell: AdsGridViewCell {
+    override public func setup() {
+        isAccessibilityElement = true
+
+        backgroundColor = .milk
+        imageView.contentMode = .scaleAspectFill
+        subtitleLabel.textColor = .milk
+        titleLabel.textColor = .milk
+        accessoryLabel.textColor = .milk
+        imageDescriptionView.backgroundColor = .clear
+        addSubview(imageBackgroundView)
+        imageBackgroundView.addSubview(imageView)
+        imageView.fillInSuperview()
+        imageBackgroundView.fillInSuperview()
+
+        let topGradient = CAGradientLayer()
+        topGradient.frame = bounds
+        topGradient.colors = [
+            UIColor.black.withAlphaComponent(0.65).cgColor,
+            UIColor.clear.cgColor
+        ]
+        topGradient.locations = [0, 0.3]
+        let topGradientImageView = UIImageView(withAutoLayout: true)
+        topGradientImageView.image = UIImage.imageWithLayer(layer: topGradient)
+        imageBackgroundView.addSubview(topGradientImageView)
+
+        let bottomGradient = CAGradientLayer()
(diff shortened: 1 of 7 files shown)


Sources/Recycling/GridViews/Ads/Cell/AdsGridViewCell.swift
```diff
@@ -17,63 +17,63 @@ public protocol AdsGridViewCellDelegate: AnyObject {
...
-    private static let subtitleHeight: CGFloat = 17.0
-    private static let subtitleTopMargin: CGFloat = 6.0
-    private static let accessoryHeight: CGFloat = 14.0
-    private static let margin: CGFloat = 8.0
-    private static let cornerRadius: CGFloat = 8.0
-    private static let imageDescriptionHeight: CGFloat = 35.0
-    private static let iconSize: CGFloat = 23.0
-
-    private lazy var imageBackgroundView: UIView = {
+    public static let titleHeight: CGFloat = 20.0
```
bstien: I guess all the modifiers changed to `public` in this file could be made `internal`. Or do we need them in the app?

Sources/Recycling/GridViews/Ads/Cell/AdsGridPromotedViewCell.swift
```diff
@@ -0,0 +1,106 @@
...
+            favoriteButton.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -.smallSpacing),
+            favoriteButton.widthAnchor.constraint(equalToConstant: 34),
+            favoriteButton.heightAnchor.constraint(equalTo: favoriteButton.heightAnchor)
+            ])
+    }
+}
+
+extension UIImage {
+    class func imageWithLayer(layer: CALayer) -> UIImage {
+        UIGraphicsBeginImageContextWithOptions(layer.bounds.size, layer.isOpaque, UIScreen.main.scale)
```
bstien: I'm not sure if I'm correct, but wouldn't the gradient be pixelated/blurry when the cell is reused and the other image is smaller/bigger than the first one? Because it's an image and the size is set based on the layer's size

3lvis: If I'm not mistaken when you stretch a gradient it works fine since it fades as you stretch it. Since gradients are blurry by default. Maybe someone else knows more about it.

bstien: You're probably right. Forgot that the gradient is transparent as well, which helps.

Sources/Recycling/GridViews/Ads/GridView/AdsGridView.swift
```diff
@@ -229,6 +244,12 @@ extension AdsGridView: AdsGridViewLayoutDelegate {
             return defaultImageSize.height / defaultImageSize.width
         }
 
+        var height = model.imageSize.height
+        // We want large ads to show in a more prominent way,
+        // hence the change in image height ratio.
+        if model.isLargeAd {
+            height *= 1.8
```
osanoj: I guess `height` should either be used or this code removed if it is no longer needed

PR #563, opened 2019-09-09
Fix SwiftLint warnings

vadymmarkov:
# Why?

Because there were a lot of warnings after building the project, especially "Trailing Whitespace Violation: Lines should not have trailing whitespace. (trailing_whitespace)".

# What?

Fixed all `SwiftLint` warnings. 

**N.B. Since we have enabled linting rules for the project, it would be nice if everyone has `SwiftLint` installed on their machines, so the warnings annoy everyone 😄**

# Show me

No UI changes.

### DIFF
--- a/Demo/Components/Profile/IdentityDemoView.swift
+++ b/Demo/Components/Profile/IdentityDemoView.swift
@@ -8,7 +8,7 @@ class IdentityDemoView: UIView, Tweakable {
 
     // MARK: - UI properties
 
-    private var identityViews: [(IdentityView,ViewModel)] = []
+    private var identityViews: [(IdentityView, ViewModel)] = []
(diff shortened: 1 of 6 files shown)


3lvis:
🗜

bstien:
I think removal of lines with only whitespace can be turned on in both Xcode and AppCode

vadymmarkov:
Yes, in Xcode it's under Text Editing preferences:

Screenshot 2019-09-09 at 11 00 37

bstien:
| Xcode | AppCode |
| --- | --- |
| Screenshot 2019-09-09 at 11 01 17 | Screenshot 2019-09-09 at 11 00 37 |

PR #564, opened 2019-09-09
Feature: VerificationView

robinsalehjan:
**NOTE**: The `BankID` icon is remaining and I will follow up with a PR when the asset is created.

# Why?

- Certain market segments require that the user is `BankID` verified in order to publish an ad.
- We have no good way of prompting the user to become `BankID` verified.

# What?

- Create a new `VerificationView` and companioning `VerificationViewActionSheet`

# Show me
Screenshot 2019-09-09 at 13 05 47


bstien:
Screenshot 2019-09-09 at 13 18 22

Didn't we add some logic to make the notch have the same background as the rootView? 🤔 Or was it only for the demo, @3lvis?

robinsalehjan:
Any ideas on the background of the notch? I would like to merge so I can continue iterating.

robinsalehjan:
The blur view is obviously causing the issue, will fix and merge.

PR #565, opened 2019-09-09
Rename favorite bottom sheets

vadymmarkov:
# Why?

Because there are a lot of files prefixed with `favorite` and it became hard navigating through them.

# What?

- Use `FavoriteAd` prefix instead of `Favorite` for types related to favorite ads
- Remove extra view controllers in `FavoriteAdActionSheet` and `FavoriteAdSortingSheet`

# Show me

No UI changes.


vadymmarkov:
Yes, I'll fix that 😉

PR #566, opened 2019-09-10
Resolved potential double-loading of profile images + race

pimms:
# Why?

When integrating the new `IdentityView` into `ios-app` I noticed that we were under some circumstances assigning the `viewModel`-property of `IdentityView` redundantly. This triggered a call to `loadProfileImage()`, which resulted in more than one concurrent request for the same image. While I'll resolve the root cause in `ios-app`, I still think `IdentityView` is at fault for being capable of performing multiple requests.

# Show me

No visual changes,

### DIFF
--- a/Sources/Components/Profile/IdentityView.swift
+++ b/Sources/Components/Profile/IdentityView.swift
@@ -27,6 +27,8 @@ public protocol IdentityViewDelegate: AnyObject {
 
 public class IdentityView: UIView {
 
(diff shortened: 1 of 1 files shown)


bstien:
Since this is now in FinniversKit, would it be possible to use `RemoteImageView` for the profile image?

pimms:
Maybe, I haven't considered it at all tbh. If you have time, please look into it 👍 Not sure if we'd gain anything from it at all though.

PR #567, opened 2019-09-10
Proper font styling & alignment in IdentityView

pimms:
# Why?

Uncoupled the profile-name text color from whether or not the view is tappable. The `isTappable` property has been removed, and the view is always considered tappable — it's up to the delegate to decide whether or not to act upon the event, rather than the View Model's responsibility to inform the view whether or not we should handle tap-events in the first place.

Introduced a new property, `displayMode: IdentityView.DisplayMode`. This enum has three properties, mapping to the three ways we present profiles in the app:

- Interactible
   Used on the object page _(for non-anonymous profiles)_. The profile name is blue, indicating that it can be interacted with.
- Non-interactible
   Used on minFINN. The profile name is black, indicating that the view cannot be interacted with. It still can, but there you go.
- Anonymous
   The profile name is black and aligns vertically with the profile image. The subtitle is always hidden for this mode.

# Show me

### Interactible
Screen Shot 2019-09-10 at 11 59 12

### Non-interactible
Screen Shot 2019-09-10 at 11 59 18

### Anonymous
Screen Shot 2019-09-10 at 11 59 25


Sources/Components/Profile/IdentityView.swift
```diff
@@ -75,13 +86,42 @@ public class IdentityView: UIView {
         return imageView
     }()
 
+    private lazy var profileNameWrapperView: UIView = {
+        let view = UIView(withAutoLayout: true)
```
ag12: It would be nice to call this let for `wrapperView` or something else than `view`.
Just my subjective taste on that matter.

Sources/Components/Profile/IdentityView.swift
```diff
@@ -75,13 +86,42 @@ public class IdentityView: UIView {
...
+        NSLayoutConstraint.activate([
+            profileNameLabel.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            profileNameLabel.topAnchor.constraint(equalTo: view.topAnchor),
+            profileNameLabel.bottomAnchor.constraint(equalTo: view.bottomAnchor),
+
+            verifiedBadge.leadingAnchor.constraint(equalTo: profileNameLabel.trailingAnchor, constant: .smallSpacing),
+            verifiedBadge.centerYAnchor.constraint(equalTo: profileNameLabel.centerYAnchor),
+            verifiedBadge.trailingAnchor.constraint(lessThanOrEqualTo: view.trailingAnchor),
+            verifiedBadge.widthAnchor.constraint(equalToConstant: 18),
+            verifiedBadge.heightAnchor.constraint(equalToConstant: 18)
```
ag12: Again, not sure, but would be nice to have this in a method called `setupProfileViewWrapper` or something. 
Just to make it more convenient. It's up to you.

PR #570, opened 2019-09-10
W36: Add panel component

osanoj:
# Why?
There is a `Panel` component in the design specs. I need to use it in the FINN app.

# What?
Added a Panel component according to design specs. Only need one of the styles now, but figured it was best to include them all. 

One detail to note here is that I made the corner radius configurable, this is because in my use case it is presented right next to a button that does not yet use the normal corner radius.

# Show me
Simulator Screen Shot - iPhone Xʀ - 2019-09-10 at 15 24 57


Sources/Components/Panel/Panel.swift
```diff
@@ -0,0 +1,56 @@
...
+        if let borderColor = style.borderColor {
+            layer.borderColor = borderColor.cgColor
+            layer.borderWidth = 2.0 / UIScreen.main.scale
+        } else {
+            layer.borderColor = nil
+            layer.borderWidth = 0
+        }
+
+        addSubview(textLabel)
+        textLabel.fillInSuperview(margin: .mediumSpacing)
```
3lvis: Looks a bit cramped on the sides, I would probably give it `.mediumLargeSpacing` on the horizontal margins but that's just me 👼

osanoj: Good catch, I think that is what it says in design specs too 👍

Sources/Components/Panel/Panel.swift
```diff
@@ -0,0 +1,56 @@
...
+        label.textAlignment = .center
+        label.textColor = style.textColor
+        return label
+    }()
+
+    private let style: Panel.Style
+
+    public init(style: Panel.Style) {
+        self.style = style
+        super.init(frame: CGRect(x: 0, y: 0, width: 320, height: 20))
```
bstien: Do we need to set the frame for this one?

Demo/Components/Panel/PanelDemoView.swift
```diff
@@ -0,0 +1,46 @@
...
+        let stackView = UIStackView(withAutoLayout: true)
+        stackView.alignment = .center
+        stackView.axis = .vertical
+        stackView.distribution = .fill
+        stackView.spacing = .mediumLargeSpacing
+        addSubview(stackView)
+        stackView.fillInSuperview(margin: .mediumSpacing)
+
+        let demoCases: [(Panel.Style, PanelViewModel)] = [
+            (.plain, PanelViewModel(text: "Plain panel. Are you sure a normal Label isn't a beter fit?")),
```
bstien: Minor typo in `better`😄

Demo/Components/Panel/PanelDemoView.swift
```diff
@@ -0,0 +1,46 @@
...
+
+        demoCases.forEach {
+            let panel = Panel(style: $0.0)
+            panel.translatesAutoresizingMaskIntoConstraints = false
+            panel.configure(with: $0.1)
+            stackView.addArrangedSubview(panel)
+        }
+        let fillerView = UIView(withAutoLayout: true)
+        fillerView.setContentHuggingPriority(UILayoutPriority(0), for: .vertical)
+        stackView.addArrangedSubview(fillerView)
```
bstien: ✨`UIStackView`✨

Demo/Components/Panel/PanelDemoView.swift
3lvis: I'm amazed how this even compiled 😂

osanoj: I beter change it 😉

Sources/Components/Panel/Panel.swift
osanoj: Probably not 👍

PR #586, opened 2019-09-16
Prepare AdsGridView for AdSense

shredlocker:
# Why?

We want to show google ads on the frontpage in the recommendation grid.

# What?

- `AdsGridView` asks data source for cell classes in order to register them.
- In stead of asking for image ratio and text height separately, it nows asks for total height for each item.
- The `AdsGridView`s data source must now dequeue cells in order to configure different cells.
- Adds `BannerAdDemoCell` to demo google adsense.

I'm open for suggestions if there is a better way to do this.

# Show me

Simulator Screen Shot - iPhone Xs - 2019-09-16 at 14 28 15


3lvis:
I think it would be great if you included a "BannerAd" cell in the demo to show the usage of this.

PR #589, opened 2019-09-16
New favourites: add swipe actions for favourite ads

vadymmarkov:
# Why?

Because it should be possible to add comments and delete ads using swipe actions on table view cells.

# What?

- Add "comment" and "delete" swipe actions using iOS11 `UISwipeActionsConfiguration` API
- Originally the title for "comment" action should be "Skriv notat", but it breaks alignment, which ofc is not possible to fix via `UIContextualAction` APIs. I'll talk to designers how we can address this issue. For now I'm using "Notere" as a title.

# Show me

Simulator Screen Shot - iPhone Xs - 2019-09-16 at 17 04 49

### DIFF
--- a/Sources/Fullscreen/FavoriteAdsList/FavoriteAdsListView.swift
+++ b/Sources/Fullscreen/FavoriteAdsList/FavoriteAdsListView.swift
@@ -7,6 +7,8 @@ import UIKit
 public protocol FavoriteAdsListViewDelegate: AnyObject {
     func favoriteAdsListView(_ view: FavoriteAdsListView, didSelectItemAt indexPath: IndexPath)
     func favoriteAdsListView(_ view: FavoriteAdsListView, didSelectMoreButtonForItemAt indexPath: IndexPath)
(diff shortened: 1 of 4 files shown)


Sources/Fullscreen/FavoriteAdsList/FavoriteAdsListView.swift
```diff
@@ -56,6 +58,9 @@ public class FavoriteAdsListView: UIView {
         didSet { tableHeaderView.sortingTitle = sortingTitle }
     }
 
+    public var commentActionTitle = ""
```
bstien: Noticed this one now, but we should probably ask the delegate for the titles. Seeing as we need different ones when the ad has a comment, vs. when it doesn't.

vadymmarkov:
| With comment | Without comment |
| --- | --- |
| Simulator Screen Shot - iPhone Xs - 2019-09-17 at 10 44 34 | Simulator Screen Shot - iPhone Xs - 2019-09-17 at 10 44 41 |

vadymmarkov:
@bstien Updated with a view model and a newline in the string.

PR #590, opened 2019-09-17
Dark mode and semantic colors

osanoj:
# Why?
We need to have semantic colors to support dark mode.

# What?
- Adds the semantic colors that have been decided so far. There might be color tweaks or additions later. 
- Updates the "moon button" to work when running iOS 13, it will override system setting. Use long press to remove override.
- When running on pre iOS 13 you will get an option to kill the app. Starting fresh is the best way to make sure everything is switched to the new mode. If we will actually support dark mode for device earlier than iOS 13 is uncertain, but this makes it easier to test.
- Added a new static config to the FinniverKit class so users of it can decide to support dark mode or not. Defaults to `forceLight`.
- Dark mode support for `Button`, `Dialogue and TextView`.
- Dark mode support for most things in the demo target.
- Builds and works on Xcode 10 and 11. 🎉 

## Notes for later work
- Should consider switching to use colors from asset catalog instead. I didn't do that now because it was quicker for me to define the colors in code AND the first Xcode 11 GM crashed on iOS 11 when trying to load colors from asset catalog.
- Should find a way to get snapshots of dark mode too.
- Go through `//DARK` code comments and verify what color to use

# Show me
## iOS 13
dark-mode-live-switch

## Pre iOS 13
dark-mode-live-switch-ios11

## Dark Button
Simulator Screen Shot - iPhone 8 Plus - 2019-09-17 at 12 29 17

## Dark Dialogue
Simulator Screen Shot - iPhone X - 2019-09-17 at 12 37 42

## Dark TextView
The placeholder text here is very hard to see, will need to run this past Thuy.
Simulator Screen Shot - iPhone X - 2019-09-17 at 12 37 58


fespinoza:
> Should consider switching to use colors from asset catalog instead. I didn't do that now because it was quicker for me to define the colors in code AND the first Xcode 11 GM crashed on iOS 11 when trying to load colors from asset catalog.

Xcode 11 GM2 seems to have fixed this.

> Should find a way to get snapshots of dark mode too.

And iPad and iPhoneSE for some screens too, I've been thinking on replacing the current screenshot library for this one https://github.com/pointfreeco/swift-snapshot-testing

osanoj:
> > Should consider switching to use colors from asset catalog instead. I didn't do that now because it was quicker for me to define the colors in code AND the first Xcode 11 GM crashed on iOS 11 when trying to load colors from asset catalog.
> 
> Xcode 11 GM2 seems to have fixed this.

Yes, I think GM2 should work, but I haven't tried it. It should be simple to switch to asset catalog whenever we want as long as we use the same names for the colors.

> 
> > Should find a way to get snapshots of dark mode too.
> 
> And iPad and iPhoneSE for some screens too, I've been thinking on replacing the current screenshot library for this one https://github.com/pointfreeco/swift-snapshot-testing

Yes it seems like a better choice, but not sure if it helps us with dark mode snapshots at the moment. It might be we need to wait until we use Xcode 11 and can run test on iOS 13. Or add another unit test target that runs the same test but forcing dark mode first.

fespinoza:
I agree with both your replies, in any case i think both are future fixes anyway

PR #591, opened 2019-09-17
Present ToastView from the top

fespinoza:
# Why?

In order to use `ToastView` to replace `TSMessages` in the finn app

# What?

- add method `presentFromTop` to `ToastView`
- refactor inner constraint for `ToastView`, so then we constraint the labels to the safe area (via layout margins) instead of the direct top/bottom of the view. This means I also removed the extra fix subview for safe area.
- modernize `ToastDemoView` constraints using `layoutMarginsGuide`

# Show me

Sep-17-2019 14-48-20


Demo/Components/Toast/ToastDemoView.swift
```diff
@@ -5,6 +5,14 @@
 import FinniversKit
 
 public class ToastDemoView: UIView {
+    private lazy var topToastButton: Button = {
+        let button = Button(style: .callToAction)
+        button.translatesAutoresizingMaskIntoConstraints = false
```
bstien: Doesn't `Button` have a new constructor now with `withAutoLayout` as an argument?

I.e. `Button(style: .callToAction, withAutoLayout: true)`

PR #598, opened 2019-09-19
Bottom sheet: don't animate dim view alpha with spring animator

vadymmarkov:
# Why?

Because it caused blinking when changing the height of the bottom sheet.

# What?

Don't animate dim view alpha with spring animator. It doesn't seem to affect anything, because when going from compact to expanded state and back it used to animate alpha from 1.0 to 1.3, while the max alpha value is 1.0.

# Show me

| Before | After |
| --- | --- |
| bs_before | bs_after |


vadymmarkov:
Oops, I'll fix the description 😉

PR #601, opened 2019-09-19
Use a BankID icon in the VerificationView

robinsalehjan:
# Why?

- The `BankID` icon was missing and needed in the `VerificationView`

# What?

- Add the `BankID` vector
- Update the VerificationView to use the `BankID` icon
- Update snapshot images

# Show me

| Before | After |
| --- | --- |
| Screenshot 2019-09-19 at 14 04 47 | Screenshot 2019-09-19 at 14 03 47 |


Sources/Components/VerificationView/VerificationView.swift
```diff
@@ -8,7 +8,7 @@ public protocol VerificationViewDelegate: AnyObject {
 
 public class VerificationView: UIView {
     private lazy var verificationImageView: UIImageView = {
-        let image = UIImage(named: .spidLogo)
+        let image = UIImage(named: .bankID)
         let imageView = UIImageView(image: image)
         imageView.clipsToBounds = true
         imageView.translatesAutoresizingMaskIntoConstraints = false
```
bstien: Should the imageView have `contentMode` set to `.scaleAspectFit`? I guess the image has the correct dimensions, but just in case? Then you could remove the line with `clipsToBounds` as well.

robinsalehjan: Good catch🥇

PR #602, opened 2019-09-19
New favorites: Edit mode for list of ads

bstien:
# Why?
The list of ads should support edit mode, so the user can remove ads from a folder in bulk.

# Notes
## `tableHeaderView`
The headerView will be "pushed up" when edit mode is activated, and slide in when deactivated. But only if the current offset within the tableView is less than the height of that view.

We chose this solution to prevent unnecessary, and confusing, scrolling if the user has scrolled far down on the list. 

## Resizing cells
This is just something I found during this task, which I wasn't aware of, so I'm noting it down in case it's useful for someone else 🙂

When entering/exiting edit mode, I'm calling `tableView.performBatchUpdates(nil)` from an overridden `UITableView.setEditing(_:)`. This forces the tableView to recalculate the height of the cells without reloading it. Without this, the cells would have the same height as before, aka. without the 24pt margin on the bottom.

## Jumping section headers
To prevent the section headers from jumping all over the place when calling `performBatchUpdates(nil)`, I had to set an `estimatedRowHeight` on the tableView so it could estimate the total height of it's content. It's currently set to 130, which seems to be +/- what the average cell is on an iPhone XS.

# What?
- Enable edit mode for `FavoriteAdsListView`.
- Animate `tableHeaderView` when entering/exiting edit mode.
- Notify delegate when cells are selected/deselected.

# Show me
favorite-ads-edit-mode


Sources/Fullscreen/FavoriteAdsList/FavoriteAdsListView.swift
```diff
@@ -156,8 +197,14 @@ extension FavoriteAdsListView: UITableViewDelegate {
 
     public func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
         tableHeaderView.endEditing(true)
-        tableView.deselectRow(at: indexPath, animated: true)
-        delegate?.favoriteAdsListView(self, didSelectItemAt: indexPath)
+        if !tableView.isEditing {
+            tableView.deselectRow(at: indexPath, animated: true)
+        }
+        selectRow(at: indexPath)
```
vadymmarkov: I think we can just call `delegate?.favoriteAdsListView(self, didSelectItemAt: indexPath)` since it's the only thing we have in `selectRow` method.

Sources/Fullscreen/FavoriteAdsList/FavoriteAdsListView.swift
```diff
@@ -298,3 +345,13 @@ extension FavoriteAdsListView: UISearchBarDelegate {
...
     }
 }
+
+// MARK: - TableView
+
+private class TableView: UITableView {
+    /// Overridden so cells are resized after entering/exiting edit mode.
+    override func setEditing(_ editing: Bool, animated: Bool) {
+        super.setEditing(editing, animated: animated)
+        performBatchUpdates(nil)
```
vadymmarkov: 😎

Sources/Fullscreen/FavoriteAdsList/FavoriteAdsListView.swift
bstien: Yeah, no point in over-engineering this😂 I'll fix it

PR #609, opened 2019-09-23
Favorites/expose remoteimagedatasource

bstien:
# Why?
There are cases in the app where we don't have the ad's image downloaded/cached before we want to present the comment sheet.

# What?
- Pass/use `RemoteImageViewDataSource` to deliver images.


Sources/Fullscreen/FavoriteAdCommentSheet/FavoriteAdCommentViewController.swift
```diff
@@ -125,6 +124,7 @@ final class FavoriteAdCommentViewController: UIViewController {
     override func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
         _ = textView.resignFirstResponder()
+        adView.resetContent()
```
vadymmarkov: Why do we need this?

bstien: Just to make sure `remoteImageView.cancelLoading()` is called from `FavoriteAdView`

PR #611, opened 2019-09-25
[BREAKING] Remove custom message templates

pimms:
# Why?

They were used a whooping 45 times in two weeks, so this is clearly not a feature that scratches and itches... 🙂

# What?

Removed all custom template-related features.

# Note!

This is a breaking change, so I will coordinate the merge & release of FinniversKit and `ios-app` when I've deintegrated the feature from the app.

# Show me

image


ag12:
👌

PR #614, opened 2019-09-25
[Experiment] dark mode screenshots

fespinoza:
# Why?

To have a notion on how our screens currently look in dark mode

# What?

- Remove FBSnapshotTestCase carthage dependency
- Add [SnapshotTesting](https://github.com/fespinoza/swift-snapshot-testing) dependency (it's my fork for the dark mode support, i will make a PR for the original library)
- Recreate all reference screenshots as the library conventions were different so it was easier to do it like this
- Add screenshots for dark mode (Xcode 11 build)
- Add test method `assertSnapshots(matching: UIViewController)...` that takes both light/dark mode screenshots and accepts iPad too optionally

# Show me

| Light Mode | Dark Mode |
|:---:|:---:|
| light | dark |

# Considerations

First there is the change of libraries: I have a branch I based this one that only swaps the FBSnapshotTestCase for the origin SnapshotTesting, to be able to review that change independently.
The main reasons to swap the libraries:
- SnapshotTesting is written in **Swift**, FBSnapshotTestCase is written in **ObjC**
- **SnapshotTesting allows to snapshot with different strategies**: .image (screenshot), .dump as the output of the dump(view) call and allows to create custom representations if needed.
- **SnapshotTesting allows to make snapshot of multiple devices**: .image(.iPhoneX) and .image(.iPhoneSE). Technically, the screenshots are been taken on the device you run your test target, but the library manipulates both UITraitCollections and container frame sizes to simulate different devices, which is not perfect but quite close in my experience.

**This changes requires that we move to Xcode 11** in order to properly take the screenshots in dark mode, this may be blocked until the Finn app is moved to Xcode 11 as well (and possibly other libraries than depend on FinniversKit).

_Speculation_: after that first change, the next consideration is the `SnapshotTesting` support for dark mode, which right now I added as an own fork of the library. But, given Xcode 11 and macOS Catalina. Xcode previews will be a thing, so the library may change the way of how to control the dark mode setting and the whole environment, so my patch on the library would be invalid.


fespinoza:
closed in favor of #662

PR #615, opened 2019-09-25
Replace snapshot testing library

fespinoza:
# Why?

The idea is to prepare for dark mode snapshots, as in #614

Then reasons to favor `SnapshotTesting ` over `FBSnapshotTestCase`:
- SnapshotTesting is written in **Swift**, FBSnapshotTestCase is written in **ObjC**
- **SnapshotTesting allows to snapshot with different strategies**: .image (screenshot), .dump as the output of the dump(view) call and allows to create custom representations if needed.
- **SnapshotTesting allows to make snapshot of multiple devices**: .image(.iPhoneX) and .image(.iPhoneSE). Technically, the screenshots are been taken on the device you run your test target, but the library manipulates both UITraitCollections and container frame sizes to simulate different devices, which is not perfect but quite close in my experience.

# What?

- Remove FBSnapshotTestCase carthage dependency
- Add SnapshotTesting to replace it
- Recreate all reference screenshots as the library conventions were different so it was easier to do it like this

# Show me

_Nothing to show_


fespinoza:
There is a test failing on circle ci, i will check why it fails.

In the meantime, what do you guys think of this change?

3lvis:
I have the same feeling than Vadym, not sure how much snapshot testing helps

osanoj:
I think it will be more useful when we get more sizes and dark mode support, a little reminder if you forgot to test in dark mode or iPad.

Snapshots will also be super useful if we change some of the core things like font sizes or colors.

fespinoza:
There still some test failures on circle CI (not locally, how fun). I asked for some support for circle CI in order to be able to copy the snapshot difference ouput file, as I couldn't figure out how to do it with the job running with SSH

fespinoza:
This is done and working correctly :D 🎉

fespinoza:
@finn-no/ios Remember to run `carthage update` to run the FinniversKit tests :point_up:

vadymmarkov:
@fespinoza Why do we need to run `carthage update` if "Carthage" folder is included in the repo?

fespinoza:
True, you don’t need, I thought you needed :P

PR #618, opened 2019-09-26
Tweak UserAdsHeader colours to match with favourites

robinsalehjan:
# Why?

- To match the styling in favourites.

# What?

- Use `.marble`
- Add gradient overlay at the bottom of the `EmphasizedCell`

# Show me

| Before | After |
| --- | --- |
| before | after |


Sources/Recycling/ListViews/UserAds/Cell/UserAdsListEmphasizedActionCell.swift
```diff
@@ -132,9 +146,13 @@ public class UserAdsListEmphasizedActionCell: UITableViewCell {
...
         actionWrapper.addSubview(actionDescriptionLabel)
-        actionWrapper.addSubview(button)
+        actionWrapper.addSubview(actionButton)
+
+        actionWrapper.addSubview(gradientWrapper)
+        gradientWrapper.layer.addSublayer(gradientLayer)
+
         actionWrapperHideConstraint = NSLayoutConstraint(item: actionWrapper, attribute: .height, relatedBy: .equal, toItem: nil, attribute: .height, multiplier: 1, constant: 0)
-        actionWrapperHeightConstraint = NSLayoutConstraint(item: actionWrapper, attribute: .bottomMargin, relatedBy: .equal, toItem: button, attribute: .bottomMargin, multiplier: 1, constant: .largeSpacing)
+        actionWrapperHeightConstraint = NSLayoutConstraint(item: actionWrapper, attribute: .bottomMargin, relatedBy: .equal, toItem: actionButton, attribute: .bottomMargin, multiplier: 1, constant: .largeSpacing)
```
pimms: Is there any difference between creating constraints like this versus for instance `actionWrapper.bottomAnchor.constraint(equalTo: actionButton.bottomAnchor, constant: .largeSpacing)`?

robinsalehjan: No, not that I know of other than it being an older API.

PR #622, opened 2019-09-26
New favorites: Concealable title view

bstien:
# Why?
According to the Flinto prototype for the redesign the titleView should appear/disappear when the user scrolls through the list.

These changes makes the titleView appear just as the titleLabel within `FavoriteAdsList` starts to scroll out of the view port.

~Currently, the concealable `TitleView` is only added to the Demo. I'm not sure if we should add it to FinniversKit, or if we should implement it in the app. It's only supposed to be used when we present the list of ads. Any inputs here?~

**Edit 1:**
The titleView has been added to FinniversKit as `PercentageDrivenTitleView`. I've added a demo as well.

**Edit 2:**
I've done some work in `DemoViewController` and changed when we add the dismiss button to the views, which now affects the snapshot tests. The snapshot tests have until now had the dismiss button visible, but after talking with @osanoj we've decided to keep them removed from the snapshot tests. This means this PR updates many unrelated snapshot images😅

# What?
- Add delegate method that gets called when the user scrolls, passing the titleLabel's visible percentage.
- Refactor `FavoriteAdsList` into a viewController, instead of a view.
  - Place it within a `UINavigationController`.
- Updated demo to use the concealable titleView.
- Make sure Tweakable can be applied to subclasses of `DemoViewController`.
- Remove the dismissButton from snapshot tests - aka. update a lot of reference images.

# Show me
| FavoriteAdsList | Demo |
| --- | --- |
| favoriteads-concealable-titleview | favoriteads-percentagedriven-titleview |


Sources/Fullscreen/FavoriteAdsList/FavoriteAdsListView.swift
```diff
@@ -306,6 +310,22 @@ extension FavoriteAdsListView: UITableViewDelegate {
 
         return configuration
     }
+
+    public func scrollViewDidScroll(_ scrollView: UIScrollView) {
+        if sendScrollUpdates {
```
vadymmarkov: Can we use `!isEditing` instead?

vadymmarkov:
Since we want to use this `TitleView` in the app I'd keep it in `FinniversKit`. We can rename it to `PercentageDrivenTitleView` or something similar.

Sources/Fullscreen/FavoriteAdsList/FavoriteAdsListView.swift
bstien: I used that initially, but this value needs to be set at different times within the `setEditing(_:)` method for this to work correctly. I decided to change the name of the variable to avoid confusion.

bstien:
> Since we want to use this `TitleView` in the app I'd keep it in `FinniversKit`. We can rename it to `PercentageDrivenTitleView` or something similar.

Sounds like a good idea. I'll move it 💪

PR #629, opened 2019-09-30
Bumped version to 13.1.2

pimms:


pimms:
CircleCI seems broken, force-merging.

PR #634, opened 2019-10-01
New favorites: Background color for section header 

bstien:
# Why?
The background color of the section headers were using the iOS default, when we should've been using `.marble`.

# Show me
Good luck spotting the difference🕵 

| Before | After |
| --- | --- |
| Screenshot 2019-10-01 at 15 34 45 | Screenshot 2019-10-01 at 15 34 05 |


osanoj:
Are you sure you shouldn’t be using .bgTertiary instead? 🖤😜

PR #635, opened 2019-10-02
New favorites: Correct edit mode animation for small content

bstien:
# Why?
I noticed an animation issue that happens when the tableView's content height is smaller than `bounds.height + tableHeaderViewHeight`. In these cases, it's not possible to set the contentOffset, since the tableView will bounce back to show all content. In other words, it's not possible to scroll out of bounds in the tableView.

With these changes I rather animate the tableView's topAnchor constant when the content height is small.

# What?
- Animate `tableView.topAnchor` constant, instead of contentOffset.
- Send titleLabel visibility in completion.

# Show me
| Before | After |
| --- | --- |
| favorites-editmode-small-content-before | favorites-editmode-small-content-after-2 |


bstien:
Wanna hit approve, @vadymmarkov?😄

vadymmarkov:
I thought I did 🙈

PR #641, opened 2019-10-03
New favorites: Draggable emptyView

bstien:
# Why?
When presenting the emptyView, we shouldn't disable user interaction/dragging. Was introduced in #637.

# Show me
| Before | After |
| --- | --- |
| N/A | favorites-emptyview-draggable-after |


vadymmarkov:
But should it actually be possible to drag when presenting the emptyView? I think it's better when the view is static.

bstien:
Thuy believes that it should be. I kinda agree because it provides some form of interaction.

vadymmarkov:
That interaction doesn't really give any value, but ok. Also, at least on the gif, it looks like the movement of empty view and header view are out of sync.

bstien:
Let's discuss this with Thuy when he's available

bstien:
It's approved in person by @vadymmarkov🥇

PR #643, opened 2019-10-04
Add adjustsFontForContentSizeCategory

3lvis:
# Why?

It's cumbersome to have to close and reopen components in order to test Dynamic Type.

# What?

This PR adds `adjustsFontForContentSizeCategory` where it's possible. `adjustsFontForContentSizeCategory` listens for changes in the system and updates the required UI, in combination with the Accessibility Inspector this makes it much easier to test Dynamic Type on the app.

# Show me

dada


Sources/Components/Button/Button/Button.swift
```diff
@@ -50,6 +50,10 @@ public class Button: UIButton {
         super.setTitleColor(style.textColor, for: .normal)
         super.setTitleColor(style.highlightedTextColor, for: .highlighted)
         super.setTitleColor(style.disabledTextColor, for: .disabled)
+
+        if FinniversKit.isDynamicTypeEnabled {
```
fespinoza: why this check? i would imagine the font size wouldn't change if dynamic type is not enabled

3lvis: I was thinking about it. Read more on the docs, the check shouldn't be needed.

Sources/Components/TextField/TextField.swift
```diff
@@ -131,7 +131,7 @@ public class TextField: UIView {
         textField.enablesReturnKeyAutomatically = true
         textField.addTarget(self, action: #selector(textFieldDidChange), for: .editingChanged)
 
-        if FinniversKit.isDynamicTypeEnabled {
+        if !TestCheck.isTesting {
```
fespinoza: i saw circle ci failing, i guess it's because of this... but what does change when running the snapshot tests regarding the content size category? :O

3lvis:
Had to update the screenshots because of this but it's all good now :)

PR #644, opened 2019-10-04
Add support for showing a label below the views in the HappinessRating component

robinsalehjan:
# Why?

- A design related to a market problem that is in the making
- I agreed with Thuy that this should be a part of the component and optional

# What?

- Make it possible to render a `TextLabel` below each `ImageView` in the `HappinessRating` as requested by the value returned from the delegate method `func happinessRatingView(_ happinessRatingView: HappinessRatingView, textFor: HappinessRating)`

- The implementation of this method is optional

# Show me

| Before | After |
| --- | --- |
| before | after |


Sources/Components/HappinessRating/HappinessRating.swift
```diff
@@ -25,4 +25,12 @@ public enum HappinessRating: Int, CaseIterable {
             return UIImage(named: .ratingFaceLove)
         }
     }
+
+    func textLabel(text: String) -> Label {
```
bstien: I think this method should rather be put in the view it's being used in, aka. the `HappinessRatingView`

robinsalehjan: It was a property but I refactored and forgot to move it 😅  I assume you mean the `RatingImageView` class? if so then I agree.

The `HappinessRatingView` shouldn't need to know if the particular `ImageView` that is being added has a `TextLabel`.

Sources/Components/HappinessRating/HappinessRatingView.swift
```diff
@@ -6,13 +6,21 @@ import UIKit
 
 public protocol HappinessRatingViewDelegate: AnyObject {
     func happinessRatingView(_ happinessRatingView: HappinessRatingView, didSelectRating rating: HappinessRating)
+
+    // MARK: - Optional
+    func happinessRatingView(_ happinessRatingView: HappinessRatingView, textFor: HappinessRating) -> String?
```
bstien: This method doesn't seem to be called from anywhere🤔

Sources/Components/HappinessRating/HappinessRatingView.swift
```diff
@@ -106,24 +116,67 @@ public class HappinessRatingView: UIView {
     }
 }
 
+extension HappinessRatingView: RatingImageViewDelegate {
+    func ratingImageViewText(for rating: HappinessRating) -> String? {
+        switch rating {
+        case .angry:
+            return "Veldig irriterende"
```
bstien: Guess the delegate method mentioned above should be called from here, so we can get rid of these strings

Sources/Components/HappinessRating/HappinessRatingView.swift
```diff
@@ -163,23 +163,31 @@ private class RatingImageView: UIImageView {
...
+                label.topAnchor.constraint(equalTo: bottomAnchor, constant: .smallSpacing),
+                label.leadingAnchor.constraint(equalTo: leadingAnchor),
+                label.trailingAnchor.constraint(equalTo: trailingAnchor, constant: paddingToFitTwoLongWords)
             ])
         }
     }
 
-    required init?(coder aDecoder: NSCoder) {
-        fatalError("init(coder:) has not been implemented")
+    func textLabel(text: String) -> Label {
```
bstien: This could be private 🕵

Sources/Components/HappinessRating/HappinessRatingView.swift
robinsalehjan: Yep, was supposed to have these in the demo and call the delegate from here.

Sources/Components/HappinessRating/HappinessRatingView.swift
robinsalehjan: Optional implementation so we don't force it to be implemented everywhere.

Sources/Components/HappinessRating/HappinessRatingView.swift
robinsalehjan: The class is `private` I assume that the methods should also be `private` unless explicitly set to `public`?

bstien: The class will be private within the scope of the file, but the methods will be internal, aka. accessible within the file scope.

I forgot that the class was private, so then it's only nitpicking 🙃

Sources/Components/HappinessRating/HappinessRatingView.swift
robinsalehjan: Never mind... herp derp

PR #645, opened 2019-10-04
New favorites: Empty folder view

bstien:
# Why?
We want to present a separate view if the list of ads is empty.

For a user's own folder, or a shared folder without ads, this view should be presented below the tableHeader.

For a shared folder that has been unshared, or for an ID that doesn't exist, this view should be presented over the entire screen. This should probably be done in the app.

# What?
- Added view `FavoriteAdsListEmptyView`.
- Renamed `FavoriteEmptyView` to `FavoriteSearchEmptyView`.
- Added properties to `FavoriteAdsListViewModel`.
- Updated demo with tweak.

# Show me
| Before | After |
| --- | --- |
| Screenshot 2019-10-04 at 15 57 20 | Screenshot 2019-10-04 at 15 29 07 |


bstien:
@vadymmarkov Feel free to make any necessary changes and merge :)

PR #646, opened 2019-10-04
Multi-device support for Demo

3lvis:
# Why?

It's quite often that we either forget to test in smaller and bigger devices or cases where we optimize UI for bigger screens and smaller screens. Since iOS 8 Apple has been pushing us to work towards size classes, with the launch of iOS 13 and Mac Catalyst Apple has made their message quite clear, we need to stop thinking about iPhone and iPad and start thinking on resizable screens. 

Xcode Previews will make this easier but while we wait for it I thought it would be a good idea to add support to our current demo to switch between different screen sizes while we are coding our components.

# What?

This PR introduces an extra option to the already existing "Tweak" button. When you tap the "Tweak" button, it will now present all the different screen sizes available for your device. Each screen size will also change its corresponding size class.

In order to make this work, we now wrap our DemoViewController into a DemoViewControllerContainer which changes the frame and traits using `setOverrideTraitCollection`.

# Important

In order for this feature to work now you have to use `DemoViewControllerContainer` instead of `DemoViewController`.

```swift
        case .reputationView:
            // Before
            return DemoViewController<ReputationDemoView>()

            // After this PR
            return DemoViewControllerContainer<ReputationDemoView>()
```

# Show me

multidevice


3lvis:
This PR is ready to be reviewed.

3lvis:
@finn-no/ios 🙉

Demo/Demo/DemoViewControllerContainer.swift
```diff
@@ -0,0 +1,103 @@
...
+            viewController.view.autoresizingMask = device.autoresizingMask
+            setOverrideTraitCollection(device.traits, forChild: viewController)
+        }
+
+        NSLayoutConstraint.activate([
+            viewController.view.topAnchor.constraint(equalTo: view.topAnchor),
+            viewController.view.bottomAnchor.constraint(equalTo: view.bottomAnchor),
+            viewController.view.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            viewController.view.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            viewController.view.centerYAnchor.constraint(equalTo: view.centerYAnchor),
```
vadymmarkov: I guess there is no need to add center X/Y constraints since there are top, bottom, leading and trailing.

Demo/Tweakable/Device.swift
```diff
@@ -0,0 +1,136 @@
+import UIKit
+
+struct Device {
+    enum `Type`: String {
```
vadymmarkov: Should we perhaps use `Kind` to avoid conflicts with Swift metatype.

Demo/Tweakable/Device.swift
```diff
@@ -0,0 +1,136 @@
...
+        case phone4inch = "iPhone 5S (4-inch)"
+
+        // swiftlint:disable:next identifier_name
+        case phone4_7inch = "iPhone 8 (4.7-inch)"
+
+        // swiftlint:disable:next identifier_name
+        case phone5_5inch = "iPhone 8 Plus (5.5-inch)"
+
+        // swiftlint:disable:next identifier_name
+        case phone5_8inch = "iPhone X (5.8-inch)"
```
vadymmarkov: Do we also need 6.1-inch iPhone11?

Demo/Demo/DemoViewController.swift
```diff
@@ -148,22 +140,3 @@ public class DemoViewController<View: UIView>: UIViewController, Containable {
         }
     }
 }
-
-extension DemoViewController: CornerAnchoringViewDelegate {
```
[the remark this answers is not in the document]
3lvis: I did the name change, I considered making BaseDemoViewController into a private class but there are a few places where we are pushing controllers instead of using plain views, in those places we need to subclass BaseDemoViewController. 🤔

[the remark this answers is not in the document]
vadymmarkov: Right, haven't thought about that.

Demo/Tweakable/Device.swift
3lvis: I thought about it but the aspect ratio it's not very different from the iPhone X.

PR #647, opened 2019-10-04
Rename EasterEggButton to FloatingButton + Badge support

3lvis:
# Why?

The usefulness of this button goes beyond the Easter Eggs, so it makes sense to rename it to something different.

# What?

I needed this button to say something extra about what was inside and thought using a badge was a good way of doing it. If the count is higher than 0 it shows the badge, otherwise it remains hidden.

# Show me

| No badge | With badge |
| --- | --- |
| Simulator Screen Shot - iPhone X - 2019-10-04 at 17 14 36 | Simulator Screen Shot - iPhone X - 2019-10-04 at 17 14 30 |


Demo/Components/Button/FloatingButton/FloatingButtonDemoView.swift
```diff
@@ -0,0 +1,48 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import FinniversKit
+
+public class FloatingButtonDemoView: UIView, Tweakable {
```
osanoj: Since it is made for the tweaking options, how about getting Tweaking in the name somehow? Like `TweakingButtonDemoView` or `TweakingFloatingButtonDemoView` if we want to be really specific about what it is and does 🤔

3lvis: It's used in the FINN app as EasterEggButton so it has already multiple uses

osanoj: But it is under `Demo/Components/...` and has `Demo` in the name so this class should not be used outside the FinniversKit demo... are you sure it is this class?

3lvis: `FloatingButtonDemoView` is the Demo view of the component `FloatingButton`.

osanoj: Ah, now I get it, it is the demo of the `FloatingButton`, not a "demo component" to use for showing tweaking options.... 🤦‍♂ 😂

osanoj: I should have probably looked at the entire changeset before commenting 😂

3lvis: Haahah, I was thinking that was the confusion

PR #657, opened 2019-10-14
Fullscreen presentation, instead of pageSheet

bstien:
# Why?
The default `modalPresentationStyle` in iOS 13 is `.pageSheet`. This doesn't work very well with our demos in FinniversKit, as they're meant to be presented in full screen.

Other demo views, such as the bottomSheets have already set this value to something else and won't be affected by these changes.

# What?
- Set `modalPresentationStyle` to `.fullScreen`, if the current value is `.pageSheet`.
- Small refactoring in the Demo's `AppDelegate`.

# Show me
| Before | After |
| --- | --- |
| demo-presentation-fullscreen-before | demo-presentation-fullscreen-after |

### DIFF
--- a/Demo/AppDelegate.swift
+++ b/Demo/AppDelegate.swift
@@ -9,10 +9,7 @@ import FinniversKit
 class AppDelegate: UIResponder, UIApplicationDelegate {
     var window: UIWindow?
 
-    lazy var navigationController: NavigationController = {
-        let navigationController = NavigationController(rootViewController: DemoViewsTableViewController())
-        return navigationController
-    }()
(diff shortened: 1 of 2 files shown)


bstien:
My initial thought was to check if the viewController we get passed is of type `DemoViewController`, but this failed as the class' signature is `DemoViewController<View: UIView>`.

Ex. for `ContactFormView`:
```swift
viewController is DemoViewController // <- false
viewController is DemoViewController<ContactFormView> // <- true
```

Not sure how to test for this, since we apparently need the concrete type. I therefor chose to change the `modalPresentationStyle` if it's set to `.pageSheet` :)

"None of our demo views should have this value"™

PR #658, opened 2019-10-14
Use semantic colors for backgrounds

fespinoza:
# Why?

As an initial step to support dark mode in all FinniversKit

# What?

- Replace `.marble` usages for `.bgTertiary`
- Replace `.milk` for `.bgPrimary`
- Replace `.ice` for `.bgSecondary`

This according to the color palette present in trello.

There will be more refinements, per-view, but for now we do a common conversion to dark mode

# Show me

| Before | After |
| --- | --- |
| loan-calculator-1 | loan-calculator-2 |
| favorites1 | favorite-list-2 |
| feedback1 | feedback-2 |
| grid1 | grid2 |
| loginentry | loginentry-2 |


pimms:
Shoutout to the login-prompt card, it looks ridiculously good with the light drop-shadow 👌

PR #659, opened 2019-10-15
New favorites: FavoriteAdsListEmptyView configuration

bstien:
# Why?
We need to configure the emptyListView independently for shared and a user's folder.

# What?
- Remove emptyListView properties from viewModel.
- Add method to configure the emptyListView.

# Show me
No UI.


bstien:
These changes are not necessary after all. Closing.

PR #661, opened 2019-10-16
Replace carthage for SwiftPM for swift-snapshot-testing + Xcode 11 updates

fespinoza:
# Why?

- SwiftPM is more flexible than carthage (ex: to use development version of packages) and has native support in Xcode 11

# What?

- Remove carthage
- Reinstall swift-snapshot-test with SPM
- Update circleCI settings for Xcode 11
- Update reference snapshots for iOS 13 + iPhone 11 Pro: ⚠️ if you see the image changes in GithHub, you can notice some updates regarding spacing, fonts coming from iOS 13, the same as the new appearance for the `UISegmentedControl`, which I will fix in a new PR after.

# Show me

| Before | After |
| --- | --- |
| before | after
 |


fespinoza:
I talked to thuy about the segmented control, details about the iOS 13 changes can be found in https://medium.com/flawless-app-stories/ios-13-uisegmentedcontrol-3-important-changes-d3a94fdd6763

Since thuy is busy right now I will do a quick fix for this and then we will improve it. My potential desings for this can be like:

Screenshot 2019-10-16 at 11 58 04
Screenshot 2019-10-16 at 12 01 18

I'm inclined for now for the second option, then the style of a segmented control looks like a UISwitch

fespinoza:
I fixed the UISegmentedControl to a design i considered convenient, we can tweak it after with mr Thuy

PR #664, opened 2019-10-16
Component refinements for dark mode

fespinoza:
# Why?

because of dark mode 🕶 

# What?

- I went through all the components in dark mode and fixed the ones that needed it
  - toast
  - review button
  - panel
  - AdsGridViewCell
  - NativeAdvertView

With this both **DNA** and **Components** are fully dark mode compatible. Next I will move to: Cells, Recyclable and Fullscreen

# Show me

| Screen | Before | After |
| :---: | --- | --- |
| AdsGridViewCell | img | img |
| NativeAdvertView | img |  img |
| Panel | img | img |
| review button | img | img |
| toast | img | img |


fespinoza:
ready to review :)

PR #665, opened 2019-10-18
New favorites: Xmas folder

vadymmarkov:
# Why?

Because we want to let people create and share Christmas wishlists with a help of FINN favorites.

# What?

- Add a new style for Xmas favorite folder
- Add `FavoriteFolderXmasView` with callout text explaining the feature.

We might want to add some cool animations, but that's left for the next PR.

# Show me

xmas


Sources/Recycling/ListViews/FavoriteFolders/Subviews/FavoriteFolderXmasView.swift
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    private lazy var button: FloatingButton = {
+        let button = FloatingButton.favoritesXmasButton()
+        button.isUserInteractionEnabled = false
+        return button
+    }()
+
+    private lazy var calloutView: CalloutView = {
+        let view = CalloutView(direction: .down, arrowAlignment: .right(.veryLargeSpacing / 2 - .mediumSpacing))
```
bstien: I think it's more readable if we just write `24` here instead, seeing as we have literals in the constraints below. What do you think?

vadymmarkov: `.veryLargeSpacing` is the width of the button here, but I agree that `24` is better in this case 👍

PR #666, opened 2019-10-22
Christmas wish list popup

fespinoza:
# Why?

HackDay. To encourage people to give used stuff for christmas

# What?

- Add `ChristmasWishListView` component
- Add `MissionPopupView`: a new generic popup-like container view
- Add christmas wish list banner asset
- Add `fillInSuperviewLayoutMargins` extension to `UIView`

# Show me

Screenshot 2019-10-21 at 12 38 28


fespinoza:
awww... #666 should had been for something dark mode related xD

Sources/Extensions/UIKit/Layout.swift
```diff
@@ -37,4 +37,22 @@ public extension UIView {
         let insets = UIEdgeInsets(top: margin, leading: margin, bottom: -margin, trailing: -margin)
         return fillInSuperview(insets: insets, isActive: isActive)
     }
+
+    @discardableResult
+    func fillInSuperviewLayoutMargins() -> [NSLayoutConstraint] {
```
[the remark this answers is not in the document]
bstien: I guess you could pass the view's `layoutMarginsGuide` into that method

Sources/Components/ChristmasWishListView/ChristmasWishListView.swift
```diff
@@ -0,0 +1,165 @@
+//
+//  Copyright © 2019 FINN AS. All rights reserved.
+//
+
+import Foundation
+
+public protocol ChristmasWishListViewDelegate: AnyObject {
+    func christmasWishListViewDidSelectReadMore(_ view: ChristmasWishListView)
```
vadymmarkov: It's probably not useful in this case, but I think we should start adding sender view in these kinds of delegate methods, mainly for presenting popovers on iPad.

Sources/Components/ChristmasWishListView/ChristmasWishListView.swift
```diff
@@ -0,0 +1,165 @@
...
+
+        NSLayoutConstraint.activate([
+            bannerImageView.heightAnchor.constraint(equalToConstant: 225),
+        ])
+    }
+
+    @objc private func firstPageButtonTap() {
+        UIView.animate(withDuration: 0.3) {
+            self.firstPageButton.transform = CGAffineTransform(translationX: -self.bounds.width, y: 0)
+            self.secondPageButton.transform = CGAffineTransform.identity
```
vadymmarkov: Can be just `.identity`.

Sources/Components/ChristmasWishListView/Views/MissionPopupView.swift
```diff
@@ -0,0 +1,138 @@
+//
+//  Copyright © 2019 FINN AS. All rights reserved.
+//
+
+public protocol MissionPopupViewDelegate: AnyObject {
+    func missionPopupViewDidSelectClose(_ view: MissionPopupView)
+}
+
+/// An attempt to create a generic pop-up like container to create views like the klimabrølet pop up
+public class MissionPopupView: UIView {
```
vadymmarkov: Why did you choose this name?

Sources/Components/ChristmasWishListView/Views/MissionPopupView.swift
```diff
@@ -0,0 +1,138 @@
...
+    // MARK: - Public properties
+    weak var delegate: MissionPopupViewDelegate?
```
vadymmarkov: Should probably be public.

Sources/Components/ChristmasWishListView/Views/MissionPopupView.swift
fespinoza: Because Thuy knows the component like this

fespinoza:
> Any plans on refactoring Klimabrølet and Earth Hour popups with this new container view btw?

No, klimabrølet is not gonna be used (for at least a year) so I don't think it's necessary until we need it again.

The earth hour pop up may not fit with the mission pop up component, because of the custom animations

Sources/Components/ChristmasWishListView/ChristmasWishListView.swift
fespinoza: I don't think we need a pop over in this case, so I would prefer to avoid adding more verbosity to the delegate methods until it's required

vadymmarkov: Agree, wrote this as a reminder because it would be another breaking change in the public API if you actually needed pop overs, but forgot to send the source view.

PR #667, opened 2019-10-22
Cells: Make detailLabel constraint open

bstien:
# Why?
iOS 13 seems to have changed some of the default spacing/padding, including those for `UITableViewCell`'s with chevron.

I noticed this when doing some work in Charcoal, where the `detailLabel` was off in iOS 13. The chevron doesn't have the padding it used to in previous versions.

Screenshot 2019-10-22 at 12 48 05

# What?
- Make `detailLabelTrailingConstraint` open, instead of private.
- Use different constraint constant if iOS 13.

# Show me

| iOS version| Before | After |
| --- | --- | --- |
| iOS 13 | Screenshot 2019-10-22 at 12 50 54 | Screenshot 2019-10-22 at 12 50 31 |
| iOS 11.4 | Screenshot 2019-10-22 at 12 53 31 | 👈 Same as before |

### Charcoal
**NB!** These screenshots are taken after setting the constraint manually in Charcoal's subclasses of `BasicTableViewCell`.

| iOS version| Before | After |
| --- | --- | --- |
| iOS 13 | Screenshot 2019-10-22 at 12 48 05 | Screenshot 2019-10-22 at 13 19 23 |
| iOS 11.4 | Screenshot 2019-10-22 at 13 07 10 | Screenshot 2019-10-22 at 13 22 42 |


fespinoza:
👏

fespinoza:
you probably need to update some snapshots

PR #669, opened 2019-10-22
Review and adapt "Recycling" components to dark mode

fespinoza:
# Why?

🕶 

# What?

- Fix small visual details for dark mode in the "Recycling" section

That section is now fully dark mode enabled

# Show me

| Before | After |
| --- | --- |
| img | img |
| img | img |
| img | img |


fespinoza:
Now it's ok to review

PR #672, opened 2019-10-23
SelectorTitleView

3lvis:
# Why?

A component used in navigation headers.

# What?

Adds the SelectorTitleView.

# Show me

| Light | Dark |
| --- | --- |
| Simulator Screen Shot - iPhone X - 2019-10-23 at 11 51 25 | Simulator Screen Shot - iPhone X - 2019-10-23 at 11 51 33 |


3lvis:
@bstien used here as well, when switching devices https://user-images.githubusercontent.com/1088217/66216916-a05eef00-e6c6-11e9-8697-2ffca77e5f48.gif

Sources/Components/SelectorTitleView/SelectorTitleView.swift
```diff
@@ -0,0 +1,126 @@
...
+        button.titleLabel?.adjustsFontForContentSizeCategory = true
+
+        let spacing = .smallSpacing / 2
+
+        button.semanticContentAttribute = .forceRightToLeft
+        button.imageEdgeInsets = UIEdgeInsets(top: spacing, leading: spacing, bottom: 0, trailing: -spacing)
+        button.titleEdgeInsets = UIEdgeInsets(top: 0, leading: -spacing, bottom: 0, trailing: spacing)
+        button.addTarget(self, action: #selector(handleButtonTap), for: .touchUpInside)
+
+        let buttonColor: UIColor = .btnPrimary
```
bstien: Could this be a property with a default value instead? For the tableHeaderView in favorites this needs to have a different color

Sources/Components/SelectorTitleView/SelectorTitleView.swift
```diff
@@ -0,0 +1,126 @@
...
+        let interfaceBackgroundColor: UIColor = .bgPrimary
```
bstien: This isn't used anywhere, so I guess it can be removed

Sources/Components/SelectorTitleView/SelectorTitleView.swift
```diff
@@ -0,0 +1,126 @@
...
+        label.textColor = .textSecondary
+        return label
+    }()
+
+    private lazy var button: UIButton = {
+        let button = UIButton(withAutoLayout: true)
+        button.titleLabel?.font = UIFont.bodyStrong.withSize(17).scaledFont(forTextStyle: .footnote)
+        button.titleLabel?.adjustsFontForContentSizeCategory = true
+
+        let spacing = .smallSpacing / 2
```
bstien: This could be `.verySmallSpacing`

Sources/Components/SelectorTitleView/SelectorTitleView.swift
```diff
@@ -0,0 +1,126 @@
...
+        setup()
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    // MARK: - Setup
+
+    public var title: String? {
```
bstien: This should probably be moved to the top with the rest of the properties

Sources/Components/SelectorTitleView/SelectorTitleView.swift
```diff
@@ -0,0 +1,126 @@
...
+        didSet {
+            button.setTitle(title, for: .normal)
+        }
+    }
+
+    private func setup() {
+        translatesAutoresizingMaskIntoConstraints = false
+        arrowDirection = .down
+        isEnabled = true
```
bstien: These two are already given the default values at the top of the file

3lvis: Yes but if I don't set them it won't trigger the "didSet" call, and if I remove the default value then it will complain.

Any suggestions?

Sources/Components/SelectorTitleView/SelectorTitleView.swift
3lvis: Made it into a method so it's possible to set the disabled color as well.

Sources/Components/SelectorTitleView/SelectorTitleView.swift
3lvis: 🔪

Sources/Components/SelectorTitleView/SelectorTitleView.swift
bstien: The button will be enabled by default, so we can skip that one, but one solution for the arrow:

```swift
public var arrowDirection: ArrowDirection = .down {
    didSet { updateArrowDirection() }
}

// […]

private func setup() {
    updateArrowDirection()
    // Other code
}

// […]

private func updateArrowDirection() {
    let asset: FinniversImageAsset = arrowDirection == .up ? .arrowUpSmall : .arrowDownSmall
    button.setImage(UIImage(named: asset), for: .normal)
}
```

3lvis: I guess that's kind of the same but different. 

Works for me tho!

3lvis:
_The changes have been made!_

PR #675, opened 2019-10-23
Experiment: Add Xcode Previews to the `ChristmasWishListDemoView`

fespinoza:
# Why?

As a test on how to use Xcode previews in catalina 😎 

# What?

- Add a wrapper for the ChristmasWishListDemoView for swiftUI
- Use swift UI to create a preview for ChristmasWishListDemoView in different settings

# Show me

Screenshot 2019-10-23 at 22 49 27
Screenshot 2019-10-23 at 22 50 46

Gif

http://www.giphy.com/gifs/YpTb00TZmzz2sGJf1o

# Notes

This should be safe to run even on Mojave, since we use Xcode 11 all the dependencies are there but the previous won't show

Since I opted for creating a preview over a demo view, it's easy to make a UIViewRepresentable for any demo view and make the previews with that UIViewRepresentable

### DIFF
--- a/Demo/Components/ChristmasWishList/ChristmasWishListDemoView.swift
+++ b/Demo/Components/ChristmasWishList/ChristmasWishListDemoView.swift
@@ -76,6 +76,7 @@ struct ChristmasWishListDemoWrapperView: UIViewRepresentable {
 }
 
 @available(iOS 13.0, *)
+// swiftlint:disable:next type_name
 struct ChristmasWishListDemoView_Previews: PreviewProvider {
     static var previews: some View {
         Group {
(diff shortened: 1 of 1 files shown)


fespinoza:
More info:

- https://nshipster.com/swiftui-previews/

- https://www.avanderlee.com/xcode/xcode-previews/

fespinoza:
Can confirm it doens't conflict with Xcode 11 and Mojave

fespinoza:
I will merge this and leave it as a small example of xcode previews until we can upgrade to catalina and take this more seriously

Demo/Components/ChristmasWishList/ChristmasWishListDemoView.swift
```diff
@@ -62,3 +62,47 @@ extension ChristmasWishListViewModel {
         )
     }
 }
+
+#if canImport(SwiftUI) && DEBUG
+import SwiftUI
+
+struct ChristmasWishListDemoWrapperView: UIViewRepresentable {
```
vadymmarkov: When we actually start using previews, this struct could be made into more generic `DemoWrapperView` I think.

PR #677, opened 2019-10-24
Add priming component

vadymmarkov:
# Why?

In order to make it easier for users to acquire the necessary knowledge about new features we're adding to the app.

# What?

- Add `PrimingView`
- Show demo in the bottom sheet
- Introduce new `DynamicShadowView` for cases when we need scroll view -driven top/bottom+top shadow 

# Show me

| iPhone11 | iPhone11 + scrolling |
| --- | --- |
| iphone11 | shadow |

| iPhone5s | Dark Mode |
| --- | --- |
| iPhone5s | Simulator Screen Shot - iPhone 11 Pro - 2019-10-24 at 14 21 18 |


Sources/Components/Priming/PrimingView.swift
```diff
@@ -0,0 +1,128 @@
...
+
+    // MARK: - Private properties
+
+    private var rows = [PrimingViewModel.Row]()
+
+    private lazy var headerView = PrimingHeaderView(withAutoLayout: true)
+
+    private lazy var tableView: UITableView = {
+        let tableView = UITableView(withAutoLayout: true)
+        tableView.translatesAutoresizingMaskIntoConstraints = false
```
fespinoza: not needed

Sources/Util/DynamicShadowView.swift
```diff
@@ -0,0 +1,47 @@
...
+        )
+
+        let intersection = contentFrame.intersection(frame)
+        layer.shadowRadius = min(intersection.height * 0.2, DynamicShadowView.maxShadowRadius)
+    }
+
+    // MARK: - Setup
+
+    private func setup() {
+        backgroundColor = .white
```
fespinoza: is this intentional?

and btw, why is this class used?

vadymmarkov: Nope, will change it to `bgPrimary` 👍 This class is used when we need to update shadow appearance based on the content offset of the scroll view. It's a header view in most cases, but this time I also needed it as a footer view, so I had to modify existing `BottomShadowView` to have the shadow not only on bottom, but on top as well.

PR #680, opened 2019-10-28
New favorites: show/hide xmas button with animation

vadymmarkov:
# Why?

To draw user's attention to the xmas button and callout.

# What?

- Remove `FavoriteFolderXmasView` since we decided to drop the dimmed background
- Add methods to show/hide xmas button in `FavoriteFoldersListView`
- Add haptic feedback when showing the callout

# Show me

xmass_btn


3lvis:
😍 😍 😍 😍

PR #682, opened 2019-10-29
New favorites: add footer view with share button

vadymmarkov:
# Why?

We want to show extra share button for Christmas wishlists.

# What?

- Rename `PrimingFooterView` into `FooterButtonView` and make it into component since it's being used in multiple places

- Rename `DynamicShadowView` to `TopShadowView` because that's what it actually is

- Change `FavoriteAdsListDemoView` to be `UIView` instead of `UIViewController` in order to make it work with `Tweakable` protocol

- Make it possible to show a footer view with share button in `FavoriteAdsListView`

# Show me

sharebutton


bstien:
Seems like the build is failing because of the modified `ReviewView` reference image. Not sure how that was updated with these changes, though🤔

vadymmarkov:
Yesterday I had the same problem locally. This particular test was failing even though there were no visual changes when I tried to compare two reference images. I had to record a new snapshot in order to fix it, but it's weird because I don't see how `ReviewView` could be affected by these changes 🤔

bstien:
Screenshot 2019-10-30 at 08 25 44

Seems weird. No clue where that green "shadow" behind the profile image is suppose to come from 🤨🤔

vadymmarkov:
I cleaned the project and it helped fixing `testReviewView` with the old reference image, rolling backache change...

PR #683, opened 2019-10-30
Add settings tableview

shredlocker:
This PR is towards my WIP branch

# Why?

We have a new design for the settings screen.

# What?

- I've added three types of cells: `SettingsViewCell`, `SettingsViewConsentCell` and `SettingsViewToggleCell`. Each cell have their own view model.
- There are separate section header and footer views with different fonts and constraints. Header view has a static height, the height of the footer view is set depending on the view.
- Updates `SettingsView` with the changes above
- Updates `SettingsViewDemoView` with the new models and delegate methods.
- New reference image for snapshot testing

# Show me

| Before | After |
| --- | --- |
| Simulator Screen Shot - iPhone 11 Pro - 2019-10-30 at 11 11 03 | Simulator Screen Shot - iPhone 11 Pro - 2019-10-30 at 11 03 14 |


Sources/Recycling/ListViews/Settings/Cell/SettingsViewToggleCell.swift
```diff
@@ -0,0 +1,75 @@
...
+
+    weak var delegate: SettingsViewToggleCellDelegate?
+
+    // MARK: - Private Properties
+
+    private var model: SettingsViewToggleCellModel?
+
+    private lazy var uiswitch: UISwitch = {
+        let toggle = UISwitch(withAutoLayout: true)
+        toggle.onTintColor = .primaryBlue
```
fespinoza: `btnPrimary`

PR #691, opened 2019-11-01
Add settings details view

shredlocker:
# Why?

There are new designs for displaying details about consents, legitimate interests and other action in the settings screen.

# What?

- Replaced `ConsentToggleView` and `ConsentActionView` with a single `SettingsDetailsView`
- `SettingsDetailsView` has to possible states based on whether the view model has a `secondaryButtonTitle`. The `secondaryButton` toggles between the two states: `normal` and `details`. (Feedback on these names are appreciated)
- Hacked a demo view controller to make the bottom sheet act as in the app.

# Show me

| Action View | Toggle View |
| --- | --- |
| Simulator Screen Shot - iPhone 11 Pro - 2019-11-01 at 12 57 41 | Simulator Screen Shot - iPhone 11 Pro - 2019-11-01 at 12 57 46 |
| New Design | |
settings-details |


Sources/Fullscreen/SettingsDetailsView/SettingsDetailsView.swift
```diff
@@ -0,0 +1,222 @@
...
+    )
+
+    private lazy var scrollView: UIScrollView = {
+        let scrollView = UIScrollView(withAutoLayout: true)
+        scrollView.delegate = self
+
+        scrollView.contentInset = UIEdgeInsets(
+            top: 0,
+            left: 0,
+            bottom: scrollViewBottomInset,
```
fespinoza: i think you can just define `UIEdgeInsets(bottom: scrollViewBottomInset)`

Sources/Fullscreen/SettingsDetailsView/SettingsDetailsView.swift
```diff
@@ -0,0 +1,222 @@
...
+        scrollView.addSubview(secondaryButton)
+        addSubview(scrollView)
+        addSubview(shadowView)
+        addSubview(primaryButton)
+
+        NSLayoutConstraint.activate([
+            scrollView.leadingAnchor.constraint(equalTo: leadingAnchor),
+            scrollView.topAnchor.constraint(equalTo: topAnchor),
+            scrollView.trailingAnchor.constraint(equalTo: trailingAnchor),
+            scrollView.bottomAnchor.constraint(equalTo: bottomAnchor),
```
fespinoza: i guess you can use `fillInParent`

Sources/Fullscreen/SettingsDetailsView/SettingsDetailsView.swift
shredlocker: I'll fix

Sources/Fullscreen/SettingsDetailsView/SettingsDetailsView.swift
shredlocker: You're right!

PR #700, opened 2019-11-06
Clear indication of currently selected image in gallery

ninarg:
# Why?

We get feedback from users that do not like new carousel effect in the image gallery on object pages. People say they keep swiping without noticing that it is the same images over and over again. It is also strange that it is possible to swipe when there's only one image.

# What?

- Added border to currently selected image
- Disabled swiping in image galleries with only one image
- Displaying the caption (1/1) also when there is only one image

# Show me

| Before | After |
| --- | --- |
| gallery-before | gallery-after |

| Before (one image) | After (one image) |
| --- | --- |
| one-image-before | one-image-after |


fespinoza:
700 🎉

Sources/Fullscreen/FullscreenGallery/GalleryPreviewCell.swift
```diff
@@ -56,11 +58,24 @@ class GalleryPreviewCell: UICollectionViewCell {
...
+        imageView.layer.borderWidth = 0.0
     }
 
     // MARK: - Public methods
 
     func configure(withImage image: UIImage?) {
         imageView.image = image
     }
+
+    func border(isVisible: Bool) {
```
vadymmarkov: Functions usually describe an action, so I'd probably call this `showBorder(_ show: Bool)` or something similar.

Sources/Fullscreen/FullscreenGallery/GalleryPreviewView.swift
```diff
@@ -96,6 +97,24 @@ class GalleryPreviewView: UIView {
         collectionView.scrollToItem(at: IndexPath(row: index, section: 0), at: .centeredHorizontally, animated: animated)
     }
 
+    public func addBorderToItem(atIndex index: Int) {
```
vadymmarkov: This function could be simplified to:

```swift
public func addBorderToItem(atIndex index: Int) {
    func galleryCell(at row: Int) -> GalleryPreviewCell? {
        collectionView.cellForItem(at: IndexPath(row: row, section: 0)) as? GalleryPreviewCell
    }
    
    if let previouslySelectedRow = selectedRow {
        galleryCell(at: previouslySelectedRow)?.border(isVisible: false)
    }
    selectedRow = index
    galleryCell(at: index)?.border(isVisible: true)
}
```

What do you think?

PR #702, opened 2019-11-13
💫Sandbox and Bootstrap 💥

3lvis:
## Background

Been playing with adding UI Frameworks to small projects for quicker iterations and it works great! Unfortunately, I was copy-pasting a lot of the code for what makes the FinniversKit demo work, also copying a few things that felt like a no brainer like the cell registering and dequeuing, the spacing, the font handling, the layout helpers and others. Because of this, I thought it would be a good idea to make this process easier while maintaining a single source of truth for building great UI Frameworks.

Sandbox is the demo container library, you initialize a SandboxViewController with a bunch of view controllers and it builds a demo for you, Sandbox provides you nice helpers such as the ability to resize your screen for easy testing and allows you to support Tweakables (which have a lot of potential by themselves).

Bootstrap, as I mentioned, is the common elements between your Demo app and your UI Framework, it's very simple at the moment but we can discuss later what else to place in there.

## The new project structure

Screenshot 2019-11-17 at 15 01 05

## How to use?

In your demo app initialize a SandboxViewController with an array of SandboxSections and set it as your root controller.

```swift
import Sandbox

func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
        // You might want to override the default fonts and colors provided by SparkCommon
        // If you don't override them you'll get the default ones
        registerFonts()
        registerColors()

        let color = SandboxItem(title: "Color", viewController: colorViewController) 
        let dna = SandboxSection(title: "DNA", items: [color])
        let sandboxViewController = SandboxViewController(sections: [dna])
        let navigationController = NavigationController(rootViewController: demoViews)

        window = UIWindow(frame: UIScreen.main.bounds)
        window?.rootViewController = navigationController
        window?.makeKeyAndVisible()
        return true
}
```

## Registering Fonts & Colors

I wanted Sandbox to be an agnostic library that could use any styling, for this Bootstrap has Palette and FontBook, these are singletons that will store your fonts and colors and use those instead of the default ones. They both work very similarly.

```swift
enum FontType: String {
    case medium = "FINNTypeWebStrippet-Medium"
}

// This method will override the title1 with the FINN font
func registerFonts() {
    FontBook.registerFonts(in: Bundle(for: AppDelegate.self), fontNames: [
        FontType.medium.rawValue
        ])

    FontBook.shared.title1 = UIFont(name: FontType.medium.rawValue, size: 34)!
}

func registerColor() {
    // Makes all primary backgrounds red, why would you do that? No idea.
    Pallete.shared.bgPrimary = .red
}
```


3lvis:
I think my MVP is ready for now, any feedback is welcome.

3lvis:
Been considering the name *Sandbox* for the UI Framework, sandbox relates to something playful also to a safe place where you can test new things, having your UI Framework build using the "Sandbox" sounds pretty fun. Also is related to cats :D.

Not sure about the common library, tho.

3lvis:
Got a nice name for the Common library: Bootstrap. Made the necessary renaming and updated the README.

vadymmarkov:
@3lvis Do we care about unique names for our libraries on https://cocoapods.org? Because both Sandbox and Bootstrap names are of course taken. But I'm not sure if we have any plans on pushing them to public CocoaPods repo.

3lvis:
I think we could add the "FINN" prefix for CocoaPods, so it will be "FINNSandbox" and "FINNBootstrap" over there. The most important is that the names are simple and clear.

vadymmarkov:
Yeah, `FINN` prefix was my initial thought as well 👍

3lvis:
Splitting this PR in 2.

First: https://github.com/finn-no/FinniversKit/pull/707

3lvis:
Closing this in favor of https://github.com/finn-no/FinniversKit/pull/707 and a future Sandbox PR.

PR #705, opened 2019-11-18
AdOut: link to effect documentation

fespinoza:
# Why?

- [Trello task](https://trello.com/c/vpOTGzPb/71-link-to-effectdocumentation-in-my-ad)
We want to show a website (as a first version) that contains the full statistics for a given ad.

# What?

- Extend `UserAdManagementStatisticsCell` to include the title and "full stats" button
- Introduce `StatisticsModel` which contains header data and the `StatisticsItemModel` array.
- Set background color to `.bgPrimary` for UserAdManagement cells

## Explanation

The design we are trying to achieve is something like this

design

For me, the title and button should be part of the statistic cell.

The data must come from the backend as the rest of the data populating this screen. For that reason we would have to extend the statistic endpoint to include this information.

## Alternative approach

Since I'm not fully familiar with the approaches of data/views used in AdIn. I also explored an alternative approach where the title and button are part of their own cell.

I was playing around to add the title/link as another cell instead, that can be found in https://github.com/finn-no/FinniversKit/tree/adout/link-to-effect-doc-new-cell-approach

## TODO

Confirm this approach is correct from the AdIn point of view and coordinate with the backend people to extend the required endpoint.

# Show me

| Before | After |
| --- | --- |
| before | after |


Sources/Recycling/ListViews/AdManagement/Cells/UserAdManagementButtonAndInformationCell.swift
```diff
@@ -101,25 +107,35 @@ public class UserAdManagementButtonAndInformationCell: UITableViewCell {
...
+
+        NSLayoutConstraint.activate([
+            separatorView.heightAnchor.constraint(equalToConstant: hairLineSize),
+            separatorView.topAnchor.constraint(equalTo: contentView.topAnchor),
+            separatorView.leadingAnchor.constraint(equalTo: contentView.leadingAnchor),
+            separatorView.trailingAnchor.constraint(equalTo: contentView.trailingAnchor),
+
+            informationLabel.leadingAnchor.constraint(equalTo: contentView.leadingAnchor, constant: labelLeftInset),
+            informationLabel.trailingAnchor.constraint(equalTo: button.leadingAnchor, constant: -labelToButtonSpacing),
+            informationLabel.topAnchor.constraint(equalTo: contentView.topAnchor, constant: .mediumSpacing),
```
bstien: Should this rather be constrained to the `separatorView` so we get the full spacing?

Sources/Recycling/ListViews/AdManagement/Cells/UserAdManagementStatisticsCell.swift
```diff
@@ -58,9 +87,19 @@ public class UserAdManagementStatisticsCell: UITableViewCell {
         }
     }
 
+    @objc func fullStatisticsButtonTapped() {
```
bstien: This could be marked as `private`

Sources/Recycling/ListViews/AdManagement/Cells/UserAdManagementButtonAndInformationCell.swift
fespinoza: nice catch!

fespinoza:
Screenshot 2019-11-29 at 14 01 25

Updated with the external website icon

fespinoza:
ready for a final review

PR #707, opened 2019-11-20
🌬🥾 Bootstrap 💥

3lvis:
## Background

Been playing with adding UI Frameworks to small projects for quicker iterations and it works great! Unfortunately, I was copy-pasting a lot of the code for what makes the FinniversKit demo work, also copying a few things that felt like a no brainer like the cell registering and dequeuing, the spacing, the font handling, the layout helpers and others. Because of this, I thought it would be a good idea to make this process easier while maintaining a single source of truth for building great UI Frameworks.

The first step to achieve this is to add a "Common" library. I have called this library Bootstrap, the main goal for this library is to have a set of components, helpers, extensions that we use in our different apps or medium-sized libraries, if the library is small we probably don't need to bring Boostrap there until we need it.

## The new project structure

Screenshot 2019-11-20 at 12 31 23

## How to use?

Import the Boostrap library and customize its styling using FontBook and Pallete. If you don't do this it will default to generic font and colors.

## Customizing Fonts & Colors

Bootstrap has Palette and FontBook, these are singletons that will store your fonts and colors and use those instead of the default ones. They both work very similarly.

```swift
enum FontType: String {
    case medium = "FINNTypeWebStrippet-Medium"
}

// This method will override the title1 with the FINN font
func registerFonts() {
    FontBook.registerFonts(in: Bundle(for: AppDelegate.self), fontNames: [
        FontType.medium.rawValue
        ])

    FontBook.shared.title1 = UIFont(name: FontType.medium.rawValue, size: 34)!
}

func registerColors() {
    // Makes all primary backgrounds red, why would you do that? No idea.
    Pallete.shared.bgPrimary = .red
}
```

Then in your AppDelegate:

```swift
@UIApplicationMain
class AppDelegate: UIResponder, UIApplicationDelegate {
    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
        registerFonts()
        registerColors()
        // ...
    }
}
```


3lvis:
Ready for reviewing.

Bootstrap/SelectorTitleView/SelectorTitleView.swift
```diff
@@ -132,8 +135,16 @@ public class SelectorTitleView: UIView {
     // MARK: - Private
 
     private func updateArrowDirection() {
-        let asset: FinniversImageAsset = arrowDirection == .up ? .arrowUpSmall : .arrowDownSmall
-        button.setImage(UIImage(named: asset), for: .normal)
+        if #available(iOS 13.0, *) {
```
vadymmarkov: Should we have a fallback for old iOS versions?

Demo/AppDelegate.swift
```diff
@@ -29,6 +32,46 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
 
+extension AppDelegate {
+    enum FontType: String {
+        case light = "FINNTypeWebStrippet-Light"
+        case medium = "FINNTypeWebStrippet-Medium"
+        case regular = "FINNTypeWebStrippet-Regular"
+        case bold = "FINNTypeWebStrippet-Bold"
+    }
+
+    func registerFonts() {
```
vadymmarkov: Does it mean we'll have to do the same kind of font registration in the app? Perhaps we could make FINN fonts and colors as default in `Bootstrap`? Or add some sort of helper `registerFINNFonts` function in `FontBook` extension in `FinniversKit` if we don't want to move our custom fonts to `Bootstrap`.

Sources/DNA/Color/Color.swift
```diff
@@ -189,29 +86,16 @@ import UIKit
         return UIColor(hex: "#262633")
     }
 
-    // swiftlint:disable:next identifier_name
-    convenience init?(r: CGFloat, g: CGFloat, b: CGFloat, a: CGFloat = 1.0) {
-        self.init(red: r / 255.0, green: g / 255.0, blue: b / 255.0, alpha: a)
+    public class var accentSecondaryBlue: UIColor {
```
vadymmarkov: Why do we need these `accent` colors here?

Sources/DNA/Color/Color.swift
```diff
@@ -427,43 +307,8 @@ extension CGColor {
...
-
-@objc extension UIColor {
-    public class var defaultCellSelectedBackgroundColor: UIColor {
-        let lightSelectedColor = UIColor(r: 230, g: 235, b: 242)!
-        return dynamicColorIfAvailable(defaultColor: lightSelectedColor, darkModeColor: lightSelectedColor.withAlphaComponent(0.4))
-    }
-}
-
 extension CGColor {
     public class var defaultCellSelectedBackgroundColor: CGColor {
```
vadymmarkov: Should this var go to `Bootstrap`?

Bootstrap/SelectorTitleView/SelectorTitleView.swift
```diff
@@ -30,6 +30,9 @@ public class SelectorTitleView: UIView {
         }
     }
 
+    public var arrowUpImage: UIImage?
+    public var arrowDownImage: UIImage?
```
osanoj: Probably the button image needs to be updated when these are set too? Or make it possible to set them in init.

Bootstrap/SelectorTitleView/SelectorTitleView.swift
3lvis: My idea was to keep Boostrap injectable, not having any assets help us to use Swift Package Manager. Also, I think it's more flexible if you inject your own arrows but this class probably doesn't have the best API since it should be included in the `init` so it's clear that it needs that in order to be displayed properly.

Demo/AppDelegate.swift
3lvis: > Does it mean we'll have to do the same kind of font registration in the app?

Yes, the idea is that the hosting app injects custom fonts, if you don't do that it will use the default fonts. If we embed the FINN fonts in Bootstrap it's most likely that we'll see a bunch of apps using the FINN font and we definitely don't want that. We'll most likely remove the colors and make the default colors the iOS colors but that can wait into a future PR.

Sources/DNA/Color/Color.swift
3lvis: Good question, I don't know why do we use the accent colors, I just move them here because they have no usage in Bootstrap.

Sources/DNA/Color/Color.swift
3lvis: They aren't used in Boostrap or FinniversKit, I think we need to add the CGColor extension for only colors that we need, not for everything.

Bootstrap/Palette.swift
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    public class var dimmingColor: UIColor {
+        return UIColor.black.withAlphaComponent(0.4) //DARK
+    }
+}
+
+// MARK: - FINN UIColors
+@objc extension UIColor {
+    class var ice: UIColor {
```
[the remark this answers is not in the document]
3lvis: >Maybe the FINN colors should not be in Bootstrap? Should we have them in FinniversKit and if no colors are supplied to Bootstrap it uses some iOS default ones instead?

Yes! I just didn't want to do that in this PR.

>Not sure what is best, but it depends on if this is a FINNBootstrap - primarily for FINN usage or not. Ideally we shouldn't be using these color names directly very often now anyway.

The biggest value of Bootstrap is to make a reusable generic library for building UI. We could use Bootstrap to build other FINN apps or apps owned by FINN, for example, Styletime (or any other apps that FINN ends up buying or building).

Bootstrap/Palette.swift
```diff
@@ -0,0 +1,264 @@
...
+    public var decorationSubtle: UIColor?
+    public var iconPrimary: UIColor?
+    public var iconSecondary: UIColor?
+    public var defaultCellSelectedBackgroundColor: UIColor?
+}
+
+// MARK: - Semantic colors, dark mode compatible
+@objc extension UIColor {
+    public class var bgPrimary: UIColor {
+        return Palette.shared.bgPrimary ?? dynamicColorIfAvailable(defaultColor: .milk, darkModeColor: UIColor(hex: "#1B1B24"))
```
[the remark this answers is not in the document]
3lvis: If we do that we won't be able to say 

`view.backgroundColor = .bgPrimary`

So this needs to stay in a UIColor extension. Palette is only used to store the state.

Bootstrap/Palette.swift
```diff
@@ -0,0 +1,264 @@
...
+            self.init(red: CGFloat(red) / 255.0, green: CGFloat(green) / 255.0, blue: CGFloat(blue) / 255.0, alpha: 1.0)
+        } else {
+            self.init(red: 0.0, green: 0.0, blue: 0.0, alpha: 0.0)
+        }
+    }
+}
+
+// MARK: - Private helper for creating dynamic color
+extension UIColor {
+    public class func dynamicColorIfAvailable(defaultColor: UIColor, darkModeColor: UIColor) -> UIColor {
```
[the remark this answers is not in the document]
3lvis: Would be interesting to discuss this I get a feeling that you have a good suggestion for a cleaner API.

Bootstrap/SelectorTitleView/SelectorTitleView.swift
3lvis: I think init is the way to go, I'll update it.

Bootstrap/SelectorTitleView/SelectorTitleView.swift
vadymmarkov: But we do inject images here, so what I mean is:
```swift
if #available(iOS 13.0, *) {
    ...
} else {
    button.setImage(arrowDirection == .up ? arrowUpImage : arrowDownImage, for: .normal)
}
```

3lvis:
Made some changes to SelectorTitleView to allow image injection instead of using San Francisco symbols.

3lvis:
✅  Ready for review

3lvis:
We discussed that there are a few things we need to agree on before making this PR a reality. @fespinoza has some notes regarding things we might want to consider. One of the biggest points was how customizable we want FinniversCore to be and if it should embed fonts and colors. I think we can use this PR as inspiration for building a solution the whole team agrees on. Since I started this for the Mobile Era and Oslo Pride app, I'll continue with the plans I already had.

Closing for now :)

fespinoza:
One note I would add as well, for the next attempt to do this, is to make a new repo from the start.
Then it's easier to test the compatibility with something like charcoal which should depend on our
core framework

PR #712, opened 2019-11-21
Split address view into the map and card components

vadymmarkov:
# Why?

According to new design the address card view should be presented as a bottom sheet (of course). Therefore it's better to have two separate components for the map and card views and setup them accordingly in the view controller in the app.

# What?

- Make `AddressCardView` public
- Rename `AddressView` to `AddressMapView` and remove address card view from it
- Add new "view mode" button to the map view because we decided to use action sheet instead of segmented control for choosing the map type. This will create more screen space for the map.

# Show me

| Before | After (Map) | After (Card) |
| --- | --- | --- |
| Simulator Screen Shot - iPhone 11 Pro - 2019-11-21 at 10 57 32 | Simulator Screen Shot - iPhone 11 Pro - 2019-11-21 at 10 30 14 | Simulator Screen Shot - iPhone 11 Pro - 2019-11-21 at 10 30 02 |


Sources/Fullscreen/AddressMapView/AddressMapView.swift
```diff
@@ -0,0 +1,207 @@
...
+
+    public func makePolygonOverlayVisible() {
+        guard let firstPolygon = polygons.first else {
+            return
+        }
+
+        let mapRect = polygons.dropFirst().reduce(firstPolygon.boundingMapRect, { (mapRect, polygon) -> MKMapRect in
+            return polygon.boundingMapRect.union(mapRect)
+        })
+        let bottomInset = UIDevice.isIPad() ? 16 : 16 + .mediumLargeSpacing
```
shredlocker: Should we start using `traitCollection.verticalSizeClass.regular` for these types of checks?

Sources/Components/AddressCardView/AddressCardView.swift
```diff
@@ -3,20 +3,32 @@
...
     func addressCardViewDidSelectCopyButton(_ addressCardView: AddressCardView)
     func addressCardViewDidSelectGetDirectionsButton(_ addressCardView: AddressCardView, sender: UIView)
 }
 
-class AddressCardView: UIView {
-    weak var delegate: AddressCardViewDelegate?
+public final class AddressCardView: UIView {
+    public weak var delegate: AddressCardViewDelegate?
+
+    public var model: AddressCardViewModel? {
```
shredlocker: Maybe use the `configure(with model: AddressCardViewModel)` method for this?

Sources/Fullscreen/AddressMapView/AddressMapView.swift
vadymmarkov: Definitely 👍

PR #713, opened 2019-11-21
Bootstrap: Color refactoring suggestion

osanoj:
**Note: This is a PR for discussing changes to Bootstrap**

# Why?
There were some things with Bootstrap color handling that didn't feel quite right. This is a suggestion for change, open for discussion.

# What?
- Moved UserInterfaceStyleSupport back to FinniversKit, this should not be needed in Bootstrap since it gets all color set into Palette.
- Refactored Bootstrap.Palette to have default values for all colors. I did not spend a lot of time choosing the default colors. 😛 
- Made Bootstrap UIColor extension internal, is not so nice to pollute the "namespace" unless everything is at least prefixed (I think FinniversKit can do this since it is for FINN use and we want to always use those color names).
- Moved all the FINN colors to FinniversKit.
- FinniversKit defines a Palette object initialized with FINN colors.
- Added a setup method to FinniversKit class. This should handle all important setup values, for now only UserInterfaceStyleSupport but things like dynamic fonts should also be there. Setup method will then configure Bootstrap to fit FinniversKit.

# Show me
No visible changes.

### DIFF
--- a/Bootstrap/Bootstrap.swift
+++ b/Bootstrap/Bootstrap.swift
@@ -1,13 +1,5 @@
 import Foundation
 
 @objc public class Bootstrap: NSObject {
-    public enum UserInterfaceStyleSupport {
-        @available(iOS 13.0, *)
-        case dynamic
-        case forceLight
-        case forceDark
-    }
-
     public static var isDynamicTypeEnabled: Bool = true
-    public static var userInterfaceStyleSupport: UserInterfaceStyleSupport = .forceLight
 }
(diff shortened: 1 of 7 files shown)


3lvis:
😍 nice work, Jonas! The only thing I would have kept is the UIColor convenience `init`s and the `dynamicColorIfAvailable` into Bootstrap since these are helper methods that you are going to need in pretty much any app.

osanoj:
> 😍 nice work, Jonas! The only thing I would have kept is the UIColor convenience `init`s and the `dynamicColorIfAvailable` into Bootstrap since these are helper methods that you are going to need in pretty much any app. Everything else works great!

Yes, probably makes sense to keep them in Bootstrap.

PR #714, opened 2019-11-22
Height bounds for images in grid views

ninarg:
# Why?

We have no limits for the height of images in grid views. Ads with a tall and narrow image will cover half of the screen in recommendations. For job ads, companies often upload wide and short logos. For these kinds of ads, the icon label and favorite button cover the logo. 

# What?

- Added border to images
- Set upper bound for image ratio to 1.5
- Set lower bound for image ratio to 0.75
- Images are by default set to scaleAspectFill
- Added a new variable canScaleImageToFillView to AdsGridViewModel protocol, so that images can be set to scaleAspectFit for some markets. E.g. for job ads, canScaleImageToFillView is set to false, because we want to display the whole logo

# Show me

(Added a bunch of company logos for the demo, so titles etc. look weird)

| Before | After |
| --- | --- |
| before | after |


3lvis:
Hahaha, great image!

Sources/Recycling/GridViews/Ads/Cell/AdsGridViewCell.swift
```diff
@@ -263,6 +271,13 @@ public class AdsGridViewCell: UICollectionViewCell {
         let colors: [UIColor] = [.toothPaste, .mint, .banana, .salmon]
         let color = colors[index % 4]
         loadingColor = color
+
+        if let model = model {
+            if !model.canScaleImageToFillView {
```
shredlocker: I would prefer to have this test for `true` instead.
`if model.scaleImageToFitView {`

ninarg: Thanks for the feedback! I agree with the two above. However, in this one, the model does not have a scaleImageToFitView variable. Since the default is to have scaleImageToFill, I think it makes sense to keep that one instead of changing the variable to scaleImageToFitView. Right now, ios-app sets scaleImageToFillView to true for all markets except jobs.

PR #718, opened 2019-11-25
Added possibility to limit map zoom range

osanoj:
# Why?
There is a new iOS 13 change in how maps work. You can now specify how much a user is allowed to zoom in and out.

# What?
This lets you set how far the user can zoom when using AddressView and SearchResultMapView.

Added a custom struct `MapZoomRange` so there is no need to `import MapKit` when limiting zoom range.

# Show me
Not visible in FinniversKit, but will be in the main app. PR there will follow...


osanoj:
Merging the AddressMapView changes from @vadymmarkov. Anyone wants to have another quick look?

PR #723, opened 2019-12-05
Refactor:`Button.Style` as a `struct` instead of an `enum`

fespinoza:
# Why?

In several occassions, by design specs, we need button that look like the FINN
buttons, but they have some small modifications. Currently it's not possible to
extend/modify button styles because of their `enum` representation, neither is
possible to create new styles for concret usages outside of the `Style` enum.

# What?

- Re-write `Button.Style` as a `struct` instead of an `enum`
- Add `overrideStyle` method to `Button.Style`

This PR doesn't change the public API for `Button` and `Button.Style`. At the
same time, this will allow to remove unnecessary subclasses of button in order
to create custom buttons for a given use case.

Also, thanks to the snapshot test, we can make sure that the button styles were
not changed with the refactoring 🎉 

## Example

Screenshot 2019-12-05 at 11 15 58

To create the call to action-like button in the screenshot, we can just do:

```swift
extension Button.Style {
    static func applyButton(with accentColor: UIColor) -> Button.Style {
        Button.Style.callToAction.overrideStyle(
            bodyColor: accentColor,
            highlightedBodyColor: accentColor.withAlphaComponent(0.8)
        )
    }
}
```

And then we have a "generic" call to action button with different colors

### DIFF
--- a/Sources/Components/Button/Button/Button+Style.swift
+++ b/Sources/Components/Button/Button/Button+Style.swift
@@ -11,7 +11,7 @@ public extension Button {
     }
 
     /// Convenience grouping of button state related properties
-    struct ButtonStateStyle {
+    struct StateStyle: Equatable {
         let textColor: UIColor?
         let backgroundColor: UIColor?
         let borderColor: UIColor?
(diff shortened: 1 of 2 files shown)


fespinoza:
I was also thinking in having a potential convenience structure called `ButtonStateStyle`. To be able to define styles like:

```

fespinoza:
any more feedback? suggestions?

PR #724, opened 2019-12-09
Improvement: loan calculator component takes bank's color

fespinoza:
Task: https://trello.com/c/GXRL62TD/115-use-the-banks-colors-in-loan-calculator

# Why?

A more aesthetic appearance to the loan calculator component

# What?

- Allow to set `accentColor` to `StepSlider`
- Add requirement of `accentColor: UIColor?` to `LoanCalculatorViewModel`
- generate thumb images for the step slider with that _accent color_
when it's set, otherwise it will use the default thumb images found in
the assets
- Add other bank test cases for the demo view for the loan calculator

This work for both light and dark mode.

Now we are able to create `UIImage` dynamically given an accent color.
We need to re-compute the image on trait collection change, as the
accent color will resolve with a different value in dark mode.

# Show me

| Before | After |
| --- | --- |
| before | after |

| Bank | Light | Dark |
| --- | --- | --- |
| DNB | dnb-light | dnb-dark |
| Nordea | nordea-light | nordea-dark |
| Danske Bank | db-light | db-dark |


fespinoza:
One thing I am not sure how to do is: _how to have a property like accent color as an optional protocol requirement (with default nil value)_, so in that sense, the integration between FinniversKit and the finn app doesn't break.

fespinoza:
**Note 2:** the bank images are not fully solved for dark mode (as you can see in the dark Nordea case)

Demo/Components/LoanCalculator/LoanCalculatorDemoView.swift
```diff
@@ -159,6 +199,24 @@ extension LoanCalculatorDemoViewModel {
     }
 }
 
+private extension UIColor {
+    static func dynamicColor(defaultColor: UIColor, darkModeColor: UIColor) -> UIColor {
```
vadymmarkov: Don't we have `dynamicColorIfAvailable` already?

Sources/Components/StepSlider/StepSlider.swift
```diff
@@ -80,6 +90,19 @@ public final class StepSlider: UISlider {
         updateActiveTrackRange()
     }
 
+    public override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+
+        guard
+            #available(iOS 12.0, *),
```
vadymmarkov: Why do we need this availability check here?

Sources/Components/StepSlider/StepSlider.swift
```diff
@@ -214,3 +252,68 @@ public final class StepSlider: UISlider {
...
+                blur: shadowSize,
+                color: UIColor(white: 0.0, alpha: 0.3).cgColor
+            )
+
+            UIColor.white.setFill()
+            context.cgContext.fillEllipse(in: bounds.applying(.init(translationX: shadowSize, y: shadowSize)))
+            context.cgContext.restoreGState()
+
+            self.accentColor.setFill()
+            let center = bounds
```
vadymmarkov: Shouldn't this always be the center of the rect or I'm missing something?

Demo/Components/LoanCalculator/LoanCalculatorDemoView.swift
fespinoza: it's not public. this is part of the demo, I was unsure if to make the `dynamicColorIfAvailable` public as the internals depend on finniversKit as well the interface style support for iOS 13

Sources/Components/StepSlider/StepSlider.swift
fespinoza: `userInterfaceStyle` is iOS 12+

Sources/Components/StepSlider/StepSlider.swift
fespinoza: maybe i can improve the name.
the intention is _the origin of the rect where the concentric circle will be_
if that origin is the center of the outer rect itself then the circle would not be centered

Demo/Components/LoanCalculator/LoanCalculatorDemoView.swift
vadymmarkov: Right, but I think it's still useful to have this kind of extension in `FinniversKit`. I know it's a bit confusing to have two functions with similar names, but this `dynamicColor(defaultColor:darkModeColor)` could handle only the interface style support and `dynamicColorIfAvailable` would also check `FinniversKit.userInterfaceStyleSupport` stuff in addition to that. What do you think?

fespinoza: I agree, I also saw myself needing other color related stuff, like the `UIColor(hex:)` initializer

fespinoza: I will make a new PR exposing these stuff in FinniversKit

fespinoza:
This should be ready to go 🚀🚀🚀

PR #728, opened 2019-12-12
Add MinFinn views and cells

shredlocker:
# Why?

All views in the Finn App should be in this repo.

# What?

- Added `MinFinnView` together with demo files.
- Added profile, verification, icon and basic cells with models
- Verification cell have gotten a facelift, more accurate towards the web version
- Tried a new thing were I pass the cells delegates with their models, any thoughts?

# Show me

| Normal | Dark |
| --- | --- |
| Simulator Screen Shot - iPhone 11 Pro - 2019-12-11 at 16 13 45 | Simulator Screen Shot - iPhone 11 Pro - 2019-12-11 at 16 13 22 |


Sources/Fullscreen/MinFinn/MinFinnView.swift
```diff
@@ -0,0 +1,132 @@
...
+        switch section {
+        case 0: return .mediumLargeSpacing
+        default: return .largeSpacing
+        }
+    }
+}
+
+// MARK: - Private methods
+private extension MinFinnView {
+    func setup() {
```
vadymmarkov: I think it's not necessary to have this function if it's only one line.

Sources/Cells/BasicTableViewCell/BasicTableViewCell.swift
```diff
@@ -65,22 +65,22 @@ open class BasicTableViewCell: UITableViewCell {
 
     // MARK: - Public methods
 
-    open func configure(with viewModel: BasicTableViewCellViewModel, indexPath: IndexPath? = nil) {
-        titleLabel.text = viewModel.title
+    open func configure(with viewModel: BasicTableViewCellViewModel?, indexPath: IndexPath? = nil) {
```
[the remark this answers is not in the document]
shredlocker: Woops, I just forgot to remove these changes. Although it is nice to have this as an optional because in `prepareForReuse()` you can simply say `configure(with: nil)`. But not in this PR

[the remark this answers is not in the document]
shredlocker: And it would force us to consider displaying an empty cell, which is a good thing

Sources/Cells/BasicTableViewCell/BasicTableViewCell.swift
```diff
@@ -89,7 +89,7 @@ open class BasicTableViewCell: UITableViewCell {
             stackViewToDetailLabelConstraint.isActive = false
         }
 
-        if viewModel.hasChevron {
+        if viewModel.hasChevron == true {
```
vadymmarkov: I think `== true` is not needed here.

shredlocker: Right, leftovers from when it was optional 👍

PR #731, opened 2019-12-16
New empty search list view and save search prompts

ninarg:
# Why?

We haven't had a proper empty search list view, only some old cell written years ago. Thuy and his team also decided to introduce a new button called utility button to FinniversKit. 

# What?

- New utility button
- New empty search list view (displayed when search gives zero results)
- New save search prompt view (displayed when there are results, but you scrolled to the end or reopened a previous search)

# Show me

**Buttons**

| Light mode | Dark mode |
| --- | --- |
| IMG_0836 | IMG_0837 |

**Empty search list view**

| Light mode | Dark mode |
| --- | --- |
| IMG_0840 | IMG_0841 |

**Save search prompt view**

| Light mode | Dark mode |
| --- | --- |
| IMG_0838 | IMG_0839 |


Sources/Components/SaveSearchView/SaveSearchPromptView.swift
```diff
@@ -0,0 +1,100 @@
...
+public class SaveSearchPromptView: UIView {
+
+    // MARK: - Public properties
+
+    public weak var delegate: SaveSearchPromptViewDelegate?
+
+    // MARK: - Private properties
+
+    private lazy var titleLabel: UILabel = {
+        let label = UILabel()
```
shredlocker: You might want to use our `Label` class here. It supports scaling fonts which is something we want to use with all labels 👍

Sources/Components/SaveSearchView/SaveSearchPromptView.swift
```diff
@@ -0,0 +1,100 @@
...
+        label.numberOfLines = 2
```
shredlocker: This labels could use more than two lines if we use large fonts. It is a good idea to test how the view looks when turning on large fonts

Sources/Components/SaveSearchView/SearchListEmptyView.swift
```diff
@@ -0,0 +1,121 @@
...
+
+            button.topAnchor.constraint(equalTo: stackView.bottomAnchor, constant: .mediumLargeSpacing),
+            button.centerXAnchor.constraint(equalTo: centerXAnchor),
+            button.heightAnchor.constraint(equalToConstant: 31)
+        ])
+    }
+
+    // MARK: - Public methods
+
+    public func configure(withViewModel viewModel: SearchListEmptyViewModel, forState state: SearchListEmptyView.SearchListEmptyViewState) {
```
shredlocker: I would write this as `configure(with viewModel: SearchListEmptyViewModel, for state: SearchListEmptyView.State)`. It looks better at the call site 👍

Sources/Components/SaveSearchView/SearchListEmptyViewModel.swift
```diff
@@ -0,0 +1,18 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import Foundation
+
+public class SearchListEmptyViewModel {
```
shredlocker: It might be a good idea to use a `protocol` instead of `class` here.

Sources/Components/SaveSearchView/SaveSearchPromptView.swift
```diff
@@ -0,0 +1,100 @@
...
+        label.textColor = .textPrimary
+        label.textAlignment = .center
+        label.font = .captionStrong
+        label.translatesAutoresizingMaskIntoConstraints = false
+        return label
+    }()
+
+    private lazy var positiveButton: UIButton = {
+        let button = Button(style: .utility, size: .small)
```
bstien: We have this init, so you could remove the `translatesAutoresizingMaskIntoConstraints` line below: `Button(style:size:withAutoLayout:)`

Sources/Components/SaveSearchView/SaveSearchPromptView.swift
```diff
@@ -0,0 +1,100 @@
...
+        button.addTarget(self, action: #selector(positiveButtonTapped), for: .touchUpInside)
+        button.translatesAutoresizingMaskIntoConstraints = false
+        return button
+    }()
+
+    private lazy var dismissButton: UIButton = {
+        let button = UIButton(frame: .zero)
```
bstien: This could use our custom init for `UIView`'s, which also gives it `frame: .zero`

```swift
let button = UIButton(withAutoLayout: true)
```

Sources/Components/SaveSearchView/SearchListEmptyView.swift
```diff
@@ -0,0 +1,121 @@
...
+
+public protocol SearchListEmptyViewDelegate: AnyObject {
+    func searchListEmptyViewDidSelectActionButton(_ searchListEmptyView: SearchListEmptyView, forState state: SearchListEmptyView.SearchListEmptyViewState)
+}
+
+@objc public class SearchListEmptyView: UIView {
+
+    // MARK: - Public properties
+
+    @objc public enum SearchListEmptyViewState: Int {
```
bstien: Does this need to be exposed to objc? It doesn't seem to be used in the PR for the app.

To shorten the name, since it's already a nested enum, how about naming it `ViewState`?

Sources/Components/SaveSearchView/SearchListEmptyViewModel.swift
vadymmarkov: I would probably use `struct` here. With a `protocol` we'll still have to create a struct in the app, which is a bit unnecessary in this case.

Sources/Components/SaveSearchView/SaveSearchPromptView.swift
```diff
@@ -0,0 +1,98 @@
...
+        button.setContentHuggingPriority(.defaultHigh, for: .vertical)
+        return button
+    }()
+
+    private lazy var dismissButton: UIButton = {
+        let button = UIButton(withAutoLayout: true)
+        button.setImage(UIImage(named: .remove), for: .normal)
+        button.tintColor = .textSecondary
+        button.addTarget(self, action: #selector(dismissButtonTapped), for: .touchUpInside)
+        button.translatesAutoresizingMaskIntoConstraints = false
```
vadymmarkov: No need for this since the we use `withAutoLayout: true`

Sources/Components/SaveSearchView/SearchListEmptyView.swift
```diff
@@ -0,0 +1,119 @@
...
+public protocol SearchListEmptyViewDelegate: AnyObject {
+    func searchListEmptyViewDidSelectActionButton(_ searchListEmptyView: SearchListEmptyView, for state: SearchListEmptyView.ViewState)
+}
+
+@objc public class SearchListEmptyView: UIView {
+
+    // MARK: - Public properties
+
+    public enum ViewState: Int {
+        case initial = 0
```
vadymmarkov: I think it starts with zero by default.

vadymmarkov: BTW, does this `ViewState` need to be in `FinniversKit`? It doesn't seem to be used anywhere in this view, so perhaps it could be stored in the app. What do you think?

ninarg: I just thought it was more messy to add it to the controller instead, since the swift extension can't contain stored properties, I would have to add it to objc and try to make that work, which is probably possible. It might be worth the time?

vadymmarkov: I see. No, I don't think it's worth it, it would be better to refactor that objc controller to Swift one day. But for now, let's keep it in `FinniversKit` 😉

ninarg: True 😄

PR #732, opened 2019-12-18
Update design of native adverts

hakloev:
**NB: Breaking, so won't merge until the PR in the iOS-app is ready and approved**

# Why?

Reklameprodukter wants to align the design of native adverts, both native and content.

# What?

I've replaced the old `NativeAdvertView` and `NativeContentAdvertView` with three views representing the different ways we show a native ad: `NativeAdvertGridView`, `NativeAdvertListView` and `NativeAdvertContentView`. Also, I've added `NativeAdvertDetailsContainer` which contains the view for showing the ad ribbon, company name, logo and title. 

I've tried a new approach for this design by using the size classes to add the different styles for phone and tablet. I quite satisfied with the solution as it is, but would like any feedback if I've done it in a suboptimal way. The only thing I'm not that happy about and really would like feedback on, is how I solved the problem of using different fonts on the title. 

To clearly state the difference between the two, I split the demo view into two separate views: `NativeAdvertDemoView` and `NativeAdvertContentDemoView`.

# Show me

_Disclaimer: I really don't know why these images are rotated 🤔_

| | Before | After |
| -- | --- | --- |
| iPhone | Simulator Screen Shot - iPhone 11 Pro - 2019-12-18 at 13 22 40 | Simulator Screen Shot - iPhone 11 Pro - 2019-12-18 at 13 19 18 |
| | | Simulator Screen Shot - iPhone 11 Pro - 2019-12-18 at 13 19 22 |
| iPad | Simulator Screen Shot - iPad Air (3rd generation) - 2019-12-18 at 13 21 55 | Simulator Screen Shot - iPad Air (3rd generation) - 2019-12-18 at 13 20 48 |
| | | Simulator Screen Shot - iPad Air (3rd generation) - 2019-12-18 at 13 20 42 |


Sources/Components/NativeAdvert/NativeAdvertDetailsContainer.swift
```diff
@@ -0,0 +1,154 @@
...
+    // MARK: - Overrides
+
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+
+        setConstraints()
+        setFonts()
+    }
+
+    private func setConstraints() {
```
osanoj: Maybe you could rename the `setConstraints`  to something more clear about which constraints are affected, so `setupSizeClassConstraints` or something similar?

Sources/Components/NativeAdvert/NativeAdvertGridView.swift
```diff
@@ -0,0 +1,142 @@
...
+        if traitCollection.horizontalSizeClass == .regular {
+            NSLayoutConstraint.deactivate(compactConstraints)
+            NSLayoutConstraint.activate(regularConstraints)
+        } else {
+            NSLayoutConstraint.deactivate(regularConstraints)
+            NSLayoutConstraint.activate(compactConstraints)
+        }
+    }
+
+    private func setColors() {
```
osanoj: Same comment for the `setFonts`, `setColors` methods, maybe `setupSizeClassFonts`, `setupSizeClassColors` is better since it doesn't setup all colors and fonts

hakloev: Yes, I agree. It's probably less verbose to do that. I'll fix it!

osanoj:
> I've tried a new approach for this design by using the size classes to add the different styles for phone and tablet. I quite satisfied with the solution as it is, but would like any feedback if I've done it in a suboptimal way. The only thing I'm not that happy about and really would like feedback on, is how I solved the problem of using different fonts on the title.

Your solution regarding size classes looks ok to me. But if anyone has a better way to do it please give some feedback 🙂

PR #735, opened 2019-12-20
Dark mode fixes

ninarg:
# Why?

The app is not ready for dark mode.

# What?

Went through the app and fixed all dark mode issues I could find.

# Show me

Showing the biggest change below.

| Before | After |
| --- | --- |
| img_0883 | img_0882 |


Sources/Components/Profile/ReputationView.swift
```diff
@@ -325,7 +325,7 @@ extension ReputationView {
 
         let imageView = UIImageView(image: model.category.icon)
         imageView.translatesAutoresizingMaskIntoConstraints = false
-        imageView.tintColor = .stone
+        imageView.tintColor = .textSecondary
```
shredlocker: Is this correct? A bit confusing to assign a text color to an image view

fespinoza: in this case @ninarg you can make a private extension of UIColor to assign a name to this semantic color and return `.textSecondary` as value, as the color may be correct, but not the semantics of it

PR #736, opened 2020-01-03
Favorites: refactor bottom sheet subclasses into views

vadymmarkov:
# Why?

Because we want to gradually move away from the old bottom sheet implementation and start using the new one.

# What?

- Use regular views, unaware of their presentation style, instead of custom view controllers and `BottomSheet` subclasses:
  - `FavoriteAdActionSheet` ->  `FavoriteAdActionView`
  - `FavoriteAdCommentSheet` -> `FavoriteAdCommentInputView `
  - `FavoriteAdSortingSheet` -> `FavoriteAdSortingView`
  - `FavoriteFolderActionSheet ` -> `FavoriteFolderActionView`
- Update demo and snapshot tests

# Show me

No UI changes.


Sources/Fullscreen/FavoriteAdCommentInput/FavoriteAdCommentInputView.swift
```diff
@@ -176,61 +142,35 @@ final class FavoriteAdCommentViewController: UIViewController {
...
-        loadingView.startAnimating()
-        navigationItem.setRightBarButton(loadingBarButton, animated: true)
-        cancelButton.isEnabled = false
-    }
-
-    public func stopLoading() {
-        loadingView.stopAnimating()
-        navigationItem.setRightBarButton(saveButton, animated: true)
-        cancelButton.isEnabled = true
+        scrollViewBottomConstraint.constant = -keyboardOffset
```
bstien: I guess this line could be moved into the animation block above, since the rest of this method is removed

PR #738, opened 2020-01-09
Beta Feature View for Image Search

fespinoza:
# Why?

We want to show a bottom sheet with info about the image search being a beta feature of the app

# What?

- Add a fullscreen view that will be presented in the app as a bottom sheet
- Add `betaPill` asset to FinniversKit

# Show me

Screenshot 2020-01-09 at 10 46 52


Demo/Fullscreen/BetaFeatureView/BetaFeatureDemoView.swift
```diff
@@ -0,0 +1,41 @@
...
+        setup()
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    func setup() {
+        addSubview(betaFeatureView)
+        layoutMargins = UIEdgeInsets(top: 0, leading: 16, bottom: 0, trailing: 16)
```
vadymmarkov: Can use `mediumLargeSpacing` instead of 16.

Demo/Fullscreen/BetaFeatureView/BetaFeatureDemoView.swift
```diff
@@ -0,0 +1,41 @@
...
+        NSLayoutConstraint.activate([
```
vadymmarkov: Can we use `fillInSuperviewLayoutMargins` here?

Sources/Fullscreen/BetaFeatureView/BetaFeatureView.swift
```diff
@@ -0,0 +1,150 @@
...
+    }()
+
+    private lazy var secondButton: Button = {
+        let button = Button(style: .flat, withAutoLayout: true)
+        button.addTarget(self, action: #selector(handleTapOnSecondButton), for: .touchUpInside)
+        button.setContentHuggingPriority(.defaultHigh, for: .vertical)
+        return button
+    }()
+
+    private let betaPillSize = CGSize(width: 52, height: 24)
```
vadymmarkov: Does it need to be an instance var? It seems like it's used only once in `setup`?

fespinoza: it doens't have to, but to me it should more intention rather than putting the concrete values in the constraint as "magic" values

vadymmarkov: Still can be a local var in `setup` I think.

Demo/Fullscreen/BetaFeatureView/BetaFeatureDemoView.swift
fespinoza: not really, as in this case the bottom contraint is `lessThanOrEqualTo`

PR #741, opened 2020-01-14
New payment screen

robinsalehjan:
# Why?

Would like to redesign the payment options screen.

# What?

Some oddities:
- I could not use the `AnimatedRadioButtonView` as I only needed the radio button functionality and not the whole view. I decided to use the underlying `SelectionboxItem` instance instead and implement the logic of selection and deselection in the `didSelectRowAt` method.

- The detailLabel `+ XX kr`only appears if there's additional fees (e.g `SMS`)

- The `TotalSum` can be updated from the outside and does not depend on the content of the `OrderSummaryView` or the `CollapseView`

# Show me

| Collapse | Expanded |
| --- | --- |
| Simulator Screen Shot - iPhone 11 Pro Max - 2020-01-14 at 11 50 29 | Simulator Screen Shot - iPhone 11 Pro Max - 2020-01-14 at 11 50 38 |


Sources/Recycling/ListViews/PaymentOptions/Cell/PaymentOptionsListViewCell.swift
```diff
@@ -0,0 +1,130 @@
...
+
+    private lazy var selectionBox: SelectionboxItem = {
+        let animatedImageView = AnimatedRadioButtonView(frame: .zero)
+        let selectionBox = SelectionboxItem(index: 0, animatedImageView: animatedImageView)
+        selectionBox.translatesAutoresizingMaskIntoConstraints = false
+        return selectionBox
+    }()
+
+    private lazy var titleLabel: UILabel = {
+        let label = UILabel(withAutoLayout: true)
```
hakloev: Any reason for not using `Label` in these labels?

robinsalehjan: Totally forgot, will fix.

PR #743, opened 2020-01-15
🚗 MP: Seller self declaration

bstien:
# Why?
Motor has a market problem to add seller self declaration to the object page.

# What?
- Added models, view, demo and snapshot test.

# Show me
Screenshot 2020-01-15 at 13 42 55


Demo/Components/SelfDeclaration/SelfDeclarationDemoView.swift
```diff
@@ -0,0 +1,64 @@
...
+        scrollView.fillInSuperview()
+        scrollView.addSubview(selfDeclarationView)
+
+        NSLayoutConstraint.activate([
+            selfDeclarationView.leadingAnchor.constraint(equalTo: leadingAnchor, constant: .mediumLargeSpacing),
+            selfDeclarationView.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -.mediumLargeSpacing),
+            selfDeclarationView.topAnchor.constraint(equalTo: scrollView.topAnchor, constant: .mediumLargeSpacing),
+            selfDeclarationView.bottomAnchor.constraint(equalTo: scrollView.bottomAnchor, constant: -.mediumLargeSpacing),
+        ])
+
```
fespinoza: i guess you could do something like

```swift
scrollView.layoutMargins = UIEdgeInsets(all: .mediumLargeSpacing)
selfDeclarationView.fillInSuperviewMargins()
```

the demo view's leading and trailing anchors will be the same as the scroll view

Sources/Components/SelfDeclaration/Views/SelfDeclarationItemView.swift
```diff
@@ -0,0 +1,56 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+
```
fespinoza: extra line :O

Demo/Components/SelfDeclaration/SelfDeclarationDemoView.swift
bstien: It works to constrain the view, but then it's not possible to scroll in the scrollView :/

PR #744, opened 2020-01-15
Viewing cells for realestate

ninarg:
# Why?

We want to make viewings more prominent on the object page. More users of the feature is an advantage for effect documentation.

# What?

- Add `ViewingsView` with `ViewingsViewModel`, which displays a header, note and a table with all viewings
- Add `ViewingCell` with `ViewingCellViewModel`, which displays a single viewing

_The views have already been reviewed in ios-app, I moved them here instead. But there are a few changes. Eg. I made an `NSAttributedString` for the notes instead, to be able to add line spacing. Also removed margins for iPad._ 

# Show me

| Light mode | Dark mode |
| --- | --- |
| IMG_1162 | IMG_1163 |


Sources/Components/Viewings/ViewingCellViewModel.swift
```diff
@@ -0,0 +1,20 @@
...
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import Foundation
+
+public struct ViewingCellViewModel {
+
+    let weekday: String?
+    let month: String?
+    let date: String?
```
osanoj: Is this actually `day`/`dayOfMonth`? With name `date` it makes me think it consists of the full date

Sources/Components/Viewings/ViewingsView.swift
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    // MARK: - Setup
+
+    private func setup() {
+        if UIDevice.isIPad() {
```
osanoj: Not sure if we need this switching of margin inside the cell here? We have some mechanisms for using different margins for the cells on the objectpage, which might be better suited. But I guess the contentMargin should always be zero here and then the whole cell/view getting a horizontal margin when used 🤔 

Do you think that could work?

Sources/Components/Viewings/ViewingCellViewModel.swift
ninarg: Yeah, day makes more sense 👍

Demo/Components/Viewings/ViewingsDemoView.swift
```diff
@@ -0,0 +1,39 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import FinniversKit
+
+class ViewingsDemoView: UIView {
+
+    private lazy var viewingsView = ViewingsView(frame: .zero)
```
fespinoza: you can just say

`private lazy var viewingsView = ViewingsView(withAutoLayout: true)`

and remove the `viewingsView.translatesAutoresizingMaskIntoConstraints`

Sources/Components/Viewings/ViewingsView.swift
ninarg: On the phone, I actually want regular margins on the left, and no margins on the right (because the table separator in the viewings shouldn't have a margin on the right). To achieve this, I used

`override func shouldAlwaysRenderAsFullWidth() -> Bool {
        return UIDevice.current.userInterfaceIdiom == .phone }`

from `FINObjectElementItem` and then I added the regular left margins to header and note label. On iPad, I had to remove these margins for the text to be aligned with other elements in the side bar. There might be some other options that I haven't found though?

Sources/Components/Viewings/ViewingCell.swift
```diff
@@ -0,0 +1,122 @@
...
+
+    required init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+        setup()
+    }
+
+    // MARK: - Setup
+
+    private func setup() {
+        addSubview(viewingStackView)
```
fespinoza: it's a good practice when working with `UITableViewCell`/`UICollectionViewCell` to add child views into the cell's `contentView` instead of directly (because of other features of cells, like accessory views)

Sources/Components/Viewings/ViewingsView.swift
```diff
@@ -0,0 +1,144 @@
...
+    private lazy var tableView: UITableView = {
+        let tableView = UITableView(frame: .zero)
+        tableView.register(ViewingCell.self)
+        tableView.translatesAutoresizingMaskIntoConstraints = false
+        tableView.dataSource = self
+        tableView.rowHeight = ViewingCell.cellHeight
+        tableView.estimatedRowHeight = ViewingCell.cellHeight
+        tableView.isScrollEnabled = false
+        tableView.separatorColor = .tableViewSeparator
+        tableView.separatorInset = UIEdgeInsets(top: 0, left: ViewingCell.dateViewWidth + ViewingCell.contentSpacing, bottom: 0, right: 0)
```
fespinoza: you can just say:

```
tableView.separatorInset = UIEdgeInsets(leading: ViewingCell.dateViewWidth + ViewingCell.contentSpacing)
```

Sources/Components/Viewings/ViewingsView.swift
```diff
@@ -0,0 +1,144 @@
...
+    private lazy var noteLabel: Label = {
+        let label = Label(withAutoLayout: true)
+        label.isHidden = true
+        label.numberOfLines = 0
+        label.lineBreakMode = .byWordWrapping
+        return label
+    }()
+
+    private lazy var tableView: UITableView = {
+        let tableView = UITableView(frame: .zero)
```
fespinoza: `withAutoLayout: true`

Sources/Components/Viewings/ViewingsView.swift
fespinoza: @vadymmarkov do you know if we have another way to do something like this for views?

@ninarg in general we want to avoid `UIDevice.isIPad` as an iPad app can be **potentially** resized and made look as thin as an iPhone, then in view controllers we usually have `traitCollection.horizontalSizeClass` as `regular` or `compact`

PR #749, opened 2020-01-21
Motor MP: Horizontal iconcollection

bstien:
# Why?
The Motor MP requires some ad attributes to be presented within a grid of icons. We already have such an element, `IconCollectionView`, that can be used to present this list of horizontally aligned cells.

# What?
- Added enum `IconCollectionView.Alignment`, that is provided to the class upon instantiation.
- Renamed `IconCollectionViewCell` into `VerticalIconCollectionViewCell`.
- Added cell type `HorizontalIconCollectionViewCell`.
- Small refactorings to size calculation of cells.

# Show me
| Vertical | Horizontal |
| --- | --- |
| Screenshot 2020-01-21 at 11 48 09 | Screenshot 2020-01-21 at 11 22 12 |


Sources/Components/IconCollection/HorizontalIconCollectionViewCell.swift
```diff
@@ -0,0 +1,116 @@
...
+        label.numberOfLines = 0
+        return label
+    }()
+
+    private lazy var bodyLabel: UILabel = {
+        let label = UILabel(withAutoLayout: true)
+        label.font = .bodyStrong
+        label.textColor = .textPrimary
+        label.adjustsFontSizeToFitWidth = true
+        label.numberOfLines = 0
```
fespinoza: are these two lines conflicting? (58-59)
- number of lines = 0 => use as many lines as you need
- adjustsFontSizeToWidth => if you don't have space shrink the font size

also, why not a use a Label instead?

Sources/Components/IconCollection/HorizontalIconCollectionViewCell.swift
```diff
@@ -0,0 +1,116 @@
...
+        bodyLabel.text = viewModel.text
+        accessibilityLabel = "\(viewModel.title): \(viewModel.text)"
+    }
+
+    private func setup() {
+        isAccessibilityElement = true
+
+        addSubview(iconImageView)
+        addSubview(titleLabel)
+        addSubview(bodyLabel)
```
fespinoza: should these be added to the `contentView` of the collection view cell?

bstien: Nice catch!😅 Will fix

Sources/Components/IconCollection/HorizontalIconCollectionViewCell.swift
bstien: Yup, they're definitely conflicting. Removing `adjustsFontSizeToFitWidth` and using `Label` instead 👍

bstien:
Screenshot 2020-01-21 at 13 35 32

Can anyone explain why the font size decreased when I changed from `UILabel` with `UIFont.body` to `Label` with `Label.Style.body`? 🤔 They both point to the same font, `UIFont.body`.

fespinoza:
can it be just padding?

bstien:
Aaaaah, `Label.setup()` has this line, which is probably the intended "culprit":

```swift
adjustsFontForContentSizeCategory = true
```

bstien:
It seems to only happen in the tests. In my simulator the fonts are exactly the same and I don't get any visual difference at all.

Updating the snapshot image and moving on.

PR #751, opened 2020-01-22
Fix: IconCollectionView calculations

bstien:
# Why?
In my previous PR (#749) I made some incorrect calculations regarding the size of the cells/labels and the `IconCollectionView` itself. This PR fixes those mistakes.

This wasn't visible in the Demo-app, because here we use autolayout. In the app, we set the cell height based on what the view returns from `systemLayoutSizeFitting(_:withHorizontalFittingPriority:verticalFittingPriority:)`, and this is where the calculations were incorrect.

# What?
- Use already existing FK method for calculating label height.
- Remove some non-existing spacing from height calculations.
  - Made the view taller than it actually was.
- Add some spacing to calculations.
  - To make the sum of label heights correct.

# Show me

| Before | After |
| --- | --- |
| Screenshot 2020-01-22 at 15 17 58 | Screenshot 2020-01-22 at 15 38 56 |


Sources/Components/IconCollection/HorizontalIconCollectionViewCell.swift
```diff
@@ -84,7 +69,12 @@ public class HorizontalIconCollectionViewCell: UICollectionViewCell {
         iconImageView.image = viewModel.image
         titleLabel.text = viewModel.title
         bodyLabel.text = viewModel.text
-        accessibilityLabel = "\(viewModel.title): \(viewModel.text)"
+
+        if let title = viewModel.title {
+            accessibilityLabel = "\(title): \(viewModel.text)"
+        } else {
+            accessibilityLabel = "\(viewModel.text)"
+        }
```
fespinoza: kudos for caring about this :)

PR #754, opened 2020-01-23
Don't set custom line height for bullet point lists

vadymmarkov:
# Why?

Because if we set custom line height it should be done everywhere across the app, not only for bullet lists.

# What?

Don't set custom line height for bullet point lists.

# Show me

**Affected components**

| Before | After |
| --- | --- |
| finn_tr_old iPhone copy | testConsentTransparencyInfo_new iPhone copy |
| testColumnListsView iPhone copy_old | testColumnListsView iPhone_new |


fespinoza:
@vadymmarkov I guess an option could be to pass the lineHeight as parameter of the function?

vadymmarkov:
@fespinoza Yes, but I talked to Thuy and he doesn't want us to set custom line heights since it leads to inconsistency.

PR #756, opened 2020-01-23
Dark mode fix for utility button

ninarg:
# Why?

The utility button's border is not updated instantly when switching from/to dark mode. When we only set the button's layout once it is initialized, the app will sometimes not draw it correctly when the environment changes. This can be resolved by moving the setup of buttons to a lifecycle method.

# What?

Since the utility button is used multiple places, I tried to fix this with a general approach in our Button class. I tried to run the whole `setup()` method from `layoutSubviews`. A problem with doing this is that the button in some cases is changed after the setup, when the button shouldn’t completely follow one of the styles, so with this solution, these changes were overwritten.

I ended up resetting the `.utility` style for all utility buttons in `layoutSubviews`. I think it’s suboptimal that we have to do this every time we create a new utility button. Or is it okay? Any thoughts? 🤔


fespinoza:
I would try to solve the button itself,

when the button instance initializes, the traits are empty and then when getting a cgColor from a UIColor this result in a concrete value that won't change on trait collection change.

I would try/explore by setting the border of the button in `layoutSubviews` of `Button` instead. `layoutSubviews` does have the value of `traitCollection` set, then UIColors will return different values for light/dark modes.

ninarg:
Putting this in `Button` works. The downside is that all usages of buttons are forced to have the default `borderColor` for their style. (It affects `FeedbackView`, which is changing the border color of default button)

```
public override func layoutSubviews() {
        super.layoutSubviews()
        layer.borderColor = style.borderColor(forState: state)
}
```

Sources/Components/Feedback/FeedbackView.swift
```diff
@@ -319,6 +317,7 @@ private class ButtonView: UIView {
             positiveButton.removeFromSuperview()
             negativeButton.removeFromSuperview()
         }
+        negativeButton.style = Button.Style.negativeButton()
```
fespinoza: you can do `let button = Button(style: .negativeButton(), size: .small)` in the lazy var instead

PR #765, opened 2020-01-31
Refactor: key value grid view

fespinoza:
# Why?

For the motor MP I realized than having a UICollectionView inside a UICollectionViewCell was way more complicated that it had to be for this component, so I tried to simplify the underlying implementation by using stack views instead.

# What?

- Rename `KeyValueCollectionView` to `KeyValueGridView`
- Refactor the implementation by using stack views instead of UICollectionView
- Remove files that are not needed anymore

This is technically a breaking change, due #759 was merged into master (although it wasn't used yet in the iOS app), but i guess i will still do a major release after this is merged.

I will also wait to merge this after testing the integration with the object page.


fespinoza:
I checked the integration with the object page 👍, this is ready to be merged

Demo/Components/KeyValueCollectionView/KeyValueGridDemoView.swift
```diff
@@ -4,15 +4,20 @@
...
+class KeyValueGridDemoView: UIView {
+    private lazy var keyValueGridView: KeyValueGridView = {
+        let view = KeyValueGridView(withAutoLayout: true)
         view.configure(with: sampleData)
         view.backgroundColor = .bgSecondary
         view.numberOfColumns = numberOfColumnsForTraits
         return view
     }()
 
+    private lazy var scrollView: UIScrollView = {
```
bstien: Could be just

```suggestion
    private lazy var scrollView = UIScrollView(withAutoLayout: true)
```

PR #768, opened 2020-02-04
Improvement: allow enabling/disabling the collapse behavior for CollapsibleContentView

fespinoza:
# Why?

In certain cases we don't want to expand/collapse the content inside the
CollpsibleContentView which is done for the specifications list in the
new motor object page. In that case we use this view to show the same
style of header for that content.

# What?

- add `isEnabled` to `CollapsibleContentView`


bstien:
Just a thought: will the view collapse/expand if you tap anywhere within the header, or just the arrow on the right?

fespinoza:
spelling is difficult man xD

fespinoza:
it's anywhere in the header

PR #776, opened 2020-02-07
Adin 7401 userads redesign

hakloev:
# Why?

We want to adjust the design of user ads in the ad insertion. 

# What?

* Added a new generic table view cell: `UserAdTableViewCell` with two different stylings; `.default` and `.compressed`. In general, this is a modification of `UserAdsListViewCell` that tries to decouple FinniversKit from knowing anything domain specific, like i.e. ad statuses. 
* Added a protocol for the view model of the cell: `UserAdTableViewCellViewModel`
* Exposed `overrideStyle` from `Button.Style` since it is useful (and needed) outside of FinniversKit

The idea is to release this redesign in two steps, so `UserAdTableViewCell` and `UserAdsListViewCell` will live side by side for some time, until we've moved the logic of `UserAdsListView`into the ad-insertion module. Hence, the ads from `UserAdsFactory` is used for the cell demo view. The next step of this process will be completed when we have a new design for emphazised ads. 

# Show me

| Compact | Regular |
| --- | --- |
| Simulator Screen Shot - iPhone 11 Pro - 2020-02-06 at 13 28 44 | Simulator Screen Shot - iPhone 11 Pro - 2020-02-06 at 13 28 40 |


Demo/Cells/UserAdTableViewCell/UserAdCellDemoView.swift
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    func remoteImageView(_ view: RemoteImageView, loadImageWithPath imagePath: String, imageWidth: CGFloat, completion: @escaping ((UIImage?) -> Void)) {
+        guard let url = URL(string: imagePath) else {
+            completion(nil)
+            return
+        }
+
+        // Demo code only.
+        let task = URLSession.shared.dataTask(with: url) { data, _, _ in
```
bstien: I've learned that these network calls should have some sleep included, so we don't mess with the snapshot tests.

From `FrontPageDemoView`:
```swift
// Demo code only.
let task = URLSession.shared.dataTask(with: url) { data, _, _ in
    usleep(50_000)
    DispatchQueue.main.async {
        if let data = data, let image = UIImage(data: data) {
            completion(image)
        } else {
            completion(nil)
        }
    }
}
```

Sources/Cells/UserAdTableViewCell/UserAdTableViewCell.swift
```diff
@@ -0,0 +1,220 @@
...
+    private var model: UserAdTableViewCellViewModel?
+
+    private var style: Style? = .default {
+        didSet {
+            setupStyleConstraints()
+        }
+    }
+
+    private static var defaultImageWidth: CGFloat = 80
+    private static var compressedImageWidth: CGFloat = 50
```
bstien: Could these lines be `let`s instead?

Sources/Cells/UserAdTableViewCell/UserAdTableViewCell.swift
```diff
@@ -0,0 +1,220 @@
...
+
+}
+
+// MARK: - ImageLoading
+
+extension UserAdTableViewCell: ImageLoading {
+
+    public func loadImage() {
+        if let imagePath = model?.imagePath {
+            adImageView.loadImage(for: imagePath, imageWidth: .veryLargeSpacing, loadingColor: loadingColor, fallbackImage: fallbackImage)
```
bstien: We should probaby try to fetch the image with the correct imageWidth here.

```suggestion
            let imageWidth = style == .compressed ? compressedImageWidth : defaultImageWidth
            adImageView.loadImage(for: imagePath, imageWidth: imageWidth, loadingColor: loadingColor, fallbackImage: fallbackImage)
```

Sources/Cells/UserAdTableViewCell/UserAdTableViewCell.swift
```diff
@@ -0,0 +1,224 @@
...
+        }
+    }
+
+    public var loadingColor: UIColor? = .toothPaste
+
+    // MARK: - Private properties
+
+    private var model: UserAdTableViewCellViewModel?
+
+    private var style: Style? = .default {
```
nickfalk: Have you considered just getting rid of this property(?) The code handling it is very close to a pure functional form already. ☺️

Sources/Cells/UserAdTableViewCell/UserAdTableViewCell.swift
```diff
@@ -0,0 +1,224 @@
...
+        case compressed
+    }
+
+    public var remoteImageViewDataSource: RemoteImageViewDataSource? {
+        didSet {
+            adImageView.dataSource = remoteImageViewDataSource
+        }
+    }
+
+    public var loadingColor: UIColor? = .toothPaste
```
nickfalk: Doesn't seem like we're doing much with this rather than using it as `.toothPaste`, but this is perhaps something we want to consider?

Sources/Cells/UserAdTableViewCell/UserAdTableViewCell.swift
```diff
@@ -0,0 +1,224 @@
...
+
+extension UserAdTableViewCell: ImageLoading {
+
+    public func loadImage() {
+        guard let imagePath = model?.imagePath else {
+            adImageView.image = fallbackImage
+            return
+        }
+
+
```
nickfalk: Being a teensy-teensy bit too generous with your line-breaks. 😛

Sources/Cells/UserAdTableViewCell/UserAdTableViewCell.swift
hakloev: It's set from the data source in the ad-insertion module.

PR #778, opened 2020-02-10
Motor MP: new safety elements view

fespinoza:
# Why?

We needed a redesign of the safety elements for the motor object page, but also
we needed a special layout for iPhone than the one we originally had.

# What?

- Add `SafetyElementsView` component
- Fix some swift lint issues

Internally the `SafetyElementsView` uses internal views
`SafetyElementsView.CompactView` and `SafetyElementsView.RegularView` to render the content
in the right way for the given configuration.

There is a certain level of internal reuse of subviews.

The layout that is used to render the view is determined by `useCompactLayout`,
so it can be manually set.

I'm also using nested classes to define the internal views, to some-what avoid long names and
not polluting the main namespace with names of internal only views.

# Show me

iphone
ipad
ipad with multiple elements


fespinoza:
ready to review mr @bstien

Sources/Components/SafetyElementsView/Regular/RegularView.swift
```diff
@@ -0,0 +1,144 @@
...
+                .arrangedSubviews
+                .compactMap({ $0 as? HeaderView })
+                .forEach({ $0.isActive = false })
+            (gesture.view as? HeaderView)?.isActive = true
+            setActiveElement(to: index)
+            delegate?.safetyElementsRegularView(self, didSelectElementAt: index)
+        }
+
+        private func setActiveElement(to index: Int) {
+            guard viewModels.count > 0 else { return }
```
bstien: Could rather be

```suggestion
            guard !viewModels.isEmpty else { return }
```

Sources/Components/SafetyElementsView/SafetyElementViewModel.swift
```diff
@@ -0,0 +1,21 @@
...
+//
+
+import Foundation
+
+public struct SafetyElementViewModel {
+    let title: String
+    let icon: UIImage
+    let body: String
+    let topLink: LinkButtonViewModel?
+    let bottomLink: LinkButtonViewModel?
```
bstien: In case we need to debug stuff in the app it'd be great if these were marked as `public` so we can access them :)

Sources/Components/SafetyElementsView/SafetyElementsView.swift
```diff
@@ -0,0 +1,77 @@
...
+    // MARK: - Initializers
+    private lazy var compactView: CompactView = {
+      let view = CompactView(withAutoLayout: true)
+
+      return view
+    }()
+
+    private lazy var regularView: RegularView = {
+      let view = RegularView(withAutoLayout: true)
+
```
bstien: Could be 

```swift
private lazy var regularView = RegularView(withAutoLayout: true)
```

Sources/Components/SafetyElementsView/SafetyElementsView.swift
```diff
@@ -0,0 +1,77 @@
...
+    public weak var delegate: SafetyElementsViewDelegate?
+
+    // MARK: - Private properties
+    private var viewModels: [SafetyElementViewModel] = []
+    private var selectedElementIndex: Int = 0
+
+    // MARK: - Initializers
+    private lazy var compactView: CompactView = {
+      let view = CompactView(withAutoLayout: true)
+
```
bstien:
```swift
private lazy var compactView = CompactView(withAutoLayout: true)
```

PR #779, opened 2020-02-10
Motor MP: LinkButton list

bstien:
# Why?
We need a component that can present a list of links, with an optional external icon.

I've refactored the component `LinkButtonView` to serve this purpose, and added `LinkButtonListView` to contain a list of these. The previous usecase for `LinkButtonView` was in `ObjectPagePriceView`, and this has been refactored into using `LinkButtonListView` instead.

# What?
- Add property `isExternal` to `LinkButtonViewModel`.
- Add an icon to `LinkButtonView`, if `isExternal` is `true`.
- Add view/component `LinkButtonListView`.
- Use `LinkButtonListView` in `ObjectPagePriceView`.
- Add demo + update snapshot reference images.

# Show me

Screenshot 2020-02-10 at 14 22 24

| Before | After |
| --- | --- |
| Screenshot 2020-02-06 at 09 15 57 | Screenshot 2020-02-10 at 13 57 59 |


UnitTests/ComponentViewTests.swift
```diff
@@ -262,4 +262,8 @@ class ComponentViewTests: XCTestCase {
     func testObjectPagePriceView() {
         snapshot(.objectPagePriceView)
     }
+
+    func testLinkButtonListView() {
+        snapshot(.linkButtonListView)
```
fespinoza: check the snapshot image, it seem there is an alignment issue between the button and subtitle (check #778 i had that issue too) 

Screenshot 2020-02-10 at 14 59 59

Sources/Components/LinkButton/LinkButtonView.swift
```diff
@@ -17,21 +17,37 @@ class LinkButtonView: UIView {
...
     private lazy var linkButton: Button = {
         let button = Button(style: linkButtonStyle, size: .normal, withAutoLayout: true)
         button.addTarget(self, action: #selector(handleTap), for: .touchUpInside)
         return button
     }()
 
+    private lazy var externalImageView: UIImageView = {
+        let imageView = UIImageView(withAutoLayout: true)
+        imageView.image = externalImage
+        imageView.tintColor = .sardine
```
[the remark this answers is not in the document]
bstien: You're right, it's not dark mode proof. We have the option to use one of these, but I'm not sure which one to pick 🤔 

```swift
imageView.tintColor = .btnDisabled
imageView.tintColor = .textDisabled
imageView.tintColor = .tableViewSeparator
imageView.tintColor = dynamicColorIfAvailable(defaultColor: .sardine, darkModeColor: .darkSardine)
```

UnitTests/ComponentViewTests.swift
bstien: What was the solution? I can't figure out what's causing this in the firsst place 🙈

Sources/Components/LinkButton/LinkButtonView.swift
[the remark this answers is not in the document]
fespinoza: i would add a private UIColor extensions at the bottom of the file with a semantic name for the icon tint color with the implementation

```
dynamicColorIfAvailable(defaultColor: .sardine, darkModeColor: .darkSardine)
```

UnitTests/ComponentViewTests.swift
bstien: _For future reference_

The solution was to set `button.contentHorizontalAlignment = .leading`.

PR #781, opened 2020-02-11
Bugfix: Constraint ambiguity when measuring UserAdTableViewCell height

hakloev:
# Why?

I want the height calculation of the cell to not be ambiguous.

# What?

Lowered priority on the content stack's bottom anchor, so it ain't ambiguous with regards to the content view height constraint set by the table view on the previous layout.

# Show me
_No UI changes_


ag12:
just question why not use `defaultHigh`?

hakloev:
I'm not sure I understand your question? 🤔 This cell uses auto layout to measure its height, and it's an active constraint set on the content view by the system that makes the layout ambiguous.

PR #791, opened 2020-02-19
Motor MP: Title view

bstien:
# Why?
We want to get rid of proxy layout for the title view on the objectPage. Like the price view in #772 this one is named `ObjectPageTitleView` since it'll only have a single usecase.

Both `title` and `subtitle` are optionals, since we can't be sure if an ad will contain both/either/none.

# What?
- Add `ObjectPageTitleView`.
- Add demo + snapshot reference image.

# Show me
| Title + subtitle | Title | Subtitle |
| --- | --- | --- |
| Screenshot 2020-02-19 at 09 41 27 | Screenshot 2020-02-19 at 09 41 31 | Screenshot 2020-02-19 at 09 41 36 |


Demo/Components/ObjectPageTitleView/ObjectPageTitleDemoView.swift
```diff
@@ -0,0 +1,54 @@
...
+        titleView = nil
+
+        let newTitleView = ObjectPageTitleView(titleStyle: titleStyle, subtitleStyle: subtitleStyle, withAutoLayout: true)
+        addSubview(newTitleView)
+
+        newTitleView.leadingAnchor.constraint(equalTo: leadingAnchor, constant: .mediumSpacing).isActive = true
+        newTitleView.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -.mediumSpacing).isActive = true
+        newTitleView.centerYAnchor.constraint(equalTo: centerYAnchor).isActive = true
+        newTitleView.configure(withTitle: title, subtitle: subtitle)
+        self.titleView = newTitleView
```
fespinoza: just nitpicking
if you have multiple constraints i think it's nicer to write

```swift
NSLayoutConstraint.activate([
])
```

instead of multiple `isActive = true`.

Also, why do you need to create a new title view each time? can you use the instance and just call the `configure` method within the tweaking actions?

Demo/Demo/DemoViews/ComponentDemoViews.swift
```diff
@@ -68,9 +68,10 @@ public enum ComponentDemoViews: String, CaseIterable {
     case linkButtonListView
     case safetyElementsView
     case contractActionView
+    case objectPageTitleView
 
     public static var items: [ComponentDemoViews] {
-        return allCases.sorted { $0.rawValue < $1.rawValue }
+        allCases.sorted { $0.rawValue < $1.rawValue }
```
fespinoza: NO RETURN! :P

bstien: 🙈

Demo/Components/ObjectPageTitleView/ObjectPageTitleDemoView.swift
bstien: I'll fix the constraints 👍

I did this, because I was unsure if we needed to simulate other markets in the demo, since some of those will have different needs in regard to label styling. We don't have this issue right now, so I'll just keep the one instance instead of recreating the view.

bstien:
We need to have an optional `Ribbon` above the title in this view. I'll add this and request review again.

bstien:
Nevermind my last comment. I need to do some refactoring for the ribbon to work as planned. Merging this one and will open another PR with those changes.

PR #810, opened 2020-03-02
Internal: rename spacing constants to standarize with Android

fespinoza:
Trello ref: https://trello.com/c/tpZJhhFN/130-redifine-spacing-names-and-convensions-for-ios-and-android

# Why?

When we call something medium, we mean: 8pts in iOS and 16pts in Android, then
from the design perspective is quite confusing.

# What?

- Renamed spacing constants:
  - spacingXXS = 2
  - spacingXS  = 4
  - spacingS   = 8
  - spacingM   = 16
  - spacingL   = 24
  - spacingXL  = 32
  - spacingXXL = 64

  This naming is also better in terms that is easy to relate spacingM with
  spacingL, rather than mediumSpacing, mediumLargeSpacing and veryLargeSpacing

  It's also better for autocompletion purposes, since the constants will be
  within the same namespace, then you can see all the available options when you
  type "spacing"

- Keep the previous constants to avoid break projects depending from FinniversKit

  ⚠️ this will produce a lot of warnings in the projects. I will fix them in the
  different dependencies as soon as possible after merge


Demo/DNA/Spacing/SpacingDemoView.swift
```diff
@@ -24,48 +24,48 @@ public class SpacingDemoView: UIView {
...
+        addSubview(spacingXSView)
 
-        let mediumSpacingView = makeLabel(text: "👈        mediumSpacing        👉")
-        addSubview(mediumSpacingView)
+        let spacingSView = makeLabel(text: "👈        spacingS        👉")
+        addSubview(spacingSView)
 
-        let mediumPlusSpacingView = makeLabel(text: "👈        mediumPlusSpacingView        👉")
-        addSubview(mediumPlusSpacingView)
+        let spacingLView = makeLabel(text: "👈        spacingLView        👉")
```
ninarg: spacingL 😄

Demo/DNA/Spacing/SpacingDemoView.swift
```diff
@@ -24,48 +24,48 @@ public class SpacingDemoView: UIView {
...
 
-        let smallSpacingView = makeLabel(text: "👈        smallSpacing        👉")
-        addSubview(smallSpacingView)
+        let spacingXSView = makeLabel(text: "👈        spacingXS        👉")
+        addSubview(spacingXSView)
 
-        let mediumSpacingView = makeLabel(text: "👈        mediumSpacing        👉")
-        addSubview(mediumSpacingView)
+        let spacingSView = makeLabel(text: "👈        spacingS        👉")
+        addSubview(spacingSView)
```
ninarg: Missing demo for spacing M

PR #811, opened 2020-03-02
Make link/subtitle tapable

ag12:
# Why?
We want to add more functionality to the reputation view.  

# What?
Added `UITapGestureRecognizer` to `subtitleLabel` and 
new method `reputationViewLinkWasTapped` to `ReputationViewDelegate`


Sources/Components/Profile/ReputationView.swift
```diff
@@ -89,6 +92,12 @@ public class ReputationView: UIView {
         label.translatesAutoresizingMaskIntoConstraints = false
         label.numberOfLines = 0
         label.textColor = .textAction
+
+        // Make label tapable
+        let gesture = UITapGestureRecognizer(target: self, action: #selector(onLinkTap))
+        label.isUserInteractionEnabled = true
```
robinsalehjan: Maybe move this to be right below the other attributes set on the label?

ag12: Wow 😍

PR #812, opened 2020-03-02
Bump FinniversKit to 35.3.0

fespinoza:
# Changes:

- #810: Internal: rename spacing constants to standarize with Android

☝️ is not a breaking change, but it does deprecate the old names for
spacing constants, that's why i bumped the minor number, although I am
not sure is the correct. (can be a patch bump)


fespinoza:
closing in favor of #813

PR #813, opened 2020-03-02
FKV36

ag12:
# What?
Having a breaking change. Bumping version to 36


fespinoza:
what's the breaking change?

ag12:
Added a new method to `ReputationViewDelegate`

PR #814, opened 2020-03-02
Create TransactionWarningView

robinsalehjan:
# Why?

- Need a way to convey to the users, that the current process is the one they have progressed the furthest with.

# What?

- Create a `TransactionWarningView` that will be placed right below the title for the current transaction process.

# Show me

| Without warning | With warning |
| --- | --- |
| Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-02 at 16 38 37 | Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-02 at 16 38 40 |


robinsalehjan:
This is not in use anywhere, so won't break anything.

PR #820, opened 2020-03-04
Use semantic colours in the TransactionView

robinsalehjan:
# Why?

- Did not look good with dark mode.

# What?

- Use semantic colors for the `TransactionStepView+Style`
- Always use `.licorice` for `TransactionWarningView`

# Show me

| Basic | Dark |
| --- | --- |
| Simulator Screen Shot - iPhone 11 Pro - 2020-03-04 at 09 54 37 | Simulator Screen Shot - iPhone 11 Pro - 2020-03-04 at 09 52 00 |


Sources/Fullscreen/TransactionView/TransactionStepView/TransactionStepView+Style.swift
```diff
@@ -38,9 +38,9 @@ public extension TransactionStepView {
         var titleTextColor: UIColor {
             switch self {
             case .notStarted:
-                return .stone
+                return .textSecondary
             default:
```
bstien: Shouldn't this one be exhaustive as well?

bstien: I see the ones for `bodyTextColor` and `detailTextColor` list all cases

robinsalehjan: The title stays the same regardless of state (atleast for now) 😄

robinsalehjan: The only special case is if it's `notStarted`

PR #821, opened 2020-03-05
Tweak constrains and load icon for TransactionView

robinsalehjan:
# Why?

- The previous constraints allowed the subviews of `UIStackView` to grow larger than the intristic content size. This did not look any good on larger devices such as iPads.

# What?

- Tweak constraints
- Load the icon image in some steps

# Show me

| iPhone | iPad |
| --- | --- |
| iphone | Ipad |


Sources/Fullscreen/TransactionView/TransactionView.swift
```diff
@@ -178,6 +183,27 @@ private extension TransactionView {
...
+        guard
+            let firstStepDot = stepDots.first,
+            let lastTransactionStepView = verticalStackView.arrangedSubviews.last
+        else { return }
+
+        verticalStackViewLeadingAnchor = verticalStackView.leadingAnchor.constraint(equalTo: firstStepDot.trailingAnchor, constant: .spacingS)
+        verticalStackViewBottomAnchor = verticalStackView.bottomAnchor.constraint(equalTo: lastTransactionStepView.bottomAnchor)
+
+        NSLayoutConstraint.activate([
+            verticalStackViewLeadingAnchor!,
```
ag12: is `!` necessary?

robinsalehjan: The value is set two lines above, I don't see a need to guard it?

PR #827, opened 2020-03-06
Feature: Error View

fespinoza:
# Why?

We showed errors in the object page using a view written in obj-c that lived
inside the iOS app.

# What?

- Add fullscreen component ErrorView
- Add layout helpers: `centerInSuperview` and `centerAndConstraintInSuperview`

# Show me

Screenshot 2020-03-06 at 11 56 37


Sources/Extensions/UIKit/Layout.swift
```diff
@@ -44,15 +44,49 @@ public extension UIView {
...
-        let constrains: [NSLayoutConstraint] = [
+        let constraints: [NSLayoutConstraint] = [
             topAnchor.constraint(equalTo: superview.layoutMarginsGuide.topAnchor),
             leadingAnchor.constraint(equalTo: superview.layoutMarginsGuide.leadingAnchor),
             trailingAnchor.constraint(equalTo: superview.layoutMarginsGuide.trailingAnchor),
             bottomAnchor.constraint(equalTo: superview.layoutMarginsGuide.bottomAnchor),
         ]
 
-        NSLayoutConstraint.activate(constrains)
+        NSLayoutConstraint.activate(constraints)
```
ag12: why do we activate and also returning the cons?

Sources/Extensions/UIKit/Layout.swift
```diff
@@ -44,15 +44,49 @@ public extension UIView {
...
 
-        return constrains
+        return constraints
+    }
+
+    @discardableResult
+    func centerInSuperview() -> [NSLayoutConstraint] {
```
osanoj: Maybe add `isActive: Bool = true` parameter like `fillInSuperview`?

Sources/Extensions/UIKit/Layout.swift
fespinoza: I guess it can be useful to keep a reference to them if you want to replace them after. I think it's ok to have the option to do this

Sources/Extensions/UIKit/Layout.swift
fespinoza: unless we need it, i would prefer not to. Specially since we can add this if needed after without breaking the API (with default arguments)

osanoj: I would say it would have been useful inside `centerAndConstraintInSuperview` 😄

Sources/Fullscreen/ErrorView/ErrorView.swift
```diff
@@ -0,0 +1,80 @@
...
+
+    public func configure(
+        title: String,
+        titleColor: UIColor = .textPrimary,
+        description: String? = nil,
+        icon: UIImage? = nil
+    ) {
+        titleLabel.text = title
+        titleLabel.textColor = titleColor
+        descriptionLabel.text = description
```
osanoj: Hide `descriptionLabel` if `description == nil`?

PR #840, opened 2020-03-19
TransactionStepView: Add external icon if action is not supported natively

robinsalehjan:
# Why?

- Would like to let the users know that by tapping the button they will be sent out to the browser.

# What?

- Add the `webView` icon to a button given that the `Action` of the button is not supported natively. As of now the only native action type is `seeAd`

# Show me

| Before | After |
| --- | --- |
| Simulator Screen Shot - iPhone 11 Pro - 2020-03-19 at 07 28 00 | 11-pro-max |


Sources/Fullscreen/TransactionView/TransactionStepView/TransactionStepView.swift
```diff
@@ -212,25 +215,67 @@ private extension TransactionStepView {
...
+    func addWebViewIconToButton(_ button: Button) {
+        let imageView = UIImageView(image: UIImage(named: .webview))
+        imageView.translatesAutoresizingMaskIntoConstraints = false
+
+        let imageWidth: CGFloat = 16
+        var leadingInset = CGFloat.zero
+
+        if model.state == .completed {
+            button.contentHorizontalAlignment = .leading
+            leadingInset = -10
```
bstien: Why this inset if its state is `.completed`? 😅

robinsalehjan: If the state of the step is `.notStarted` or `.completed` all the labels and button labels should be vertically aligned with the `leadingAnchor` of the title. If the state of the step is `.active` then everything should be indented inwards toward the centre.

This button has the additional icon along with custom leading and trailing insets to centre in correctly. If I don't provide it with a negative inset the `titleLabel` in the button looks a few points off compared to the first step 😄

bstien: Ok, thanks for the explanation :) My cabin fevered mind couldn't figure it out why this was needed 😅

PR #841, opened 2020-03-19
KeyValueGridView: Make value label copyable and center UIMenuController

bstien:
# Why?
The `UIMenuController` presented when copying `Label`s were centered on the label, not the text itself. If the text is shorter than the label, this looks weird.

And to make the value label within `KeyValueGridView` copyable.

# Show me
| Before | After |
| --- | --- |
| Screenshot 2020-03-19 at 10 21 00 | Screenshot 2020-03-19 at 10 20 09 |


bstien:
Found another thing to fix, so the basis for this PR has changed :)

PR #843, opened 2020-03-25
A-case: Create AdManagementTransactionProcessCell

robinsalehjan:
# Why?

- We would like to have an entry point in `AdManagement`

# What?

- Implement `AdManagementTransactionProcessCell`
- Use the `TransactionProcessView` to fill the content of the cell.

# Show me

| Before | After |
| --- | --- |
| before | after |


Sources/Recycling/ListViews/AdManagement/Cells/UserAdManagementTransactionProcessCell.swift
```diff
@@ -0,0 +1,42 @@
+//
+//  Copyright © 2020 FINN AS. All rights reserved.
+//
+
+public class UserAdManagementTransactionProcessCell: UITableViewCell {
+    // MARK: - Public
+
+    private(set) public var model: TransactionProcessViewModel?
```
hakloev: Does the `model` need to be public?

robinsalehjan: Not right now, but that may change in the future. I can keep it private until it's actually needed 👍

PR #844, opened 2020-03-25
TransactionProcess: Change WarningView to AlertView

robinsalehjan:
# Why?

- We updated the API model on the backend and would need some changes here and in the host app.

```
The new model

alert = {
    title: String
    body: String (previously text: String)
    imageIdentifier: String (previously imagePath: String)
}
```

# What?

- Rename from `TransactionWarningView` to `TransactionAlertView` for consistency
- Update `AlertViewModel` properties according to the new model
- Update all conforming demo `Model`'s to reflect the changes
- Add the `multipleContracts` asset to have a mapping for the `alert-multiple-contracts` identifier

# Show me

| known identifier | placeholder identifier |
| --- | --- |
| Simulator Screen Shot - iPhone 11 Pro - 2020-03-25 at 12 38 28 | Simulator Screen Shot - iPhone 11 Pro - 2020-03-25 at 12 38 26 |


robinsalehjan:
This requires some minor changes in the host app, so will follow up with a PR there before I merge this.

PR #846, opened 2020-03-26
Create FinnUI project and restructure FinniversKit files

osanoj:
# Why?
Because we want to split FinniversKit.

# What?
This creates the new FinnUI project and a FinniversKit workspace to contain all the projects.

It also, and this is the biggest change in this PR, moves the FinniversKit files into a new FinniversKit folder. 

Because of the moving of files this PR will be quite breaking on any work in progress branches. That is why I'm making it now before everything is complete. I don't think it will affect anything else than WIP branches... 😄


osanoj:
Ok, this should be OK to merge now, basically it just moves the files and should not really change anything else. Hopefully 😛

bstien:
Behind the scenes look at me reviewing this PR

Kapture 2020-03-26 at 14 55 43

PR #848, opened 2020-03-26
FinnUI - part deux

osanoj:
# Why?
Because the first PR was just a step on the way.

# What?
- Adds podspec for FinnUI.
- Moves SplashView to FinnUI.
- Creates a new project for the FinniversKit demo, this can be used to demo both FinniversKit and FinnUI. It might be that we should separate it, but this was the easiest solution for now.
- Moved all the FinniversKit/Demo files to Demo/Demo. (So @bstien gets to do some more PR-scrolling 😂)

# Show me

## Demo runs
Simulator Screen Shot - iPhone 8 - 2020-03-26 at 15 35 30

## Splash view (start)
Simulator Screen Shot - iPhone 8 - 2020-03-26 at 15 37 57


osanoj:
Ok, I think this is ready to review now, not much actual code to review, mainly a lot of moved files and project configuration. I will make a PR in main app, to use splash view from FinnUI instead.

osanoj:
What I did to separate between FinniversKit and FinnUI releases is that the FinnUI release tag should be prefixed with `finnui-`, e.g. `finnui-1.0.0`. 

If you have any other suggestions about this please let me know.

FinnUI.podspec
```diff
@@ -0,0 +1,39 @@
+Pod::Spec.new do |s|
+  s.name         = 'FinnUI'
+  s.version      = '1.0.0'
```
bstien: > What I did to separate between FinniversKit and FinnUI releases is that the FinnUI release tag should be prefixed with finnui-, e.g. finnui-1.0.0.

Shouldn't this be equal to the tag we're checking out, i.e. `finnui-1.0.0`? Or doesn't Cocoapods look at this string?

bstien:
> What I did to separate between FinniversKit and FinnUI releases is that the FinnUI release tag should be prefixed with `finnui-`, e.g. `finnui-1.0.0`.

And I guess this also means we have to create two new releases if one of our PRs add changes to both targets? Guess it's time to create some automation for this 😅

FinnUI.podspec
osanoj: No, as far as I understand it is the actual version number of the library, but it is used to specify tag a little later on:
`s.source        = { :git => "https://github.com/finn-no/FinniversKit.git", :tag => "finnui-#{s.version}" }`

bstien: Ah, I see. Ok, let's try it out 👍

osanoj:
> > What I did to separate between FinniversKit and FinnUI releases is that the FinnUI release tag should be prefixed with `finnui-`, e.g. `finnui-1.0.0`.
> 
> And I guess this also means we have to create two new releases if one of our PRs add changes to both targets? Guess it's time to create some automation for creating releases 😅

Yes, we need to create some automation for this, anyone volunteering? 😂

FinnUI.podspec
```diff
@@ -0,0 +1,39 @@
...
+  s.platform      = :ios, '11.2'
+  s.swift_version = '5.0'
+  s.source        = { :git => "https://github.com/finn-no/FinniversKit.git", :tag => "finnui-#{s.version}" }
+  s.requires_arc  = true
+
+  s.source_files = 'FinnUI/Sources/*.{h,m,swift}', 'FinnUI/Sources/**/*.{h,m,swift}', 'FinnUI/Sources/**/**/*.{h,m,swift}'
+  s.resources    = 'FinnUI/Sources/Resources/Fonts/*.ttf', 'FinnUI/Sources/Resources/*.xcassets', 'FinnUI/Sources/Resources/Sounds/*.{mp3,wav,sf2}'
+  s.resource_bundles = {
+      'FinnUI' => ['FinnUI/Sources/Resources/*.xcassets', 'FinnUI/Sources/Resources/Fonts/*.ttf', 'FinnUI/Sources/Resources/Sounds/*.{mp3,wav,sf2}']
+  }
```
bstien: Noticed this now. I guess these can be removed, since `FinnUI` probably shouldn't contain resources?

osanoj: It will need to contain resource, images and things, they should not be public though, but this is needed to build the framework. 

At the moment FinniversKit has them public, but that should be changed. 

For instance in this PR the splash view actually uses a FinniversKit image, but that should also be changed later.

PR #850, opened 2020-03-30
Feature: corona help popup

fespinoza:
# Why?

We want to promote the corona help section of our search, also give a shout out to people
working on the crisis

# What?

- Add component CoronaHelpView
- Add UIView.resetDropShadow
- Add `DemoWrapperView` to easily use Xcode previews with our current UIKit demo views
- Fix some Xcode / SwiftLint warnings on existing code

The CoronaHelpDemoView uses a private method `setupOnHorizontalSizeClass` which I use to define certain constraints/styles depending on regular/compact horizontal size class (the correct version of isIPad)

# Show me

| Variant | |
|:---:|:---:|
| **iPhone** |  | 
| **iPad** | ipad | 
| **iPhone Dark** | dark | 
| **iPhone SE** | se |


fespinoza:
The only thing left is to make sure the test pass with the latest snapshot library (required by the latest Xcode) and potentially move this components to FinnUI, what do you think @osanoj ?

fespinoza:
The code is ready to review @finn-no/ios, there may be still a failure with the test as I need to use the same simulator version than circle ci uses, there are some changes apparently in 13.4

osanoj:
> The only thing left is to make sure the test pass with the latest snapshot library (required by the latest Xcode) and potentially move this components to FinnUI, what do you think @osanoj ?

Yes, this should be in FinnUI later, but since there is still some things to fix in that project we can move things later on. 👍

PR #851, opened 2020-03-30
Support note for each viewing

ninarg:
# Why?

Last time we did a redesign of the viewings on the object page, it was decided to remove the individual note per viewing. The realestate team changed their mind about this now, since many brokers use this note field to write useful information, e.g. about streaming viewings.

# What?

Added a new note label inside `ViewingCell`. Cell height is now dynamic, depending on note length. Also added support for a top and bottom margin in the cells.

# Show me

IMG_1502


FinniversKit/Sources/Components/Viewings/ViewingCell.swift
```diff
@@ -95,22 +114,48 @@ class ViewingCell: UITableViewCell {
...
         }
-        weekdayLabel.text = viewModel.weekday
-        monthLabel.text = viewModel.month
-        dayLabel.text = viewModel.day
-        addToCalendarButton.setTitle(addToCalendarButtonTitle, for: .normal)
+
+        var bottomMargin: CGFloat = 0
+        if let note = viewModel.note {
+            noteLabel.text = note
+            bottomMargin = noteBottomMargin
```
bstien: We should add this, just in case the cells gets re-used:

```swift
noteLabel.isHidden = false
```

PR #854, opened 2020-04-02
FinnUI should support assets the same way as FinniversKit

osanoj:
# Why?
FinnUI should be easy to use and have all the setup we are used to.

# What?
- Improved the resource symbol generation script so the same can be used for both projects. 
- Removed the objc resource symbols. Because objc.
- Deleted remaining swiftformat things.
- Tweaked build step order.
- Moved splash view images to FinnUI.
- Updated splash view snapshot. There should not be any difference and I can't see any difference, but there is apparently a difference...


bstien:
What I don't understand with these changes is why the splashView snapshot reference image would be modified 😅 Its assets were moved, yes, but it's the same ones being used and the image looks exactly the same.

osanoj:
> What I don't understand with these changes is why the splashView snapshot reference image would be modified 😅 Its assets were moved, yes, but it's the same ones being used and the image looks exactly the same.

Yup, all I know is the new one is bigger, `+47 Bytes` 😛

PR #857, opened 2020-04-14
Internal: ensure button styles conform to the abstract spec

fespinoza:
# Why?

Working on the link buttons within the object page, with Thuy we realize that the link buttons's font for the normal size was caption (14pt), while the size defined in abstract was body (16pt) as it can be seen in [Buttons in Abstract](https://app.abstract.com/share/86a12fb5-1b04-4e2a-9aac-6e615ee2d38c?collectionId=4707317b-ac14-4cc4-9dd8-25dc5b466fa8&collectionLayerId=bc630990-13e7-4591-9f5e-918141c3161e&mode=build&selected=1488545417-1A13F18B-09C1-42CC-8CF7-1BC73EE7ACA1&sha=14fab54af4eb12455d704421751367aedece3448).

The font for the `small` size for links is unchanged.

The `link` and `flat` buttons differ mostly on font weight now.

Some of the existing components look slightly weird with this change, please see the file difference where you can see how the components change. I'm unsure if we want to accept this change for the components or replace some of the links as `small` (12pt). What do you think Thuy?

# What?

- Set the link button font to `body` for the normal size
- Update snapshots of components using link buttons
- Revamp demo view for buttons, to look more like the abstract spec
- Fix issue with `destructive flat button` where the font for the normal size was wrong
- Make ad management view, statistics button small (as it should have been)
- Update the default button's border color

# Show me

| Before | After |
| --- | --- |
| before  | after |


osanoj:
In general I guess this type of link should be the same size as surrounding text. Which is usually body, so that makes sense. 

Some things I'm not sure of:
- Safety elements: These links maybe should be a bit smaller since it is something that will be seen lots of times, but probably is very seldom someone wants to open them. They need to be there but is not very useful, sort of.

- Infobox: The text is smaller than body size, so maybe the link should be similar. But probably still fine with bigger link button in this one.

- LoginView: Here I'm slightly worried the label to left and right will have very little spacing in the middle on a small screen (iPhone SE). But I also notice that the "Ny bruker" and "Glemt passord" are not aligned either, was the same before this change though 😛

fespinoza:
Nice @osanoj. I will go through the things that change with Thuy, to avoid this issues, which derived from wrong initial styles.

I am adding other changes to this PR as I found other issues related to buttons, I will mark this PR as WIP until I'm done checking with Thuy

fespinoza:
@finn-no/iOS, I added a couple more changes to this PR, it's everything updated in the description. It would be nice if you guys re-review.

I went through the changes with Thuy. We talked that we may add back a 14pt link in the future, but we need to think that a bit more thoroughly, since we have 12pt and 16pt right now and both sizes are used around.

Then the infobox component is rendered as it should, the same with the safety elements.

The LoginView I understand it's not really used anywhere, so we should remove this from FinniversKit (I can do it in a new PR after).

Then this is ready for the final review

PR #858, opened 2020-04-20
Demo should be runnable from device

osanoj:
# Why?
The demo app didn't run on device.

# What?
Fixed so FinnUI and FinniversKit frameworks gets copied correctly (to Frameworks not Products).

### DIFF
--- a/Demo/Demo.xcodeproj/project.pbxproj
+++ b/Demo/Demo.xcodeproj/project.pbxproj
@@ -243,7 +243,7 @@
 			isa = PBXCopyFilesBuildPhase;
 			buildActionMask = 2147483647;
 			dstPath = "";
-			dstSubfolderSpec = 16;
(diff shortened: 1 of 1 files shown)


Demo/Demo.xcodeproj/project.pbxproj
```diff
@@ -243,7 +243,7 @@
 			isa = PBXCopyFilesBuildPhase;
 			buildActionMask = 2147483647;
 			dstPath = "";
-			dstSubfolderSpec = 16;
+			dstSubfolderSpec = 10;
```
fespinoza, line 246: mmm.. why not 11?

osanoj: 11? That is just crazy! 🙃😂 

I was thinking of 8 at first, but 10 worked much better! 😜

PR #862, opened 2020-04-27
Update TransactionProcessSummaryView with external view

robinsalehjan:
# Why?

- Need to support the `style` attribute provided by the backend (in case of errors in a step)
- Need a view to show the user `Mine kjøretøy`

# What?

- Refactor the `TransactionProcessSummaryView` to support the style property, as well as having a external view for `Mine kjøretøy`
# Show me
| Before | After |
| --- | --- |
| Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-27 at 17 17 36 | Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-27 at 16 56 01 |


robinsalehjan:
The error icon is not part of the design. I just used it for now, until I get the actual asset.

PR #863, opened 2020-04-28
Replace the errorCircleFilledMini

robinsalehjan:
# Why?

- We are supposed to use the `exclamationMarkTriangleMini` in the `TransactionProcess`, but I noticed that we did not have that asset. I asked Ingrid and she thought it was the one used in the `ToastView`. Turns out that the `errorCircleFilledMini ` was supposed to be replaced, but it hasn't been done yet.

# What?

- Replace `errorCircleFilledMini` with `exclamationMarkTriangleMini`
- Remove references to `errorCircleFilledMini`
- Update the `toastView` snapshot

# Show me
| Before | After |
| --- | --- |
| before | after |


ag12:
👌

PR #864, opened 2020-04-28
Fix: Add color lime + change ChatAvailabilityView color

bstien:
# Why?
We have a task to add the color `lime`, `#007200`, to FinniversKit. This color should also be the color indicator within `ChatAvailabilityView` when the dealership is online.

# What?
- Added color `lime`.
- Update color for `ChatAvailabilityView`.
- Add `lime` to color demo.

# Show me

| Before | After |
| --- | --- |
| Screenshot 2020-04-03 at 12 23 09 | Screenshot 2020-04-28 at 11 26 29 |


bstien:
Good call, forgot about those 😄

PR #865, opened 2020-04-28
Adin 8096 external button in ad confirmation

nickfalk:
# Why?

Because AdIn needs to get feedback from the customers.

# What?

Support for an explanatory text and a button to take the users to an external Questback-form in the AdConfirmation-page.

# Show me
| Before | After |
| --- | --- |
| before |after|


FinniversKit/Sources/Fullscreen/AdConfirmationView/Views/AdConfirmationLinkView.swift
```diff
@@ -0,0 +1,129 @@
...
+            linkIcon.centerYAnchor.constraint(equalTo: linkButton.centerYAnchor)
+        ])
+
+        addSubview(linkButton)
+        NSLayoutConstraint.activate([
+            linkButton.centerXAnchor.constraint(equalTo: centerXAnchor),
+            bottomAnchor.constraint(equalTo: linkButton.bottomAnchor)
+        ])
+
+        let maxWidth: CGFloat = 640 - .spacingXL
```
robinsalehjan, line 86: Why this specific width :P?

nickfalk: 👍  I'll get rid of it. 😄 It was mostly an idea to limit the possibility of extremely wide texts on iPad in landscape, but it's not really a likely scenario for this element at all.

PR #867, opened 2020-04-30
SwiftUI: implement UI for settings view

vadymmarkov:
# Why?

In order to experiment and gradually re-write our UI components in SwiftUI for iOS13+ devices.

# What?

- Add `BasicListCell` with the same layout and subviews as in `BasicTableViewCell`. It's possible to customise text labels by passing `@ViewBuilder` closures to the init method.

- Add `ListSeparatorStyleNoneModifier` to hide default `UITableView` separators. Haven't found a better way of solving this.

- Implement UI for `SettingsView`. State management, action closures and other logic will be added in a separate PR. Initially I started with built-in `Section` component, but ended up implementing section headers and footers as regular cells, because I couldn't get rid of default spacings between sections. It looks like `List` API is not as open for customisations as `UITableView`.

For now both SwiftUI "foundation" and the settings view reside in FinnUI project under "Fullscreen/Settings", but we should of course find a better place for them at some point.

Also, if we want to have snapshot tests in addition to previews, where should we place them?

# Show me
| Basic list cell | Settings view |
| --- | --- |
| Screenshot 2020-04-30 at 18 32 29 | Screenshot 2020-04-30 at 18 32 47 |


fespinoza:
It looks good. Finally some swiftUI :D! 🎉 

> Also, if we want to have snapshot tests in addition to previews, where should we place them?

I'm not sure how we handle the snapshot tests for FinnUI, but I would imagine to place them in unit tests (I imagine it's possible to reuse the preview code for snapshots, I guess we can try to figure out that).


I took this branch and I was playing a bit with it. One thing we need to solve is dark mode colors. Using `Color(.uiColorConstant)` doesn't seem to work with dark mode or changing to dark mode. I have the same issue in my current work

vadymmarkov:
@fespinoza As we discussed on Slack, dynamic colours work well if `FinniversKit.userInterfaceStyleSupport` is set to `dynamic`. We should perhaps make it to be the default on iOS13+.

I've also added new `SwiftUIViewTests` class for snapshot testing of our SwiftUI components. Let me know what you think.

vadymmarkov:
Any idea why the tests for the views I didn't touched are failing?

fespinoza:
> Any idea why the tests for the views I didn't touch are failing?

Not really, it seems are only the table related ones that fail. Maybe there is a clash in file names?, I'm not sure. Something went wrong within the last 3 commits, the test passed before

vadymmarkov:
@fespinoza This is ready for another review. The tests should be fixed now. `UITableView.appearance().separatorStyle = .none` affected some of the table views in FinniversKit, so after hours of trying different workarounds I came up with [this solution](https://github.com/finn-no/FinniversKit/pull/867/files#diff-a26dda7d2e779903ea230b5322cf5b0aR16), which seems to work. Still it's weird we have to do hacks like this to customise the list view.

FinnUI/Sources/Fullscreen/Settings/SettingsView.swift
```diff
@@ -0,0 +1,181 @@
...
+                if self.sections[section].title != nil {
+                    Header(text: self.sections[section].title!)
+                }
+
+                ForEach(0..<self.sections[section].items.count) { row in
+                    self.cell(at: row, in: section)
+                        .bottomDivider(row != self.sections[section].items.count - 1)
+                }
+
+                if self.sections[section].footerTitle != nil {
```
fespinoza, line 28: maybe you can use `map` here?

FinnUI/Sources/Fullscreen/Settings/SettingsView.swift
```diff
@@ -0,0 +1,181 @@
...
+    public var body: some View {
+        List {
+            ForEach(0..<sections.count) { section in
+                if self.sections[section].title != nil {
+                    Header(text: self.sections[section].title!)
+                }
+
+                ForEach(0..<self.sections[section].items.count) { row in
+                    self.cell(at: row, in: section)
+                        .bottomDivider(row != self.sections[section].items.count - 1)
```
fespinoza, line 25: maybe it would be nice to have a function to express you want this for the last item (for readability)

FinnUI/Sources/Fullscreen/Settings/SettingsView.swift
```diff
@@ -0,0 +1,181 @@
...
+                .font(Font(UIFont.body))
+                .foregroundColor(.textSecondary)
+        })
+    }
+}
+
+@available(iOS 13.0.0, *)
+private struct ToggleCell: View {
+    let model: SettingsViewToggleCellModel
+    @SwiftUI.State private var isOn = true
```
fespinoza, line 109: Why the SwiftUI prefix? is there a type that crashes?

FinnUI/Sources/Fullscreen/Settings/SettingsView.swift
vadymmarkov: Good idea!

FinnUI/Sources/Fullscreen/Settings/SettingsView.swift
vadymmarkov: I don't know. For some reason I'm getting "Unknown attribute 'State'" compiler error if `SwiftUI` module name is not specified. Can't find any conflicting types in Finnivers though.

vadymmarkov:
As for the demo, I can add new section for SwiftUI views, similar to what I've done for snapshot tests. But probably in a new PR.

PR #870, opened 2020-05-05
Dark mode fix in primings

ninarg:
# Why?

You can barely see the difference, but I noticed `PrimingTableViewCell` background color was a little different from the bottom sheet background. 

# What?

Use `.bgPrimary` on the cell, which is the background color used in the bottom sheet. 

# Show me

| Before | After |
| --- | --- |
| IMG_1748 | IMG_1749 |


fespinoza:
image

PR #871, opened 2020-05-06
SwiftUI: prepare settings view for usage from UIKit

vadymmarkov:
# Why?

In order to experiment and gradually re-write our UI components in SwiftUI for iOS13+ devices.

# What?

- Add version label to the bottom of the settings view
- Fix section header and footer on iPad
- Implement basic list cell as a button
- Add `onToggle` and `onSelect` closures as a part of public API
- Add SwiftUI demo views as a separate section in the Demo app

All my attempts to fix the "iPad problem" without using UIKit failed. Popovers apparently don't work in nested `ForEach` components and there is no way to access underlying uiviews or their frames. 

Therefore I wrapped every cell in `UIViewControllerRepresentable` view, which creates `UIHostingController`. This allows us to return actual UIKit view back to the app when the cell is selected. Not super happy about this solution, but at least it will minimise changes we have to do in the FINN app.

# Show me
| Light mode | Dark mode |
| --- | --- |
| Simulator Screen Shot - iPhone 11 Pro - 2020-05-06 at 12 54 51 | Simulator Screen Shot - iPhone 11 Pro - 2020-05-06 at 12 54 57 |


FinnUI/Sources/Fullscreen/Settings/Foundation/Views/BasicListCell.swift
```diff
@@ -7,24 +7,33 @@ import FinniversKit
...
+        self.action = action
         self.title = title(model.title)
         self.subtitle = subtitle(model.subtitle)
         self.detailText = detailText(model.detailText)
     }
 
     public var body: some View {
+        Button(action: { self.action?() }) {
+            content
+        }.buttonStyle(BasicButtonStyle())
```
fespinoza, line 33: It's possible to write this like

```swift
Button(content, action: { self.action?() })
  .buttonStyle(BasicButtonStyle())
```

just as a style option

FinnUI/Sources/Fullscreen/Settings/Foundation/Views/BasicListCell.swift
```diff
@@ -79,6 +86,27 @@ public struct BasicListCell: View {
     }
 }
 
+@available(iOS 13.0, *)
+private struct BasicButtonStyle: ButtonStyle {
+    @Environment(\.colorScheme) var colorScheme: ColorScheme
```
fespinoza, line 91: I guess this is not needed

FinnUI/Sources/Fullscreen/Settings/Foundation/Views/BasicListCell.swift
vadymmarkov: I think it's a different initialiser that requires `StringProtocol`, while `content` here is `some View`.

fespinoza: right :P

PR #880, opened 2020-05-12
Design of notification center v2

shredlocker:
# Why?

Notification Center has new designs and added functionality. This PR implements the design changes.

# What?

### NotificationCenterView
- Split `NotificationCenterView` into segments. Data source will ask for number of segments and their title.
- A `SegmentContainer` contains the content of a segment, which is the `title`, `tableView` and the leading constraint of the tableView. The constraint is used for transition between segments.
- Added `NotificationCenterCellType` enum which contains supported cells.
- Show maximum 5 notifications pr saved search pr day, if there are more notifications we add a gradient to the last cell for that search and add a footer view with button to show more notifications.

### Cells
- The `NotificationCell` can display two types of content: `PersonalNotificationCellContent` and `SavedSearchNotificationCellContent`.
- Added `FeedbackCell`

### NotificationCenterSearchView
- Added this view as a middle step between `NotificationCenterView` and the search results. This view will show up to 20 notifications from a specific search within a single day.

# Show me

### NotificationCenterView

| Before | After 1 | After 2 |
| --- | --- | --- |
| Simulator Screen Shot - iPhone 11 Pro - 2020-05-12 at 10 16 49 | Simulator Screen Shot - iPhone 11 Pro - 2020-05-12 at 10 10 02 | Simulator Screen Shot - iPhone 11 Pro - 2020-05-12 at 10 11 23 |

### NotificationCenterSearchView

Simulator Screen Shot - iPhone 11 Pro - 2020-05-12 at 10 24 22


FinnUI/FinnUI.xcodeproj/project.pbxproj
```diff
@@ -98,6 +118,7 @@
 		46AE6FAE242BAAAE007BB005 /* Fullscreen */ = {
 			isa = PBXGroup;
 			children = (
+				DED0A07E24685F120016176A /* NoificationCenter */,
```
osanoj, line 121: No**t**ificationCenter maybe 😄

shredlocker: Yea, maybe 🤔

FinnUI/Sources/Fullscreen/NotificationCenter/Views/NotificationCenterView.swift
```diff
@@ -0,0 +1,304 @@
...
+            cell.remoteImageViewDataSource = remoteImageViewDataSource
+            cell.configure(with: model, timestamp: timestamp, hideSeparator: isLast, showGradient: isLast && overflow)
+            return cell
+        case let .feedbackCell(delegate, state, model):
+            let cell = tableView.dequeue(FeedbackCell.self, for: indexPath)
+            cell.delegate = delegate
+            cell.configure(for: state, with: model)
+            return cell
+        case .none:
+            fatalError("Cell type not specified")
```
osanoj, line 126: It would have been nice with an error log for this case. But you don't have to do it in this task since we do not have that infrastructure in place. But perhaps we should add a task to add it

shredlocker: Sure, that would be good to have 👍

FinnUI/Sources/Fullscreen/NotificationCenter/Views/NotificationCenterView.swift
```diff
@@ -0,0 +1,305 @@
...
+    func setupSegmentedControl() {
+        segmentContainers = []
+        
+        guard let dataSource = dataSource else {
+            return
+        }
+        
+        for segment in 0 ..< dataSource.numberOfSegments(in: self) {
+            let title = dataSource.notificationCenterView(self, titleInSegment: segment)
+            let tableView = UITableView.notificationCenterTableView()
```
osanoj, line 202: You could perhaps rename this so it is clearer that it creates a new table view, I was a bit unsure first if it was some singleton thing 😛 

`createNotificationCenterTableView()` or something similar would make it more clear.

FinnUI/Sources/Fullscreen/NotificationCenter/Views/PersonalNotificationIconView.swift
```diff
@@ -0,0 +1,60 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    func configure(with icon: Kind?) {
+        switch icon {
+        case .favorite:
+            backgroundColor = .secondaryBlue
+            imageView.image = UIImage(named: .heartMini)
+        case .myAds:
+            backgroundColor = .primaryBlue
```
osanoj, line 49: Is it on purpose that the dark mode compatible colors are not used in this view?

shredlocker: Yes, this view represents the icon as a whole and the icons have the same background regardless of dark mode 😉 They use some new colors as well

osanoj: Ok, so darkmode-compatible without changing colors then I suppose 🤔

FinniversKit/Sources/Extensions/UIKit/UITableViewExtensions.swift
```diff
@@ -30,4 +30,8 @@ public extension UITableView {
         // swiftlint:disable:next force_cast
         return dequeueReusableHeaderFooterView(withIdentifier: headerFooterClass.reuseIdentifier) as! T
     }
+    
+    func stopScrolling() {
+        setContentOffset(contentOffset, animated: false)
```
osanoj, line 35: Interesting way to make the scrolling stop 👍

PR #883, opened 2020-05-13
Update TransactionProcessSummaryView model

robinsalehjan:
# Why?

- Last minute changes which we all love!
- The text and url to `Mine kjøretøy` will be provided by the backend

# What?

- Update ViewModel
- Add url to delegate method


osanoj:
Do you really need to include the url here in the view code? Since it is not used in the views? "Someone" creates the view model, giving it a url, then when the "external view" is tapped and delegate called shouldn't it then be possible to know which url belongs to the external view without receiving it from delegate? 

I might be wrong since I don't have that part of the code in front of me though 😄

robinsalehjan:
Yes, that is possible. Just did something quick for testing purposes.

PR #896, opened 2020-05-25
Feature: SwiftUI version of BAP ads

fespinoza:
# Why?

To start integration SwiftUI in the FINN app, but also do something about the way we render ads.

**Note**: Before merging this PR I will attempt to integrate with the app to refine the API to include whatever we may need for the integration.

The way of structuring the code and such is just an initial draft, so none of this ideas is fixed in stone as we will refine as we learn more about the swiftUI ways and the needs of the app.

# What?

- Add `BapAdView` and subviews
- Add View extension `finnFont`

# Architecture + Decisions

### BapAdView

The main properties for this view are:
  
  - BapAdViewModel: in this case is composed of static data (like the ad title, etc) and static text that will be used in the UI (labels and so on, for localization purposes)
  - `BapAdViewActions` is a protocol that defines all possible actions that require interaction with external views/view controllers. It contains actions that will be pass down to subcomponents like `TryHeltHjemView`
  - `GalleryImageProvider` and `AuthorImageProvider`, they contain injectable logic to fetch a collection of images or a single image and inform the view when the image is ready to be use
  
### Complex View Structure

- view definition
- extension containing sub views
- extension containing actions

Within the view itself I call subviews with a simple name, like `title` instead of `titleView` since at this top level everything is a view, since we are working in the scope of a SwiftUI's `ViewBuilder`.

Finally, the idea of these views, is to read the contents of them like:

```swift
public var body: some View {
     ScrollView {
        VStack(alignment: .center, spacing: .spacingS) {
            imageGallery

            Group {
                title
                price
                favoriteButton
                sendMessageButton
                phoneNumber
                author
                location
                description
                extras

                Group {
                    heltHjemPromotion
                    extraLinks
                    metadata
                }
            }
            .padding(.horizontal, .spacingS)
        }
    }
}
```

Which makes quite explicit which are the parts that create the bap ad view.

### Directory Structure

Given that there will be multiple object page implementations, I decided to create their own directory, `AdView`, which makes the directory structure look like this

```
FinnUI
  |- AdView
    |- Components
      |- AuthorView
      |- DescriptionView
      |- ImageGalleryView
      |- PhoneNumberView
      |- TryHeltHjemView
    |- BapAdView
      |- BapAdView.swift
  |- Components
    |- ChatAvailability
  |- Fullscreen
    |- Settings
```

In this case I put this `AuthorView` under `AdView -> Components` instead of the top level `Components` directory, as the contents of `AdView -> Components` is intended for specialized object page components that are not gonna be reused outside the object page.

This is the reason that `AuthorView` is an `internal` type instead of a `public` one. We can always makes this subviews `public` and move them into the top level `Components` directory if we need to.

### New Object Pages

When we create a new AdView SwiftUI View, let's say `JobAdView`, we start by copying the code of `BapAdView` and then modify the structure on which sub components do we need.

Any sub component that is not trivial (like a title), we should create a new subview.

# UI

Screenshot 2020-05-26 at 10 44 33


fespinoza:
**Note**: Xcode Previews don't work inside non-app targets, since FinnUI is a library target, then Xcode previews don't work there :(

fespinoza:
I rebased to avoid some xcodeproj file conflicts, since fixing them are quite tricky, so I just re-added the files instead. I will merge when the tests pass

PR #898, opened 2020-05-28
Feature: Favorite sold view

ninarg:
# Why?

Users receive mail updates about their favorite ads being sold. The mail links to a favorite sold page on web, but we want to have a native page for it. We're also gonna implement favorite sold notifications to our notification center, which will link to the same page. 

# What?

- Added `FavoriteSoldView` with demos. 
- Renamed `FrontPageRetryView` to `LoadingRetryView`, so it can be reused in favorite sold
- Use a separate `adsGridViewDataSource` in `FrontPageViewController`, since `RecommendationController` in `ios-app` will serve as `adsGridViewDataSource` for both `FrontPageViewController` and `FavoriteSoldViewController`

# Show me

favoritesold-demo


FinniversKit/Sources/Fullscreen/FavoriteSoldView/FavoriteSoldView.swift
```diff
@@ -0,0 +1,265 @@
...
+}
+
+public class FavoriteSoldView: UIView {
+    public var model: FavoriteSoldViewModel? {
+        didSet {
+            if let model = model {
+                self.configure(with: model)
+            }
+        }
+    }
```
shredlocker, line 29: Do you need to keep a reference to this model? Maybe it is enough to just call `configure(with:)`

FinniversKit/Sources/Fullscreen/FavoriteSoldView/FavoriteSoldView.swift
```diff
@@ -0,0 +1,265 @@
...
+            similarAdsTitleLabel.trailingAnchor.constraint(equalTo: headerView.trailingAnchor, constant: -FavoriteSoldView.margins),
+        ])
+
+        adsGridView.fillInSuperview()
+        adsGridView.headerView = headerView
+        setupFrames()
+    }
+
+    private func setupFrames() {
+        let stackViewWidth = frame.size.width - FavoriteSoldView.imageWidth - 3 * FavoriteSoldView.margins
```
shredlocker, line 187: Have you tried using `headerView.systemLayoutSizeFitting` instead? You can use that to calculate its size using its current constraints. Use it like this:
```
let targetSize = CGSize(width: frame.width, height: 0)
let size = headerView.systemLayoutSizeFitting(
    targetSize, 
    withHorizontalFittingPriority: .required,
    verticalFittingPriority: .defaultLow
)
```
That should in theory give you the smallest height given the width of the frame

FinniversKit/Sources/Fullscreen/FavoriteSoldView/FavoriteSoldView.swift
ninarg: You're right, changed it now! 👍

FinniversKit/Sources/Fullscreen/FavoriteSoldView/FavoriteSoldView.swift
ninarg: I tried now, it doesn't seem to work. I get height 0. Maybe that's because `headerView` is a `UICollectionView.elementKindSectionHeader` of a collection view. I'm not sure if it's supposed to work in that case?

shredlocker: I think the problem is that you haven't constrained the `similarAdsTitleLabel.bottomAnchor` to `headerView.bottomAnchor`.

ninarg: Yes, it worked! Thanks, that's much nicer 😄

PR #899, opened 2020-05-28
Show notification center in horizontal paging scroll view

shredlocker:
# Why?

This is how it was suppose to be like.

# What?

- Added a paging scroll view.
- Replaced `segmentContainers` with a simple array of UITableViews instead.
- Both table views will now layout its content at the same time, so find the index of the current table view in each data source method and some delegate methods.
- Added `UIScrollViewDelegate` and implemented `willEndDragging` and change `selectedSegment` based on the target content offset. 

# Show me

| Before | After |
| --- | --- |
| notification-center-no-pageing | notification-center-pageing |


FinnUI/Sources/Fullscreen/NotificationCenter/Views/NotificationCenterView.swift
```diff
@@ -67,42 +78,46 @@ final public class NotificationCenterView: UIView {
...
     public func numberOfSections(in tableView: UITableView) -> Int {
-        dataSource?.notificationCenterView(self, numberOfSectionsInSegment: selectedSegment) ?? 0
+        guard let index = tableViews?.firstIndex(of: tableView) else { return 0 }
+        return dataSource?.notificationCenterView(self, numberOfSectionsInSegment: index) ?? 0
     }
 
     public func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
-        dataSource?.notificationCenterView(self, segment: selectedSegment, numberOfRowsInSection: section) ?? 0
+        guard let index = tableViews?.firstIndex(of: tableView) else { return 0 }
+       return dataSource?.notificationCenterView(self, segment: index, numberOfRowsInSection: section) ?? 0
```
ninarg, line 112: Nitpick... a space is missing here. 😄 Everything else looks good to me! 💯

PR #903, opened 2020-06-04
Improvement: add footer to splashView

fespinoza:
Why?

to show that finn is a part of schibsted

- Trello: https://trello.com/c/Ih6jHC1G/875-we-need-to-add-en-del-av-schibsted-to-the-bottom-of-the-splash-screen

# What?

- add a logo for Schibsted in the splash view
- add a internal container to the finn logo so it centers between the Schibsted logo and the top of the screeen
- update snapshots

# Show me

| Before | After |
| --- | --- |
| before | after |


FinnUI/Sources/Fullscreen/SplashView/SplashView.swift
```diff
@@ -39,6 +39,14 @@ public final class SplashView: UIView {
         }
     }()
 
+    private lazy var logoContainer: UIView = UIView(withAutoLayout: true)
```
ninarg, line 42: Can just have `private lazy var logoContainer = UIView(withAutoLayout: true) `

PR #907, opened 2020-06-09
Add ConfettiView to FinnUI

hakloev:
# Why?

We want to show some confetti when an ad is marked as sold and maybe in other cases as well? 🎉

# What?

* Add `ConfettiView`. This view does not contain any visible views and is without user interaction enabled. Once `start(withDuration:completion)` is called, It adds a `CAEmitterLayer` sublayer that emits confetti for the given duration. 
  * `completion` is called when the emitter layer is no longer creating emitter cells, not when the last cell is off screen. 

# Show me

confetti_finnivers


FinnUI/Sources/Components/ConfettiView/ConfettiView.swift
```diff
@@ -0,0 +1,103 @@
...
+        cell.birthRate = 10
+        cell.lifetime = 5
+
+        cell.contents = confettiImages.randomElement()?.cgImage
+        cell.color = color.cgColor
+
+        cell.emissionLongitude = CGFloat.pi
+        cell.emissionRange = CGFloat.pi / 8
+
+        cell.velocity = UIDevice.current.userInterfaceIdiom == .pad ? 600 : 500
```
ninarg, line 92: You could use `UIDevice.isIPad()`. We try not to have the isIPad-check, but in this case it doesn't make sense to use `traitCollection.horizontalSizeClass` instead, since the vertical size is what matters here, I guess? 😄

fespinoza: there is a verticalSizeClass too :)

ninarg: Right, that could work, I haven't tested it before though 🤔

hakloev: Yes, I thought about it. It made sense to me to differentiate for both size classes (even though we don't support the landscape orientation), so that's why I just checked for `pad`.

hakloev: But in retrospect, maybe I should just have one velocity 🤔

osanoj: The 2 hardest things in programming:
1) Naming
2) Deciding which velocities to use for your confetti-rain

:trollface:

PR #908, opened 2020-06-11
Feature: ImageLinkListView

bstien:
# Why?
We have a task to redesign the links to videos and virtual viewings in the app.

This PR adds two new views: `ImageLinkView` and `ImageLinkListView`. The first is the single item view, while the latter is composed of a list of `ImageLinkView`s. I decided to create the list view so we don't have to take care of this in the app.

`ImageLinkListView` will be composed of groups containing 2 `ImageLinkView`s. The groups can be arranged horizontally or vertically. Vertically will be used for smaller devices, where the items would potentially be too small when presented in a horizontal grouping.

~I'm currently waiting for the icons used on the overlay to be made, but I guess I'll have them tomorrow. Will update the PR then.~

**Edit:** The new icons have been added.

# What?
- Add `ImageLinkView` and `ImageLinkListView`.
- Make one Array extension public.
- Create a shared/default implementation of `RemoteImageViewDataSource` to be used for all demos implementing this protocol.

# Show me
| Grouped horizontally | Grouped vertically |
| --- | --- |
| Screenshot 2020-06-15 at 10 07 50 | Screenshot 2020-06-15 at 10 07 56 |


bstien:
Just noticed my 'UNDO' commit😂 I've been jumping between branches, so I just forgot to actually undo that commit and give it a better message🤷‍♂️

FinnUI/Sources/Components/ImageLinkView/ImageLinkView.swift
```diff
@@ -0,0 +1,158 @@
...
+// MARK: - Private class
+
+private class OverlayView: UIView {
+
+    // MARK: - Private properties
+
+    private lazy var iconImageView: UIImageView = {
+        let imageView = UIImageView(withAutoLayout: true)
+        imageView.contentMode = .scaleAspectFit
+        imageView.tintColor = .milk
```
fespinoza, line 124: is this correct for both light and dark modes?

FinnUI/Sources/Components/ImageLinkView/ImageLinkView.swift
```diff
@@ -0,0 +1,158 @@
...
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    // MARK: - Setup
+
+    private func setup() {
+        layer.cornerRadius = .spacingS
+        backgroundColor = UIColor.black.withAlphaComponent(0.8)
```
fespinoza, line 143: light dark/mode?

FinnUI/Sources/Components/ImageLinkView/ImageLinkView.swift
bstien: Yes :) The images won't change between light- or dark-mode, so this is correct for both.

FinnUI/Sources/Components/ImageLinkView/ImageLinkView.swift
bstien: Same as above 👆

bstien:
I'll close this PR for now, ref. my comment in #911. We currently have a hard time retrieving thumbnails for both videos and virtual viewings, but we can use this if we manage to this at some point.

PR #915, opened 2020-06-17
Feature: Search results with recent and saved searches

ninarg:
# Why?

We are going to perform a search experiment for the front page free text search. We are providing a new view, with access to saved searches, and will investigate if this increases the use of saved searches. 

# What?

- Added `SearchResultsView` to `FinnUI`, that contains segments with `SearchResultsListView`, with subviews `SearchResultsRowView`
- Added new icon to `FinniversKit` called `arrowCounterClockwise`

# Show me

| Light mode | Dark mode |
| --- | --- |
| IMG_2044 | IMG_2045 |


Demo/Demo/Fullscreen/SearchResults/SearchResultsData.swift
```diff
@@ -0,0 +1,15 @@
+import FinnUI
+
+class RecentSearchData: SearchResultsListViewModel {
+    let title = "Siste søk"
+    let icon = UIImage(named: .arrowCounterClockwise)
```
osanoj, line 5: If the icon is used by the Demo project it should be a resource in there, the long term plan is to be able to remove the "public" modifier from `FinniversImageAsset`, now we have to little control over which image is used where

ninarg: Yes, it is used in the demo, but also in the ios-app. I will move this icon to the demo then, and also add the icon to ios-app.

PR #922, opened 2020-06-26
Feature: VotingView

bstien:
# Why?
We want to implement a voting view in the app, so we can get other types of feedback from our users.

We only have an initial design for this view currently, and we don't have the icons needed yet, so this is a WIP until then. Guess it'll be finished when the vacation is over and the designers have time to help us again.👨‍🎨🖼👩‍🎨

# What?
- Added `VotingView` and `VotingButtonView`, plus their corresponding view models.
- Added demo and snapshot tests.

# Show me

Kapture 2020-06-26 at 11 17 34

| Initial state | Left selected | Right selected |
| --- | --- | --- |
| Screenshot 2020-06-26 at 10 59 43 | Screenshot 2020-06-26 at 10 59 45 | Screenshot 2020-06-26 at 11 15 41 |


bstien:
Closing. We haven't had the need for user voting during the last year, no need to keep this open

PR #923, opened 2020-06-30
Save search view: Reduce spacings

ninarg:
# Why?

Is wasn't possible to dismiss the save search view presented as form sheet on iPad iOS 12. I proposed to add a navigation bar with a close button to fix it. If we want to go through with that solution, the spacings of this view needs to be a little reduced, since the navigation bar takes up some space. 

# What?

- Added max 2 lines for search name label, so that all buttons below will be visible on small devices
- Reduced some spacings
- Updated snapshot

# Show me

See [#2415](https://github.schibsted.io/finn/ios-app/pull/2415) for screenshots.


FinnUI/Sources/Components/SaveSearchView/SaveSearchView.swift
```diff
@@ -47,7 +47,7 @@ public class SaveSearchView: UIView {
     private lazy var searchNameLabel: Label = {
         let label = Label(style: .title3Strong, withAutoLayout: true)
         label.textAlignment = .center
-        label.numberOfLines = 0
+        label.numberOfLines = 2
```
bstien, line 50: I'm not sure if we should set this one🤔 It's up to the user if he/she wants a long name for the search, and I assume the owners of iPhone SEs are used to not seing all the content all of the time😄

ninarg: I kinda agree, it should be possible to see the full name somewhere 😅  Will remove it. Can't always do what's best for small screen users 😂

PR #924, opened 2020-07-01
Fix: SavedSearchView scrollview

bstien:
# Why?
The scrollview inserted into `SavedSearchView` wasn't scrollable, since someone™ forgot to set constraints on the contentView added to it🙈

# What?
- Call `contentView.fillInSuperview()`.
- Revert spacing changes from #923.
- Remove spacing between title and edit button.

# Show me
Kapture 2020-07-01 at 10 23 57


ninarg:
Great 😎  Maybe you can re-add the spacings I changed for in #923 again? Since it's not really needed anymore when we can scroll the view.  Then I don't have to revert. And maybe we also can reduce the spacing between the search name and "Endre navn på søket", like Thuy wanted? 😄

bstien:
You're right, I might as well add them here to avoid another PR😄👍

bstien:
@ninarg Ref. our talk with Thuy: I'm not able to remove all the spacing he wants. Since it's a button, it should keep its internal margins so it's still tappable. With the current style the height of the button is 31px

Screenshot 2020-07-01 at 10 59 57

ninarg:
Yeah, we don't want to reduce and make it difficult to tap. This is the closest we can have them 👍

PR #926, opened 2020-07-03
Add title and text labels to ChatAvailabilityView

ninarg:
# Why?

The button with a title was not sufficient for users to understand what this feature is. Therefore we want to provide a title and text to describe it better. 

# What?

Added `titleLabel` and `textLabel` to `ChatAvailabilityView`.

# Show me

| Before | After |
| --- | --- |
| IMG_2194 | IMG_2193 |


Demo/Demo/Components/ChatAvailability/ChatAvailabilityDemoView.swift
```diff
@@ -5,6 +5,12 @@
 import FinnUI
 import FinniversKit
 
+private struct ChatAvailabilityData: ChatAvailabilityViewModel {
```
robinsalehjan, line 8: Maybe move these strings into `Localizable.strings`

robinsalehjan: Never mind - my mind thought this was in the core app.

PR #928, opened 2020-07-08
Add shadow when scrolling to SaveSearchView

ninarg:
# Why?

Since other views presented in the bottom sheet has a shadow below the navigation bar when scrolling, we want to be consistent and do the same for `SaveSearchView`.

# What?

- Add scroll view below `topShadowView` if `usingShadowWhenScrolling`
- Make `SaveSearchView` into a `ShadowScrollView`, which handles the shadow for us

# Show me

| Before | After |
| --- | --- |
| before | after |


ag12:
if you need large text: https://www.lipsum.com/
💯

PR #929, opened 2020-07-14
Feature: SavedSearchSortingView

bstien:
# Why?
We need to implement sorting of saved searches, like we have for favorite ads.

Since these views are supposed to look the same, the view we currently use for favorite ads was a perfect candidate for refactoring into something reusable.

Within `SortSelectionView`, the models used must provide a `String` identifier. This will be used to store/check selection, as well as when communicating with its delegate. Most of the time, the sorting options will be limited and known beforehand, so this shouldn't be a problem.

~**NB!** We don't have the icons we need for this view, so I'll have to wait to merge this until we do. I reused the icons from `FavoriteAdSortingView` to have something to present🤓~

# What?
- Add generic view for sorting: `SortSelectionView`.
- Refactor `FavoriteAdSortingView` to use `SortSelectionView`.
- Add `SavedSearchSortingView` + models.

# Show me
Screenshot 2020-08-26 at 16 28 10


ninarg:
Could `SavedSearchSortingView` and `FavoriteAdSortingView` be moved to FinnUI (since they will not be reused elsewhere in the app), and `SortSelectionView` be kept in FinniversKit (since it’s a view we want to reuse multiple places)? Tell me if you disagree, not sure if I understand the FinnUI/FinniversKit distribution correctly. 😅

bstien:
Yes, they are not generic views, so I think they should go in FinnUI🙂 Will move them!📦

PR #930, opened 2020-07-15
Next phase of filters: Search filter tags & search display type selection menu

ninarg:
# Why?

We want to make filters even more user friendly in a next phase for filters! We are gonna provide search filter tags on top of the search results. The tags give you a good overview of what you're currently searching for, and gives you shortcut to filters in charcoal by tapping on a specific tag, and easy removal of a filter by pressing a delete button. As a part of the task, we also improve the navigation between search display types (which includes sorting and switching between list/grid/map), since the old menu for that in the app will be replaced by our new search filter tags.

# What?

This PR introduces 3 new views: 

- `SearchFilterTagsView`
- `SearchDisplayMenuView`: A floating menu in the bottom of search results, lets you change sorting or display type.
- `SearchDisplayTypeSelectionView`: Lets you switch between map/list/grid view for searches.

**SEARCH FILTER TAGS**
searchfiltertagd

**SEARCH DISPLAY MENU**
This view is pretty simple. It contains two buttons, to either sort or change display type.
Skjermbilde 2020-10-05 kl  13 21 13

**SEARCH DISPLAY TYPE SELECTION**
- Here I am reusing FinniversKit's`SortSelectionView` used multiple places in the app, but I renamed it to `SelectionView`, since it will now be used for a purpose that isn't sorting
Skjermbilde 2020-10-05 kl  13 19 48


bstien:
Wow, this is amazing!😍 Great job!👏👏👏

ninarg:
This PR is now ready for review! 😄

FinnUI/Sources/Components/SearchFilterTags/SearchFilterTagsView.swift
```diff
@@ -0,0 +1,255 @@
...
+        let searchFilterTagsDidChange = !searchFilterTagsAreEqual(self.searchFilterTags, searchFilterTags)
+        self.searchFilterTags = searchFilterTags
+
+        guard
+            searchFilterTagsDidChange,
+            reloadSection
+        else { return }
+
+        UIView.performWithoutAnimation {
+            collectionView.reloadSections([0])
```
ninarg, line 140: I had a bit of trouble right here. Ideally I would just do a `reloadData()`. But I discovered that `reloadData()` does not reload all visible cells (I expected `cellForItemAt` to be called for all visible cells). It’s an issue in the rare cases where two search filter tags switch order, but the count of search filter tags did not change. Then the cells would still contain their old label value, since cells were not configured with the new data in `cellForItemAt`.

However, 
`reloadSections[0]`
or
```
collectionView.reloadItems(at: collectionView.indexPathsForVisibleItems)
collectionView.reloadData()
```
works, since they reload all visible cells.

Does anyone know why `reloadData()` doesn’t??

bstien: Could it be because `reloadData()` is not called from the main thread?

FinnUI/Sources/Components/SearchFilterTags/SearchFilterTagsView.swift
```diff
@@ -0,0 +1,255 @@
...
+    // MARK: - Setup
+
+    private func setup() {
+        backgroundColor = .bgPrimary
+
+        addSubview(filterButtonView)
+        addSubview(collectionView)
+        addSubview(separatorView)
+
+        filterButtonWidthConstraint = filterButtonView.widthAnchor.constraint(equalToConstant: maxFilterButtonWidth)
```
bstien, line 103: Could this be placed as `private lazy var filterButtonWidthConstraint = filterButtonView.widthAnchor.constraint(equalToConstant: maxFilterButtonWidth)` instead?

FinnUI/Sources/Components/SearchFilterTags/SearchFilterTagsView.swift
```diff
@@ -0,0 +1,255 @@
...
+        delegate?.searchFilterTagsView(self, didTapSearchFilterTagAt: indexPath.item)
+    }
+}
+
+// MARK: - UICollectionViewDelegateFlowLayout
+
+extension SearchFilterTagsView: UICollectionViewDelegateFlowLayout {
+    public func collectionView(_ collectionView: UICollectionView,
+                               layout collectionViewLayout: UICollectionViewLayout,
+                               sizeForItemAt indexPath: IndexPath) -> CGSize {
```
bstien, line 230: Should probably be formatted like this to be consistent our style:

```suggestion
    public func collectionView(
        _ collectionView: UICollectionView,
        layout collectionViewLayout: UICollectionViewLayout,
        sizeForItemAt indexPath: IndexPath
    ) -> CGSize {
```

FinnUI/Sources/Components/SearchFilterTags/SearchFilterTagsView.swift
```diff
@@ -0,0 +1,255 @@
...
+        cellWidth = min(collectionView.bounds.width, cellWidth)
+        cellWidth = max(cellWidth, SearchFilterTagCell.minWidth)
+        return CGSize(width: cellWidth, height: SearchFilterTagCell.height)
+    }
+}
+
+// MARK: - Private extensions
+
+private extension SearchFilterTagsView {
+    private func searchFilterTagsAreEqual(_ lhs: [SearchFilterTagCellViewModel], _ rhs: [SearchFilterTagCellViewModel]) -> Bool {
```
bstien, line 242: I'm not sure if it makes sense/is possible, but could `SearchFilterTagCellViewModel` implement `Equatable`? If so, we could remove this extension.

bstien (approved):
Awesome job!😍 I know this has taken some time to finish, but the result is great!🔥

FinnUI/Sources/Components/SearchFilterTags/SearchFilterTagsView.swift
ninarg: No, I checked and it is called from main thread 🤷‍♀️

FinnUI/Sources/Components/SearchFilterTags/SearchFilterTagsView.swift
ninarg: Thanks for the feedback! 😁  I fixed all of them. 

The ViewModel was a protocol, but I changed it to a struct instead so it could implement `Equatable`. Then I could remove the extension. Looks much cleaner now ✨

PR #934, opened 2020-08-05
MotorTransaction: Add support for insurance confirmation

robinsalehjan:
# Why?

- The buyer needs to confirm that they have insurance before they can proceed.

# What?

- Add support for the "CONFIRM_INSURANCE" action
- Separate the demo into a buyer and seller segment
- Simplify and group parts of the demo code

# Show me

I noticed that the title of the segments in the screenshots below is incorrect. I fixed it the demo project but did not update the screenshots.

| Segment | Buyer - without insurance | Buyer - with insurance |
| --- | --- | --- |
| seller | insurance-not-confirmed | insurance-confirmed |


FinniversKit/Sources/Fullscreen/MotorTransactionView/MotorTransactionButton/MotorTransactionButton.swift
```diff
@@ -6,6 +6,7 @@ public enum MotorTransactionButton: String {
     case flat = "FLAT"
     case callToAction = "CALL_TO_ACTION"
     case republishAd = "REPUBLISH_AD"
+	case confirmInsurance = "CONFIRM_INSURANCE"
```
hakloev, line 9: Is it something weird about the indentation here, or is it just GitHub?

FinniversKit/Sources/Fullscreen/MotorTransactionView/MotorTransactionButton/MotorTransactionButton.swift
```diff
@@ -48,6 +51,7 @@ extension MotorTransactionButton {
         // Native actions
         case seeAd = "SEE_AD"
         case republishAd = "REPUBLISH_AD"
+		case confirmInsurance = "CONFIRM_INSURANCE"
```
hakloev, line 54: Here as well 🤔

FinniversKit/Sources/Fullscreen/MotorTransactionView/MotorTransactionButton/MotorTransactionButton.swift
robinsalehjan: I commited this before I changed from tabs to spaces in Xcode 😅

PR #935, opened 2020-08-12
MotorTransaction: Add support for FOCUS style

robinsalehjan:
# Why?

- The final step in the process (when handover of the vehicle is confirmed by both parties) will use the `FOCUS` styling on the last step.
- I added this logic in the client for the first iteration. Since we get it it from the backend now, it would be nice to have support for it, as we can apply the same styling for arbitrary steps if needed.

# What?

- Add the `FOCUS` case to the `CustomStyle` enum
- Split up the constraints for each state
- Add `currentStep` to the init signature where necessary.

### DIFF
--- a/Demo/Demo/Fullscreen/MotorTransactionView/MotorTransactionDemoView.swift
+++ b/Demo/Demo/Fullscreen/MotorTransactionView/MotorTransactionDemoView.swift
@@ -166,10 +166,10 @@ extension MotorTransactionDemoView: MotorTransactionViewDataSource {
     }
 
     func motorTransactionViewCurrentStep(_ view: MotorTransactionView) -> Int {
-		let hasCompletedAllSteps = model.steps.allSatisfy({ $0.state == .completed })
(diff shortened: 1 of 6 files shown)


Demo/Demo/Fullscreen/MotorTransactionView/MotorTransactionDemoView.swift
```diff
@@ -166,7 +166,10 @@ extension MotorTransactionDemoView: MotorTransactionViewDataSource {
     }
 
     func motorTransactionViewCurrentStep(_ view: MotorTransactionView) -> Int {
-        return model.steps.firstIndex(where: { $0.state == .active }) ?? 0
+        let hasCompletedAllSteps = model.steps.allSatisfy({ $0.state == .completed })
+        let activeStep = model.steps.firstIndex(where: { $0.state == .active }) ?? 0
+        let lastStep = model.steps.count - 1
```
bstien, line 171: `model.steps` will always contain at least one element?

bstien (approved):
LGTM!

PR #938, opened 2020-08-26
Redesign: Variant BFFT links

bstien:
# Why?
We have a task to include a variant design for BFFT links within the price view. 

Currently, we reuse `LinkButtonView` and `LinkButtonListView` for the links here. Since we need to add a heading and subheading, as well as conditionally change the order of the views within each link, I decided to do the ol' copy-paste™ of `LinkButtonView` and modify it to match our needs, which then became `PriceLinkButtonView`. 

# What?
- Added `PriceLinkButtonView` and `PriceLinkButtonListView`, plus viewmodels.
- Use these views in `ObjectPagePriceView`.

# Show me
| Normal links | Variant link, if both `heading` and `subheading` are set |
| --- | --- |
| Screenshot 2020-08-26 at 12 40 51 | Screenshot 2020-08-26 at 12 29 59 |


FinniversKit/Sources/Components/ObjectPagePriceView/Subviews/PriceLinkButton.swift
```diff
@@ -0,0 +1,118 @@
...
+    private lazy var subheadingLabel: Label = {
+        let label = Label(style: .bodyStrong, withAutoLayout: true)
+        label.textColor = .textPrimary
+        label.numberOfLines = 0
+        return label
+    }()
+
+    private lazy var subtitleLabel: Label = {
+        let label = Label(style: .detail, withAutoLayout: true)
+        label.textColor = .stone
```
ninarg, line 66: Shouldn't it be dark mode compatible? 🌔

bstien: `.stone` seems to work fine in darkmode as well, but just to be sure I'll ask the designers

Screenshot 2020-08-26 at 13 26 24

ninarg: I see, was wondering if we should use `.textSecondary` which has a dark mode version of `.stone` 😄

ninarg (approved):
Great work! 💯

FinniversKit/Sources/Components/ObjectPagePriceView/Subviews/PriceLinkButton.swift
bstien: Nice! Using that one instead😄👍

PR #940, opened 2020-09-01
Internal: replace `isIPad` checks

fespinoza:
# Why?

Because of split screen, many times we use this we really want to check for the horizontal size class instead

# What?

- Replace usages of `isIPad` for `isHorizontalSizeClassRegular`

- Add `isHorizontalSizeClassRegular` extensions to FinniversKit and FinnUI
  I added the code twice to create internal extensions instead of public ones


osanoj (approved):
LGTM 👍

PR #941, opened 2020-09-02
Redesign: ObjectPagePrice

bstien:
# Why?
We need to implement some new changes made to the price view. Currently, it's only used for BAP + motor, but it should support realestate as well.

The task specifies to add two new layouts for BFFT links, which will apply for motor and realestate.

# What?
The proxy will now send a couple of new properties: `heading`, `subheading` and `kind`. The latter specifies the layout for the link, and the proxy chooses which one to use based on what information is available as well as the market the ad belongs to.

To not break functionality with the current realestate design, a list of key/value items will be presented between the main price and the list of links. I decided to reuse `KeyValueGridView` for this purpose.

The layout can be one of three:
**`regular`**: Used for links with just a title and a URI.
**`variantCompact`**: A subheading and the button horizontally side-by-side. Will only be used by realestate for now.
**`variantFull`**: Heading, subheading, subtitle and button presented vertically. Will only be used by motor for now.

# Show me
| Before | After |
| --- | --- |
| Screenshot 2020-09-02 at 11 40 53 | Screenshot 2020-09-02 at 11 32 00 |
| Screenshot 2020-09-02 at 11 42 01 | Screenshot 2020-09-02 at 11 29 31 |


ninarg (approved):
Great work 👏

PR #942, opened 2020-09-03
Add flag to hide grouping options

fespinoza:
This is because some backend issues makes us to have some unleash toggle
in the app to release the grouping for notification center


fespinoza:
😈

PR #943, opened 2020-09-04
Improvement/readable content for notification view

fespinoza:
# Why?

Better iPad support

# What?

Use readable content guide to define horizontal constraints when it makes sense.

# Show me

| Before | After |
|---|----|
| 01-before | 01-after |
| 02-before | 02-after |
| 03-before | 03-after |
| 04-before | 04-after |


shredlocker (approved):
Looks good 👍 Didn't know about this layout guide

hakloev:
Awesome improvement 🙌 

I just want to point out a learning I had from doing this with the input form in AdIn. There is an option to still use the `layoutMarginsGuide` in each cell (making them more reusable) by setting `cellLayoutMarginsFollowReadableWidth = true` on the table view. That will make `layoutMarginsGuide` on each cell equal to the `readableContentGuide` so that you don't have to constrain the `contentView` subviews to the `readableContentGuide` specifically.

fespinoza:
oh cool!

we are waiting for the talk in the iOS meetup

PR #945, opened 2020-09-09
Fastlane: Fix Github token issue

bstien:
# Why?
We've seemed to always be asked about our GitHub token when creating a new release with Fastlane, even though it was present in either the environment or in `.env`.

# What?
- Rename function to verify and fetch github token from environment.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -35,12 +35,12 @@ end
 
 desc "Verify that environment variable exists"
 lane :verify_environment_variable do |options|
-  github_token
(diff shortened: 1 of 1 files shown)


osanoj (approved):
I assume you tested that it works like this? 😂

osanoj:
And if someone figures out why the old way has been working but doesn't work now, please enlighten us 😛

bstien:
Not able to fully test this without committing the changes and trying to create a new release from `master`, but either way we're not able to break it further😄

PR #946, opened 2020-09-10
Extended profile for business center customers

ninarg:
# Why?

Customers in admin get the extended profile element from proxy, built with proxy layout. We want to provide the same service to job customers in business center. This extended profile view supports extended profile for job. If we are going to support other markets in the future, a few more details would have to be appended. 

# What?

- Add all components extended profile for job suppoorts 
- Make sure to show/hide the relevant views based on the placement of extended profile. We have three placements, top and bottom for iPhone/compact, and sidebar for iPad/regular
- The top extended profile for compact screens is expandable, so there is some state handling for that as well
- `ToggleButton` was copied from extended profile in ios-app

# Show me

| Top compact | Bottom compact | Sidebar regular | 
| --- | --- | --- |
| Skjermbilde 2020-09-10 kl  08 50 13 | Skjermbilde 2020-09-10 kl  08 50 56 | Skjermbilde 2020-09-10 kl  09 20 07 |


bstien (approved):
Awesome job!🎉 Nice touch with the animation on `ToggleButton`!😁

bstien:
Just noticed the default highlight color of the button, which is a bit off. Could we change this so the opacity is set to i.e. 50%? Not sure what the default is😅

Screenshot 2020-09-10 at 10 47 47

ninarg:
> Just noticed the default highlight color of the button, which is a bit off. Could we change this so the opacity is set to i.e. 50%? Not sure what the default is😅
> 
> Screenshot 2020-09-10 at 10 47 47

Yes. The `ToggleButton` class was actually copied from ios-app, we use just the same one there, with the gray highlight. I think we haven't noticed because the button is so small, we always have our finger on top when it's highlighted. But I can change opacity instead 👍 😄  and do the same in ios-app.

ninarg:
@bstien I think I will move `ToggleButton` to FinniversKit, so it can be used in the old extended profile. No need to keep the same code in both repos

bstien:
> I think I will move `ToggleButton` to FinniversKit, so it can be used in the old extended profile. No need to keep the same code in both repos

I agree, that sounds like the better solution 👍

ninarg:
@bstien I added a new component `ExpandCollapseButton` and a demo for it. Fixed the highlighted color too. Ready for another look 😄

bstien (approved):
Re-approval 😄

PR #947, opened 2020-09-11
Improvement: save searches styling

fespinoza:
# Why?

People reacted a lot to the destructive delete button, since it was the only button in the bottom sheet, people felt like pressing it with the objective to confirm/quit the bottom sheet.

# What?

- Update the styling of the button
- Update the notification icon

Regarding the style of the button, I overrode a bunch of properties to have the destructive flat button styling to be able for its content to be left aligned (`Button` styles are centered by default). I still prefer to override those parameters as a special case rather than just use `UIButton`.

I also added a small hack to align the delete button image to the rest of the content. The image itself has a square shape, but then the actual content looks misaligned with the rest.

explanation

Regarding the icon, I discussed a bit with Thuy about the bell icon and dark mode. It didn't look nice, when I had that discussion (some time ago) he sent me a new version of the icons for both light and dark. I've updated them now as well.

Also notice in the before image, the icon got slightly cut off, but not anymore :D

# Show me

| Before | After |
| --- | --- |
| before-light | after-light |
| before-dark | after-dark |


ninarg (approved):
💯 👏 👍 😄

PR #948, opened 2020-09-14
Feature: MotorTransactionInvalidUserView

robinsalehjan:
# Why?

- When the user is redirected back to FINN from Swiftcourt. We have to check and ensure that the user is logged in with the account that is connected to Swiftcourt. 

- If the user is logged in with an account that is not connected to Swiftcourt, this view will be presented to ensure that they understand the implications. 

# What?

- Create a separate group `MotorTransactionInvalidUserView`
- Implement MotorTransactionInvalidUserView
- Add the 'motorTransaction' asset to the project
- Add snapshot tests

# Show me

| iPhone 11 pro | iPad pro |
| --- | --- |
| iPhone-11-pro | iPad-pro |


ag12 (approved):
Couldn't find anything here. Good job.

PR #949, opened 2020-09-14
Feature: settings button for saved searches notifications

fespinoza:
# Why?

To configure saved searches directly in the notification center

# What?

Add "more" button to the notification section header

This button will be used to present a bottom sheet for the save search
settings.

This commit includes 2 new requirements for the notification header

- `includeMoreButton` for the `NotificationCenterHeaderViewModel`
- `notificationCenterView(_:segment:didSelectMoreButtonIn)` delegate
method

This means **this is a breaking change for FinnUI**.

By default the "more" button will be hidden until the view model
indicates it needs it.

When adding the button I also gave some extra padding to the more button to
increase its tappable area, as the rest of the row is tappable to navigate to the
save search results.

I also did the improvements suggested by @hakloev on #943 after I merged that PR.

# Show me

Screen Shot 2020-09-14 at 10 02 59

The circled button will optionally appear depending on the view model setting.


shredlocker (approved):
Perfect 💯

PR #950, opened 2020-09-14
Broadcast fixes

osanoj:
# Why?
There is an issue where the same broadcast message (same id) can be displayed twice. 

# What?
- If asked to present broadcast messages only present those that are not already added.
- Update demo to set "proper" ids to make this scenario possible to test.

## Fix CornerAnchoringView
`CornerAnchoringView` is there to be able to move the tweak/settings button for the demo to the 4 corners of the screen. The problem was that it was "stealing" touches in all those corners, in the broadcast demo it was impossible to dismiss the top broadcast message, this should no longer be the case.

# Show me
(Before is shown with the `BroadcastDemoView ` changes in this PR)

| Before | After |
| --- | --- |
| broadcasts_before | broadcasts_after |


shredlocker (approved):
Looks good 👍

osanoj:
Updated snapshot test and made a small snapshot improvement, you can now quickly change the individual test case to `snapshot(.broadcast, record: true)` (i.e. add `, record: true`) to make 1 single new snapshot. Just remove it again after.

PR #951, opened 2020-09-16
Always show top shadow when scrolling SaveSearchView

ninarg:
# Why?

Today you can configure whether `SaveSearchView` should have a top shadow when scrolling. This is because there used to be a usage of this view in the app with no navigation bar on top. However, now we always want to have a navigation bar containing a button to dismiss the bottom sheet, for accessibility purposes. So this option is not needed anymore. I'm removing the option for simplicity, as we would just always set it to true. 

# What?

- Remove initializer with `usingShadowWhenScrolling` option.
- Configure view to always have top shadow

_This is a breaking change of FinnUI, since I am removing an initializer used in the app._


fespinoza (approved):
👍 LGTM

PR #954, opened 2020-09-21
Font extensions: use class instead of static properties

vadymmarkov:
# Why?

Because I want to use `FinniversKit` components with different fonts and found a way to do that with method swizzling 😅. It doesn't seem to work with static vars, but works great with class vars.

# What?

Use class instead of static properties in `UIFont` extensions. It doesn't break public API and shouldn't affect any fonts from `FinniversKit`.

# Show me

No UI changes.


osanoj (approved):
Great 👍 Now FinniversKit supports font theming 😂

PR #958, opened 2020-09-25
Feature: Blink view for object page

bstien:
# Why?
To increase the knowledge of Blink, we have a task to add a new view with a link and some information about this.

# What?
- Added `ObjectPageBlinkView` + view model.
- Added convenience init for `UIStackView`.
- Added helper method for easier conversion from `String` to `NSAttributedString`.

# Show me
Screenshot 2020-09-25 at 10 03 45


ninarg (approved):
Great work 🥇

PR #959, opened 2020-09-25
SearchListEmptyView: Add new state for save search disabled

ninarg:
# Why?

`SearchListEmptyView` was originally made to help users save a search when there's no hits. We recently did a change in the app where you're no longer allowed to save search for a user's ads. Instead of showing a blank page in these cases, I am adding a new state to this view, called `saveSearchDisabled`, to be used when there's no hits, but you're not allowed to save the search. 

Also made body text optional, since I am only providing a header title for this state.


ag12 (approved):

bstien (approved):
🎉🎉

PR #961, opened 2020-09-28
Add emailRegEx property to TexField

robinsalehjan:
# Why?

- Would like to reuse this regular expression pattern in conjunction with the `TextField` class

# What?

- Add `var emailRegEx` to the public interface
- Assign the hardcoded regular expression for email validation to the `emailRegEx` property

### DIFF
--- a/FinniversKit/Sources/Components/TextField/TextField.swift
+++ b/FinniversKit/Sources/Components/TextField/TextField.swift
@@ -137,6 +137,7 @@ public class TextField: UIView {
 
     public let inputType: InputType
     public var phoneNumberRegEx = "^(?:\\s*\\d){8,11}$"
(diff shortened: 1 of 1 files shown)


FinniversKit/Sources/Components/TextField/TextField.swift
```diff
@@ -137,6 +137,7 @@ public class TextField: UIView {
 
     public let inputType: InputType
     public var phoneNumberRegEx = "^(?:\\s*\\d){8,11}$"
+    public var emailRegEx = "[A-Z0-9a-z._%+-]+@[A-Za-z0-9.-]+\\.[A-Za-z]{2,64}"
```
bstien, line 140: Do we need to change this anywhere in the app, or can it be `let` instead?

ag12: I think the first person who implemented this just uses vars

ag12 (approved):

PR #962, opened 2020-09-30
Feature: implement saved searches notification pagination

fespinoza:
# Why?

This is a **breaking change for FinnUI** as I introduce a new protocol
requirement for `NotificationCenterViewDataSource`.

```swift
func notificationCenterView(_ view: NotificationCenterView, fetchNextPageFor segment: Int)
```

Which will be called when the user scrolls down the list of
notifications to fetch the next page if necessary.

This only will apply when the saved searches notifications are rendered
chronologically (this logic is handled by the iOS app).

We also keep track of the new fetch request with
`isFetchingNextPageForSavedSearches` to avoid making request when it's
not needed.


FinnUI/Sources/Fullscreen/NotificationCenter/Views/NotificationCenterView.swift
```diff
@@ -254,6 +259,18 @@ extension NotificationCenterView: UIScrollViewDelegate {
        reloadOnEndDragging = false
        delegate?.notificationCenterView(self, segment: selectedSegment, didPullToRefreshUsing: refreshControl)
     }
+
+    public func scrollViewDidScroll(_ scrollView: UIScrollView) {
+        let offsetY = scrollView.contentOffset.y
+        let contentHeight = scrollView.contentSize.height
+
+        let hasScrolledEnough = offsetY > contentHeight - scrollView.frame.height * 4
```
fespinoza, line 267: I am not sure about this constant name. I'm open to suggestions

shredlocker: Hmm, not sure. The only other thing I can think of is `willReachEndOfContent`, but not sure if that is better

shredlocker (approved):

FinnUI/Sources/Fullscreen/NotificationCenter/Views/NotificationCenterView.swift
fespinoza: sounds better to me at least, regarding intent

PR #963, opened 2020-09-30
Feature: IconRibbonView

bstien:
# Why?
We need to present a ribbon within the search results for realestate ads which have Blink. This ribbon needs to contain an image.

I initially looked into refactoring `RibbonView` to accept either a title or an icon, but because of the way we're using this view in the app that solution didn't seem viable. So I copied some of the code and made `IconRibbonView` instead.

The icon's size is set to 12x12, as per the designers and to not mess with the current limitations of cell height in the search results. In other words, this ribbon's height is equal to the regular `RibbonView` with text.

# What?
- Added `IconRibbonView`.
- Added icon `blinkRocketMini`.
- Refactored demo.

# Show me
Screenshot 2020-09-30 at 14 40 30


Demo/Demo/Components/Ribbon/RibbonDemoView.swift
```diff
@@ -5,73 +5,42 @@
...
-        defaultRibbonView.translatesAutoresizingMaskIntoConstraints = false
-        successRibbonView.translatesAutoresizingMaskIntoConstraints = false
-        warningRibbonView.translatesAutoresizingMaskIntoConstraints = false
-        errorRibbonView.translatesAutoresizingMaskIntoConstraints = false
-        disabledRibbonView.translatesAutoresizingMaskIntoConstraints = false
-        sponsoredRibbonView.translatesAutoresizingMaskIntoConstraints = false
+            let rowStackView = UIStackView(axis: .horizontal, withAutoLayout: true)
+            rowStackView.addArrangedSubviews([ribbonView, UIView(withAutoLayout: true)])
+            return rowStackView
+        }
```
fespinoza, line 35: mmm, I was touching ribbon view in another PR. I was thinking on potentially set the content hugging .required as part of the internal setup of ribbon view since I would imagine we wouldn't need it expandable.

then I guess this code may not be needed. It also strikes me like we could use `UILayoutGuide` here, like @hakloev taught us :D

FinniversKit/Sources/Components/Ribbon/IconRibbonView.swift
```diff
@@ -0,0 +1,73 @@
...
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import UIKit
+
+public class IconRibbonView: UIView {
+
+    // MARK: - Private properties
+
+    private lazy var iconImageView: UIImageView = {
```
fespinoza, line 11: I guess we don't need the multi-line declaration

fespinoza (approved):
LGTM 👍

FinniversKit/Sources/Components/Ribbon/IconRibbonView.swift
bstien: Yup, forgot about this one😅 Fixing

Demo/Demo/Components/Ribbon/RibbonDemoView.swift
bstien: We seem to use `setContentHuggingPriority` a lot for ribbons, so that may be a good idea! It would require a lot of changes, so we should probably add a task for it.

Regarding the layout guides: do you have any input on how they could help us here, @hakloev?🙂

FinniversKit/Sources/Components/Ribbon/IconRibbonView.swift
```diff
@@ -0,0 +1,69 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
```
tfheen, line 2: This doesn't seem to be compatible with the top-level LICENSE, nor is FINN.no an Inc.

Demo/Demo/Components/Ribbon/RibbonDemoView.swift
hakloev: Not actually, here I would have just done: 

```
private func setup() {
    stackView.alignment = .leading

    stackView.addArrangedSubviews(ribbonViews)
    addSubview(stackView)

    NSLayoutConstraint.activate([
        stackView.leadingAnchor.constraint(equalTo: leadingAnchor, constant: .spacingS),
        stackView.topAnchor.constraint(equalTo: topAnchor, constant: .spacingS),
    ])
}
```

hakloev: The default `UIStackView.Alignment` is `.fill`, so the perpendicular axis to the distribution will fail unless it's set to `.leading`

bstien: Wow, awesome!😻 Thanks!

PR #967, opened 2020-10-07
Use readable content guide in invite counter party

robinsalehjan:
# Why?

- To align the experience and design with the rest of the components in this flow

# What?

- Use `readableContentGuide` instead of `leading`/`trailing` anchor for pick a buyer view.


fespinoza (approved):
👍 LGTM

PR #969, opened 2020-10-07
Feature: color and font theming

vadymmarkov:
# Why?

Because it should be possible to use custom fonts and colors for `FinniversKit` components. I managed to implement theming with method swizzling before https://github.com/finn-no/FinniversKit/pull/954, but that was more like a temporary solution.

# What?

- Use new `Config` struct instead of `FinniversKit` class. I did it mainly because it wasn't possible to use `FinniversKit` as a module name, which is quite useful sometimes. ⚠️ **This is a breaking change for the FINN app and all our frameworks that have `FinniversKit` as a dependency. Please help me with updating them as soon as this PR is merged.**
- Introduce `FontProvider` protocol and `DefaultFontProvider` as its default implementation using FINN fonts.
- Introduce `ColorProvider` protocol and `DefaultColorProvider` as its default implementation using FINN colors.

# Usage

```swift
struct CustomFontProvider: FontProvider {
   // ...
}

struct CustomColorProvider: ColorProvider {
   // ...
}

FinniversKit.Config.fontProvider = CustomFontProvider()
FinniversKit.Config.colorProvider = CustomColorProvider()
```

# Show me

No UI changes.


fespinoza (approved):
I like it!, nice to get rid of `FinniversKit` as an object.

Nice that you followed up with a proper solution.

LGTM 👍

osanoj (approved):
Change looks good to me. But will have quite an impact on our other code 😛

PR #970, opened 2020-10-08
Internal: update to Xcode 12 / iOS 14

fespinoza:
# Why?

To make @vadymmarkov and all the iOS team happy :)

# What?

- update snapshots
- bump circle ci's config to use iOS 14 / Xcode 12.0.1
- update snapshot-testing-ios from 1.8.1 to 1.8.2


fespinoza:
Mmmm... I don't understand, I had to do a workaround for ssh taken from: https://support.circleci.com/hc/en-us/articles/360044709573-Swift-Package-Manager-fails-to-clone-from-private-Git-repositories

The error I was looking at the [old build][build] was:

```
xcodebuild: error: Could not resolve package dependencies:
  The server SSH fingerprint failed to verify.
```

Which is weird as we are not using private SPM packages and this bug was supposed to be present since Xcode 11, so changing to Xcode 12 should not have affected this issue.

I am not sure about this 🤷‍♂️, but at least the build is green 😜

[build]: https://app.circleci.com/pipelines/github/finn-no/FinniversKit/1104/workflows/a902679a-540a-4fa9-9a75-579e7ad1dac9/jobs/8769

vadymmarkov (approved):
Awesome 💯

PR #971, opened 2020-10-09
FavoriteButtonView: Remove button variant without counter

ninarg:
Our `FavoriteButtonView` had two different layouts that have been AB-tested. We concluded to keep the favorite button with counter, so I'm removing the other one.


bstien (commented):
👍

PR #976, opened 2020-10-15
Feature: notification center search view exposes favorite button

fespinoza:
# Why?

delegate method.

The `NotificationCenterSearchView` uses the `NotificationCell`, the same
as `NotificationCenterView`, then both show the `Favorite Button`
depending on the view model configuration of the cell, but
`NotificationCenterSearchView` didn't bubble up the appropriate delegate
method. This fixes that.

This is a minor change for FinnUI.


ninarg (approved):
💯

PR #978, opened 2020-10-23
Remove id property from InviteCounterpartyViewModel

robinsalehjan:
# Why?

- The property is redundant, so I would like to remove it🙈


robinsalehjan:
Closing this in favour of this [PR](https://github.com/finn-no/FinniversKit/pull/979)

PR #980, opened 2020-10-27
Fix: make grouping alternatives optional for Notification Center

fespinoza:
# Why?

We had some issues in the backend related to "group by search", then
solfrid wanted a feature flag for a single grouping option.

For that reason, if the `NotificationGroupOptionsViewModel` doesn't
include a given title, then the grouping option won't be used when
presenting the available grouping options.

I use a more "functional" approach to implement this with `Optional.map`
and `[SortOption].compactMap` to deal with the optionals. Hopefully it's
understandable enough.

# What?

- `NotificationGroupOptionsViewModel` model properties are now optional.

This is a patch-like change for FinnUI


ninarg (approved):
💯 👍 🥇

PR #986, opened 2020-11-05
Highlighted project units for realestate

ninarg:
# Why?

 A part of the new construction ads (nybygg) redesign is a new view for brokers to display some highlighted units in their project. 

# What?

Created a new `ProjectUnitsView`, which contains a scrollview with paging for iPhone. 

# Version Change

Minor change for FinnUI, patch for FinniversKit

# UI Changes

highlighted-project-units


FinnUI/Sources/Recycling/ProjectUnits/ProjectUnitCell.swift
```diff
@@ -0,0 +1,132 @@
...
+        stackView.axis = .vertical
+        stackView.alignment = .fill
+        stackView.distribution = .equalSpacing
+        return stackView
+    }()
+
+    private lazy var remoteImageView: RemoteImageView = {
+        let imageView = RemoteImageView(withAutoLayout: true)
+        imageView.contentMode = .scaleAspectFill
+        imageView.layer.cornerRadius = 8
```
fespinoza, line 25: i guess is better to use a `.spacing` constant, then if we change those constants, our radius will change accordingly

FinnUI/Sources/Recycling/ProjectUnits/ProjectUnitsView.swift
```diff
@@ -0,0 +1,270 @@
...
+
+// MARK: - UICollectionViewFlowLayout
+
+private final class PagingCollectionViewLayout: UICollectionViewFlowLayout {
+    /// Returns the centered content offset to use after an animated layout update or change.
+    override func targetContentOffset(
+        forProposedContentOffset proposedContentOffset: CGPoint,
+        withScrollingVelocity velocity: CGPoint
+    ) -> CGPoint {
+        guard let bounds = collectionView?.bounds, let layoutAttributes = layoutAttributesForElements(in: bounds) else {
```
fespinoza, line 241: i would put a comment here on what are you trying to achieve here.

why you need this? for what? i cannot tell much by just looking at the code

FinniversKit/Sources/Extensions/UIKit/UITraitCollectionExtensions.swift
```diff
@@ -1,7 +1,7 @@
 import UIKit
 
 extension UIView {
-    var isHorizontalSizeClassRegular: Bool { traitCollection.horizontalSizeClass == .regular }
+    public var isHorizontalSizeClassRegular: Bool { traitCollection.horizontalSizeClass == .regular }
```
fespinoza, line 4: if i'm not mistaken this code may be already be copied into FinnUI and the iOS app, maybe it will trigger a warning

fespinoza (approved):
Excellent work! 👍  the code looks very neat the functionality looks nice :D

FinnUI/Sources/Recycling/ProjectUnits/ProjectUnitsView.swift
ninarg: Tbh, I'm not 100% sure. I copied `PagingCollectionViewLayout` from our `NeighborhoodProfile`, which also has paging, so I didn't have to implement the paging functionality myself.
But it looks like those variables are what we need to implement the paging, and if they're not present, we just pass it on to `UICollectionViewFlowLayout` so we'd just get a regular scroll. 🤔

fespinoza: there is a `isPaginated` property in the collection view to achieve that behavior

ninarg: Thanks, will test and see if that works just as fine 😄

FinniversKit/Sources/Extensions/UIKit/UITraitCollectionExtensions.swift
ninarg: I have now copied this file to FinnUI as well, and kept it internal in both FinniversKit and FinnUI, to avoid any potential warnings 👍  (I didn't find one, but I guess this should have triggered a warning..)

FinnUI/Sources/Recycling/ProjectUnits/ProjectUnitsView.swift
ninarg: I tried with `isPagingEnabled`, and it didn't behave as I wanted. `isPagingEnabled` is made for the cases where the cell width is the same as the collectionView's width, but for this view I want the previous and next cell to be visible as well. 😄

fespinoza: I see, fair enough.

as a side note, this layout pagination I think can be achieved with the new uicollection view compositional layout, but ofc is iOS13+

PR #973, opened 2020-10-09
Experiment: SPM support

fespinoza:
# Why?

To make the new FINNNotficationCenter package to be compatible with SPM as Finnsta uses it.

I tried to quickly add SPM compatibility to see what would it take, limitations and so on.

I will probably create another PR later with the "official" spm support for FinniversKit. For that, I will need to create a new repo for FinnUI as I explained in the iOS meeting.

I will keep this PR open for a while in case you have any feedback, while I try to make the other things work.

# What?

- Add missing import statements
- Update script to generate image assets with the new Assets folder
- Rename FinnUI Resources folder to Assets
- Rename FinniversKit Resources folder to Assets
- Remove unnecessary FinnUI files
- Add initial package declaration for FinniversKit and FinnUI
- Remove FinniversKit Info.plist
- Remove FinniversKit.h and do a general UIKit import in Config
- Remove objc implementation of spacing
  SPM doesn't like multi-language packages
- Update gitignore with SPM

The rename of `Resouces` to `Assets` was because a `CodeSign` [error](  https://stackoverflow.com/questions/25835443/xcode-swift-domain-nsposixerrordomain-code-22/26617294#26617294)

I'm also deleting the objc files, including spacing. This can be probably recreated within the iOS app if needed, as SPM packages cannot have multiple languages yet.

I still need to figure out some things like:
- what to do with the Demo app
- which target should keep the snapshot tests
- fix circle ci :P

PR #987, opened 2020-11-09
Internal: SPM support for FinniversKit

fespinoza:
# Why?

<details>
Hello there

apple-svp-craig-federighi-is-a-mood-and-also-a-daddy_dj66 960

</details>

We want to support both Cocoapods and SPM with the objective to migrate to SPM in the future.

In this PR I added support for SPM for the FinniversKit module.

This means that FinniversKit can be used without problems from a project that uses swift package manager dependencies.

In order to add support for SPM, I needed to adapt the FinniversKit code with some **breaking changes**:
- remove the obj-c code
- renaming the `Resources` folder to `Assets`

The removal of the obj-c code is due that SPM packages cannot support multiple languages at a time. The `Spacing.m` file will be implemented in the iOS app directly then.

Renaming the `Resource` folder to `Assets` was done to avoid the following runtime error ( [stack overflow](https://stackoverflow.com/questions/25835443/xcode-swift-domain-nsposixerrordomain-code-22/26617294#26617294) ):

```
An error was encountered while running (Domain = NSPOSIXErrorDomain, Code = 22)
```

The version number is not specified in the `Package.swift` file as it comes from the git tag of the repo. Then releasing a new version of the pod “automatically” releases a new version of the swift package.

Finally, this PR adds SPM support for `FinniversKit`, but not for `FinnUI`. Given the version number is determined by the git tag it’s not feasible to do this for 2 packages.

This PR does not change the setup for the demo app, it still uses cocoapods, we just expose the FinniversKit package to SPM.

This PR will replace #973 

# What?

- Add support for SPM for FinniversKit
- Remove obj-c code. As SPM doesn't work with multiple languages. This means that the spacing needs to be defined inside the iOS app for obj-c. I will prepare a PR with that and check also AdInsertion.
- Rename `FinniversKit/Sources/Resources` for `FinniversKit/Sources/Assets` as spm can have runtime errors related to that folder name.
- Add an exported import of UIKit in `Config.swift`
- Handle bundle for spm and cocoapods
- Update cocoapods reference to the Assets folder

# Version Change

A breaking change for FinniversKit


ag12:
Remember to fix resources and resource bundles in the pod spec. 👌

fespinoza:
I updated the PR description @osanoj, I think this PR is ready to be merged/tested by understanding the caveats i explain above.

---

Apart from that, I also mention a bit some of the unsolved challenges i'm facing still

## Unsolved problems

There are 2 main questions remaining regarding the use of SPM in our packages. Both problems are not required to be solved before we start using SPM.

### Demo app project setup

Right now we use a `FinniversKit.xcworkspace` setup for hosting 3 Xcode projects: `Demo`, `FinniversKit` and `FinnUI`.

If we were to remove `FinnUI` from the repo, we need to make `Demo` to depend on `FinnUI` through cocoapods or spm. Right now it seems that Demo manually depends on `FinnUI` since there is no `Podfile` for that project.

When using spm packages, you don’t really use a `.xcodeproj` file, you can generate one with

```
swift package generate-xcodeproj
```

But it’s not that recommendable to edit that file, if you need to re-generate it later (*I’m not entirely sure about this*)

Then, for a spm package a way I’ve tried to add a demo app is to create a `Demo/Demo.xcodeproj` within the package repo. Then that repo can depend on the main package through SPM, even by using a “local” spm package through drag and drop.

*This is the stuff I have experimented with, but I am not sure which one is the right approach.*

### Ensure pod / spm package builds with its dependencies

When running travis we build and run the test of a project through `xcodebuild`, then an Xcode project is already configured with the dependencies/targets and all and that configuration can be done with SPM **or** Cocoapods, but not both.

Then while running the test you are testing the package being used in either cocoapods and spm. For a lot of packages this is ok, but there can be some issues with this.

Let’s say that `FINNNotificationCenter` depends on `FinniversKit`, later a breaking change on `FinniversKit` makes up update `FINNNotificationCenter`. While updating `FINNNotificationCenter` we adapt to the breaking changes by modifying its code and by bumping the version of the `FinniversKit` dependency. That last part happens either in cocoapods or spm.

Then it’s easy to forget to update the `FinniversKit` dependency in cocoapods or spm and **we don’t have an automated way to ensure the FINNNotificationCenter package still builds with both cocoapods and spm**

fespinoza:
I tried to explain as well as i could, I would like to all @finn-no/ios takes a look at this, to be both aware and also to propose solutions if you know.

osanoj:
> If we were to remove `FinnUI` from the repo, we need to make `Demo` to depend on `FinnUI` through cocoapods or spm. Right now it seems that Demo manually depends on `FinnUI` since there is no `Podfile` for that project.

I would think that if we move FinnUI to another repo the best is to split the Demo app too. So that the Demo app is for the code in the same repo. It is designed to simplify testing of the code you are changing at the moment. What do you think about that @fespinoza?

fespinoza:
> I would think that if we move FinnUI to another repo the best is to split the Demo app too. So that the Demo app is for the code in the same repo. It is designed to simplify testing of the code you are changing at the moment. What do you think about that @fespinoza?

It's possible, but I was thinking on still have one demo app and have it depend on FinnUI. Then when developing:

fespinoza:
The last thing i had in my notes is this:

* why SPM?
	* native apple / Xcode support
	* doesn’t require ruby
	* “less magic” as cocoapods does a fair amount of it
	* no “unnecessary” xcworkspace
	* quicker update dependency process
	* easier local dependencies

osanoj:
> It's possible, but I was thinking on still have one demo app and have it depend on FinnUI. Then when developing:
> 
> * you drag and drop FinnUI into the demo app to do a local dependency
> * do the changes in FinnUI because it's a local dependency
> * commit on FinnUI since it's a separate repo
> * update the FinnUI dependency after in the demo app
> 
> Maybe this is not necessarily the best way either, as there are many steps it can be more error prone.

Yes, to me, this seems like a lot of steps just to be able to test some small tweaks you are doing.

fespinoza:
Yeah, we can start by splitting the Demo app too #FinnUIExit

osanoj:
> Ensure pod / spm package builds with its dependencies

Can we have 2 targets or buildconfigs for the demo app? One for pods and one for SPM? Or do we need 2 projects then 🤔

fespinoza:
I don't really know, that's something we can potentially try to look further into

fespinoza:
ping @osanoj :)

osanoj (approved):
Sorry for the review delay, let’s merge this and continue the road to SPM! The issues you mention are not blockers and can be looked at later. Good work!

fespinoza:
after 10000 years...it's done

PR #989, opened 2020-11-13
SearchDisplayMenuView: Accessibility fixes

ninarg:
# Why?

Our sort/change display type menu is not possible to use with VoiceOver. Instead, when users try to click it, VoiceOver reads up what's behind.

# What?

- Turn `UIImageViews` into `UIButtons` with `isAccessibilityElement = true`
- Add `accessibilityLabel`s for the buttons

# Version Change

It's actually a breaking change for FinnUI, since I am changing parameters required in init.

# UI Changes

| Before | After |
| --- | --- |
| IMG_3351 | voiceover |


osanoj (approved):
LGTM 👍

PR #991, opened 2020-11-17
Redesign: SearchResultMapView

bstien:
# Why?
We want to change the annotation markers for the map search result. More specifically, the new cluster annotations should indicate how many ads are behind each cluster.

I've added a few helper methods to `MKMapView` to make registering and dequeuing annotations easier, similar to what we have for `UITableView`.

# What?
- Add custom annotation views: `PinAnnotationView` and `ClusterAnnotationView`.
- Add helper methods to `MKMapView`.

# Version Change
A breaking change for FinniversKit.

# UI Changes
| Before | After |
| --- | --- |
| Screenshot 2020-11-17 at 12 44 55 | Screenshot 2020-11-17 at 12 43 03 |


FinniversKit/Sources/Fullscreen/SearchResultMapView/Annotations/ClusterAnnotationView.swift
```diff
@@ -0,0 +1,88 @@
...
+    func configure(with annotation: SearchResultMapViewAnnotation) {
+        hitsLabel.text = "\(annotation.hits)"
+    }
+
+    // MARK: - Overrides
+
+    override func layoutSubviews() {
+        super.layoutSubviews()
+
+        layer.cornerRadius = max(bounds.width, bounds.height) / 2
```
shredlocker, line 85: I think `cornerRadius` should be `min(bounds.width, bounds.height) / 2`, then there won't be any weird clipping on the short side.

shredlocker (approved):
Looks good 👍

FinniversKit/Sources/Fullscreen/SearchResultMapView/Annotations/ClusterAnnotationView.swift
bstien: Yes, you're right👍 I'll fix it

bstien:
Closing.

PR #997, opened 2020-12-04
Job recommendation adjustment

hakloev:
# Why?

Job wants to increase the border around the company logos to make it easer to differentiate the ads. Also, they want to remove the suitcase icon when the ad is presented in the job vertical. 

# What?

* Change `imageBorder` to `sardine` to match web and meet the universal design requirements. 
* Add all subviews in `AdsGridViewCell` to `contentView`
* Add `showImageDescriptionView` to enable hiding of `imageDescriptionView`. Did some iterations on how to solve this, including passing it in `configure` or use an `OptionSet` for styling, but ended on this public property for now. This component is starting to diverge from the original simple ad cell and the number of conditionals to the design are increasing, so we should probably look into separating it into multiple cells if we add more conditions. 

# Version Change

Patch 

* I've done a search and `imageBorder` seems to only be used in `AdsGridViewCell`
* The default of `showImageDescriptionView` is `true` and it's reset in `prepareForReuse`

# UI Changes

| Before | After |
| --- | --- |
| Simulator Screen Shot - iPhone 12 Pro - 2020-12-04 at 09 22 58 | Simulator Screen Shot - iPhone 12 Pro - 2020-12-04 at 09 21 34 |


bstien (approved):

bstien:
Think you might need to re-add the snapshot image for the tests to run green again ✅

hakloev:
> Think you might need to re-add the snapshot image for the tests to run green again ✅

Yes, doing it right now. Saw afterwards that the component were used in multiple demo views, not just for `AdsGridView`

PR #998, opened 2020-12-04
MotorTransaction: Insurance picker

ninarg:
# Why?

We are adding an extra step to the transaction journey for motor, where users will be provided some free (for x days) insurances after signing the contract. This will help users to have a valid insurance instantly after buying the car, so you don't drive away without one. This PR introduces the view to select a free insurance. 

# What?

- Added `MotorTransactionInsurancePickerView`, which is filled with some text and a list of insurances. This view is supposed to be used in a bottom sheet with navigation bar on top, so therefore it's configured as a `ShadowScrollView`.
- Added `MotorTransactionInsuranceView` which contains one insurance
- `MotorTransactionInsuranceView` is a subview of the new `NavigationLinkView`. I decided to make `NavigationLinkView` reusable and put it in FinniversKit, since I saw some other use cases with a similar layout in abstract (it just had some other subview). Then it will be easier to reuse in the future. 

# Version Change

Minor change for FinnUI and FinniversKit.

# UI Changes

`NavigationLinkView`
Skjermbilde 2020-12-04 kl  09 32 12

`MotorTransactionInsurancePickerView`

insurances


bstien (approved):
Awesome job!🙌🥇

PR #999, opened 2020-12-04
Update eastereggs

osanoj:
# Why?
New year requires changes.

# What?
- Updated assets for new year. Also switched balloons and snowflake assets to pdf.
- Moved New Years and snow globe Easter eggs to FinnUI. They are technically not FINN specific, but they are not meant as component that can be used in various different places either.
- Fixed resource generation script for FinnUI by renaming "Resources" folder to "Assets".

# Version Change
Oh yes, this breaks things.

# UI Changes
Should not really be any noticeable changes except for the year.
Simulator Screen Shot - iPhone 12 - 2020-12-04 at 10 48 13
Simulator Screen Shot - iPhone 12 - 2020-12-04 at 10 48 25


ninarg (approved):
Great! 🎆

fespinoza:
#999 🎉 :octocat:

osanoj:
One more thing: `Resources -> Assets needs to be reflected in podspec too` 😅

PR #1001, opened 2020-12-07
Turn ErrorView into generic ResultView, with more config options

ninarg:
# Why?

We need a view in the app to display if ordering car insurance was successful or not. The view is just a simple view, with an icon, title and text.

# What?

The view is quite similar to our already existing `ErrorView`, just with different spacings and background color. Tried to configure it to be usable in this context as well, so we don't need another success/error view. One problem is that it felt weird to use the name `ErrorView` for a successful result. I tried to come up with a name that covers both error and success. I think `ResultView` could work, since we usually associate a result with a success or error, at least when it comes to networking. Tell me if you have a better suggestion 😄  If we decide to go with that, our existing error views in the app could keep their variable name, so it's easier to understand the context, so it would be like `let errorView = ResultView()` etc..

- Rename `ErrorView` to `ResultView`
- Add 2 tweaking options, showing how it can be used as success and error view
- Add more config options to `ResultView`, so you can choose icon size, background and icon color, and spacings.
- Add new icon to be used in success view for motor transaction

# Version Change

This is a breaking change 😬 

# UI Changes

| Dark success | Dark error |
| --- | --- |
| Skjermbilde 2020-12-07 kl  12 59 08 | Skjermbilde 2020-12-07 kl  12 59 01 |

| Light success | Light error |
| --- | --- |
| Skjermbilde 2020-12-07 kl  12 58 29 | Skjermbilde 2020-12-07 kl  12 58 39 |


robinsalehjan (approved):
LGTM🥇

bstien (approved):
🔥

PR #1002, opened 2020-12-07
Minimum deployment changed to iOS 12

osanoj:
# Why?
No need to support iOS 11.2 here when we do not do it in main app. 

# What?
- Changed minimum deployment. 
- Configured build warnings like Xcode recommended.
- Fixed deprecation warning. 

# Version Change
Technically breaking changes, but should not require any changes in main app 😄


ninarg (approved):

fespinoza (approved):
👍 LGTM

PR #1004, opened 2020-12-09
Set text font and colour through model

robinsalehjan:
# Why?

- All styling should be provided through the host app.

# What?

- Add `var titleFont: UIFont?` and `var titleTextColor: UIColor?` to `MotorTransactionStepContentViewModel`
- Set font and text color of `TitleView` in `MotorTransactionStepContentView` from model.
- Replace `.textToast` with `.licorice` for consistency

# Version Change

This is a breaking change in `FinniversKit`


osanoj:
> * Replace `.textToast` with `.licorice` for consistency

Why? We should be using semantic colors like `.textToast` since then you will have dark mode support (I know currently it is always licorice, but that might change)

robinsalehjan:
> > * Replace `.textToast` with `.licorice` for consistency
> 
> Why? We should be using semantic colors like `.textToast` since then you will have dark mode support (I know currently it is always licorice, but that might change)

If we introduce an equivalent dark mode color it might break this component later on. The color `.licorice` works regardless of the device being in `.light` or `.dark` mode.

ag12 (approved):

PR #1007, opened 2020-12-22
No public Image Assets

ninarg:
# Why?

We are going to make all Image Assets in FinniversKit internal instead of public. This will make it easier to change out icons, since you don't have to check all other SDKs that use the same one. 

# What?

- Change so image asset script doesn't support the public option anymore
- Rename `FinniversImageAsset` to `ImageAsset`, to be consistent with how it is for FinnUI and Demo
- For every FinniversKit asset used in FinnUI/Demo, I had to move the icon over to the internal assets instead
- Create image assets enum for Demo project at build time, just the way it's done for FinnUI and FinniversKit
- To fix that ☝️ I also had to do some folder renaming in Demo, from `Demo/Demo/Assets` to `Demo/Sources/Assets`. Now the folder structure is similar to FinniversKit/FinnUI, so the script just works right away
- Remove image assets demo

There might be icons that are not used anymore now in FinniversKit. I think I will delete them in a separate PR, so there's not too much happening in this one.

# Version Change

Breaking change.

# UI Changes

Hopefully no UI changes 🙏


ag12:
Wild stuff 🔥 
GJ.

ag12 (approved):
LGTM

PR #1008, opened 2021-01-06
Fix/redesign: ChatAvailabilityView

bstien:
# Why?
We want to make some changes to the texts presented in `ChatAvailabilityView`, as well as adding a new label + link at the bottom the user can click to book time instead of chatting directly when the dealership is online.

I've removed the enum `Status` so we instead keep track of the state from the app, and just configure this view to show what we give it.

I'm not completely sure if I should remove the image we used to present in the button. It's not used anywhere else at the moment, so let me know what you think.

# What?
- Add new properties to view model.
- Add new subviews.
- Remove enum `ChatAvailabilityView.Status`.

# Version Change
Breaking change of FinnUI.

# UI Changes
| Before | After: Online | After: Offline/Unknown | After: Loading |
| --- | --- | --- | --- |
| Screenshot 2021-01-06 at 12 27 39 | Screenshot 2021-01-06 at 11 54 54 | Screenshot 2021-01-06 at 11 54 56 | Screenshot 2021-01-06 at 11 55 00 |


ag12 (approved):
LGTM

PR #1009, opened 2021-01-08
Remove: Inline consent from Frontpage + related views.

bstien:
# Why?
We don't need to display inline/popup consent in the main app anymore. This means we can remove the views involved.

Neither `DialogueView` nor `InlineConsentView` was used anywhere else but for inline consent. Can't remember what convention we have for this, but I figured we might as well remove them since they were tailored only for this purpose.

# What?
- Remove inline consent from both subviews and view model in `FrontPageView`.
- Remove view `DialogueView`.
- Remove view `InlineConsentView`.
- Remove demos + snapshot test reference images.
- Fix view model used in `PopupViewDemoView`.

# Version Change
Breaking change for FinniversKit.

# UI Changes
Same as before, only less code/views🤓


robinsalehjan (approved):
Great job 🏅

PR #1010, opened 2021-01-10
Open message bottom sheet height api

ag12:
# Why?
Would be nice to use this extension for the new messaging component in the main app. 
No need for inventing the wheel again. 

# What?
Make extension BottomSheet.Height.messageFormHeight public. 

# UI Changes
None.

### DIFF
--- a/FinniversKit.podspec
+++ b/FinniversKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = 'FinniversKit'
-  s.version      = '71.0.0'
+  s.version      = '71.0.1'
   s.summary      = "FINN's iOS Components"
   s.author       = 'FINN.no'
   s.homepage     = 'https://schibsted.frontify.com/d/oCLrx0cypXJM/design-system'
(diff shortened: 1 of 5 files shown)


robinsalehjan (approved):

FinniversKit.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = 'FinniversKit'
-  s.version      = '71.0.0'
+  s.version      = '71.0.1'
```
osanoj, line 3: Better to follow this: https://github.com/finn-no/FinniversKit/blob/master/RELEASING.md

And isn’t it a minor version change since it adds functionality instead of fixing bug?

ag12: I will do! 71.1.0 then?

osanoj: Yes, or whatever the script tells you is next minor when you have merged 😄

ag12: great, AI taking over 💯

PR #1011, opened 2021-01-15
Build to update ImageAsset enum

ninarg:
# Why?

Forgot to include a file before merging #1007 . 

# What?

Building the project resulted in a change in `ImageAsset` enum, that I want to include before bumping FinniversKit + FinnUI.


ag12 (approved):
gj.

PR #1012, opened 2021-01-15
SwiftUI textfields

hakloev:
# Why?

Did an experiment with SwiftUI in the job team. We decided to stop the experiment, but I felt these components could be added to FinniversKit so we do not loose the value of the work. At least as initial versions.

# What?

* Added `FinnTextView`
* Added `FinnTextField`
* Slight refactor in the demo app to remove redundant code 

I tried multiple approaches before landing on this one: 

1.  Tried to use the existing views in FinniversKit by creating SwiftUI components using `UIViewRepresentable`. That solution did'nt work as intented, as it required us to always set a specific frame height on the view in SwiftUI. A learning from this attempt were that only default UIKit views worked nicely with `UIViewRepresentable`. 
2. Tried to make a more generic SwiftUI view that wrapped `SwiftUI.TextField` and `SwiftUI.SecureField`. That did not work well either, as `SecureField` does'nt expose `onEditingChanged` and thus making it impossible to detect if the textfield were focused (not able to adjust the underline color and height)

### Naming:

Since `TextField` and `TextView` is already used in `FinniversKit`, I ended up calling them `FinnTextField` and `FinnTextView`. Suggestions are welcome. 

This actually occured other places as well, for i.e. `Button`, `EmptyView` and `Identifiable`, since they're already used in FinniversKit, thus being a namespace collision with SwiftUI. 

### Location:

I've placed the textviews in a folder named `SwiftUI` within the already existing `TextView` and `TextField` folders. Not sure if that's the best location? 

# Version Change

Minor?

# UI Changes

| | Feature A | 
| --- | --- | 
| **TextView** |  Simulator Screen Shot - iPhone 12 Pro - 2021-01-15 at 16 32 03 |
| **TextField** |  Simulator Screen Shot - iPhone 12 Pro - 2021-01-15 at 16 31 58 |


FinniversKit/Sources/Components/TextField/SwiftUI/FinnTextField.swift
```diff
@@ -0,0 +1,266 @@
...
+                .foregroundColor(disclosePassword ? Color.accentSecondaryBlue : Color.iconPrimary)
+        })
+    }
+}
+
+@available(iOS 13.0, *)
+extension FinnTextField {
+
+    private static let emailRegex: String = "[A-Z0-9a-z._%+-]+@[A-Za-z0-9.-]+\\.[A-Za-z]{2,64}"
+    private static let phoneRegex: String = "^(?:\\s*\\d){8,11}$"
```
nickfalk, line 134: I think I'll just trust you on this. 😅

FinniversKit/Sources/Components/TextField/SwiftUI/FinnTextField.swift
```diff
@@ -0,0 +1,266 @@
...
+
+        return isValid
+    }
+
+    private func evaluateTextState() {
+        if !isValidInput {
+            updateStyle(.error)
+        } else {
+            updateStyle(.default)
+        }
```
nickfalk, line 162: I won't comment on the fact that you're checking if-not-else. 😜

Have you considered: `updateStyle(isValidInput ? .default : .error)`

FinniversKit/Sources/Components/TextField/SwiftUI/FinnTextField.swift
```diff
@@ -0,0 +1,266 @@
...
+    private func isValidEmail(_ emailAdress: String) -> Bool {
+        return evaluate(FinnTextField.emailRegex, with: emailAdress)
+    }
+
+    private func isValidPhoneNumber(_ phoneNumber: String) -> Bool {
+        return evaluate(FinnTextField.phoneRegex, with: phoneNumber)
+    }
+
+    private func isValidPassword(_ password: String) -> Bool {
+        return !password.isEmpty
```
nickfalk, line 179: So spaces are OK?

FinniversKit/Sources/Components/TextField/SwiftUI/FinnTextField.swift
hakloev: Hehe, I'll also just trust the one who created them. Shamelessly copied them from `TextField` 🙈

FinniversKit/Sources/Components/TextField/SwiftUI/FinnTextField.swift
hakloev: Will do!

FinniversKit/Sources/Components/TextField/SwiftUI/FinnTextField.swift
hakloev: I'll add a trim. Shamelessly copied this one from `TextField` as well 🙈

nickfalk (approved):
LGTM! 👍

PR #1014, opened 2021-01-20
Add accessibilityLabel

hakloev:
# Why?

We didn't read the content

# What?

The panel component had `isAccessibilityElement` set to `true`, but didn't set an `accessibilityLabel` and thus didn't read the content of the label. 

# Version Change

patch

# UI Changes

No


nickfalk (approved):
👍

PR #1016, opened 2021-01-26
Remove FinnUI

osanoj:
# Why?
FinnUI has been moved to https://github.com/finn-no/finnui-ios

# What?
Removed FinnUI code and updated docs.

# Version Change
It is not breaking FinniversKit, that should not have changed. FinnUI has moved but is it broken here then...? 🤔


ninarg (approved):
🎉 👏

PR #1029, opened 2021-03-24
Motor transaction promo: Slide view and link

ninarg:
# Why?

We want to promo the motor transaction journey on the front page 🎉 

# What?

We're doing it as an AB-test. The control version is just to include a link on our front page, which will redirect to a landing page at FINN.no. The B-version is a bigger redesign of the front page, with a slide view, with optional number of slides. The first promo will just have 1 slide, but we'll expand it in the future.

- Modified our `FrontPageView`, to contain an optional `PromoLinkViewModel`. If the view model isn't nil, we append the new `PromoLinkView` on the front page.
- Create `PromoSlidesView`, which can be configured to contain any number of slides. This isn't included in any front page yet, @bstien is working on it!
- I tried to make `PromoSlidesView` as tall as the tallest slide, with full width slides. We'll see if this works when integrating it to the front page.

# Version Change

Breaking, but merging this into our other base branch!

# UI Changes

**Slide view**
slide

**Link view**
simulator_screenshot_3A98BD33-B764-4DA9-8FBE-338F77201F4D

**iPad**
| Link | Slide view |
| --- | --- |
| Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2021-03-24 at 10 32 57 | Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2021-03-24 at 10 35 24 |


bstien (approved):
Great job!🔥🔥🔥

PR #1032, opened 2021-03-25
Motor promotion: Add PromoSlidesView to promotional frontpage

bstien:
# Why?
We want to show a promotion on the frontpage😄

# What?
- Added `PromoSlidesView` to `PromotionFrontpageView`.
- Hardcode height for `PromoSlidesView` (for now).
  - This should definitely be set dynamically, but if everyone is OK with it we'll ship it with hardcoded values for now.

# Version Change
None yet, going into a WIP branch.

# UI Changes
https://user-images.githubusercontent.com/1901556/112473977-4d23a500-8d6f-11eb-8f58-ce7586f58928.mp4


ninarg (approved):
Perfect 🔥

PR #1033, opened 2021-03-29
Front page: BasicFrontPageView protocol, UI adjustments

ninarg:
# Why?

Some various adjustments needed to be made to insert the promo in the app.

# What?

- Added `BasicFrontPageView` protocol that both versions of the front page conform to, so they can be used the same way in the app
- Added some extra spacing to `PromoSlidesView` 
- Icon adjustments to primings
- Expose a public Y anchor to point a promotion callout to in the app
- Height adjustments for for iOS 12 & iOS 13, they couldn’t have the same height as iOS 14 for some reason, I think `UIPageControl` is occupying more space 🤔 , but since we don’t support dynamic height for this view yet, I’ll just increase the constant as a quick fix in the first release (we’ll add dynamic height support later)

# Version Change

- Not bumping yet, merging this into our base branch `motor-transaction-promotion`

# UI Changes

Snapshot changes

Skjermbilde 2021-03-29 kl  09 57 25


bstien (approved):

ninarg:
Is there some cache to delete in circle ci? I can't make my new snapshot for `promotionFrontPageView` pass, even though it's recorded on same device as circle ci (iPhone 11 Pro), and it's passing locally 😅

PR #1039, opened 2021-05-04
Feature: Add video view to ContractActionView

bstien:
# Why?
As part of the TJM we want to add a link to a video to the contract teaser to explain the transaction journey process.

# What?
- Added `ContractVideoLinkView`.
- Updated model to accept an optional title and `VideoLink`.

# Version Change
Breaking. Adding a new delegate method.

# UI Changes

| Today's state | With video + title |
| --- | --- |
| Screenshot 2021-05-04 at 14 30 13 | Screenshot 2021-05-04 at 14 29 55 |


FinniversKit/Sources/Components/ContractActionView/Subviews/ContractVideoLinkView.swift
```diff
@@ -0,0 +1,123 @@
...
+            playOverlayView.centerYAnchor.constraint(equalTo: thumbnailImageView.centerYAnchor),
+            playOverlayView.widthAnchor.constraint(equalToConstant: 40),
+            playOverlayView.heightAnchor.constraint(equalToConstant: 40)
+        ])
+    }
+
+    func configure(with videoLink: ContractActionViewModel.VideoLink, remoteImageViewDataSource: RemoteImageViewDataSource) {
+        titleLabel.text = videoLink.title
+        thumbnailImageView.dataSource = remoteImageViewDataSource
+        thumbnailImageView.loadImage(for: videoLink.thumbnailUrl.absoluteString, imageWidth: .zero)
```
ninarg, line 77: Looks kinda weird with zero width, but I guess the image is a ImagePathKind `.url`, because then we don't really use the image width? 😄

bstien: Yeah, I have no idea how to get around this😅 We use it to append the size when fetching from our own CDN, but when we need images from other sources this property makes no sense.

I'm integrating this the app now. If it doesn't work I'll look into fixing it :)

ninarg: Maybe we can do some changes in `RemoteImageViewDataSource` to make width optional 🤔  That doesn't necessarily have to be done in this PR though

ninarg (approved):
Great work 🥇

PR #1040, opened 2021-05-06
Set number of lines in TransactionStepView to 5

robinsalehjan:
# Why?

- We might get strings that exceeds 2 lines

# What?

- Increase `numberOfLines` from `2` to `5`

# Version Change

- Minor

### DIFF
--- a/FinniversKit/Sources/Components/TransactionEntryView/TransactionStepView.swift
+++ b/FinniversKit/Sources/Components/TransactionEntryView/TransactionStepView.swift
@@ -17,7 +17,7 @@ public class TransactionStepView: UIView {
 
     private lazy var textLabel: Label = {
         let label = Label(style: .detail, withAutoLayout: true)
-        label.numberOfLines = 5
+        label.numberOfLines = 2
         return label
     }()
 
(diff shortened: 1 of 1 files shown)


FinniversKit/Sources/Components/TransactionEntryView/TransactionStepView.swift
```diff
@@ -17,7 +17,7 @@ public class TransactionStepView: UIView {
 
     private lazy var textLabel: Label = {
         let label = Label(style: .detail, withAutoLayout: true)
-        label.numberOfLines = 2
+        label.numberOfLines = 5
```
ninarg, line 20: Good catch 😄 
We decided to have a maximum of two lines on the front page entry, just because that fits the designs best.
I suggest we can make it possible to configure differently, but keep 0 as the default, like this:
```
public func configure(withTitle title: String?, text: String?, showWarningIcon: Bool, numberOfTextLines: Int = 0) {
         textLabel.numberOfLines = numberOfTextLines
```
And from `TransactionEntryView`, we can keep `numberOfTextLines: 2`
Or what do you think? 😄

robinsalehjan: Sure, i'll fix it 👍

ninarg (approved):
Nice 👍

PR #1044, opened 2021-07-02
Add block user view

ag12:
# Why?
We want to figure out why users are blocking other users. 

# What?
Added block user view

# Version Change
Minor change

# UI Changes
Screen Shot 2021-07-02 at 10 46 36


ag12:
Screen Shot 2021-07-02 at 10 48 57
Screen Shot 2021-07-02 at 10 49 08

Block button is enabled or disabled based on if user have selected the radio btn

FinniversKit/Sources/Components/BlockUserView/BlockUserView.swift
```diff
@@ -0,0 +1,182 @@
...
+        return button
+    }()
+
+    public weak var delegate: BlockUserViewDelegate?
+
+    private var viewModel: BlockUserViewModel? {
+        didSet {
+            title.text = viewModel?.title
+            subtitle.text = viewModel?.subtitle
+            radio.fields = viewModel?.reasons ?? []
```
osanoj, line 77: It shouldn't happen I guess, but maybe we should enable the button if there are no reasons? Like if the backend is down users will still be able to block other users even if we will not get to know why

ag12:
Its hardcoded for now based on spesifications..

FinniversKit/Sources/Components/BlockUserView/BlockUserView.swift
ag12: should not happen. but its good to go

ag12:
🍒

osanoj (approved):

PR #1045, opened 2021-07-12
Add optional action button in ResultView

ninarg:
# Why?

Modify our `ResultView` to have an optional action button. Going to use it as a retry-this-search view in search quest, when backend fetch fails.

# What?

- Add action button to `ResultView`
- Make it optional in `configure` func

# Version Change

Patch 

# UI Changes

| Before | After |
| --- | --- |
| Simulator Screen Shot - iPhone 12 Pro - 2021-07-12 at 09 21 02 | Simulator Screen Shot - iPhone 12 Pro - 2021-07-12 at 09 21 49 |


ag12 (approved):
lgtm

PR #1047, opened 2021-08-09
Update the price tag on StandardAdRecommendationCell

magnustviberg:
# Why?

Trying a blurred background that can adapt to the image background and can maintain accessibility

# What?

The price tag on a StandardAdRecommendationCell displayed on torget has got a new design. Pill shaped corner radius, blur effect and new insets and constraints.

# Version Change

no

# UI Changes

Simulator Screen Shot - iPhone 12 Pro - 2021-08-11 at 12 26 44


ninarg (commented):
Looks great 😸

hakloev:
Probably not in the scope of this PR, but is the price tag added even if the string is empty? Seems like the empty string is rendered and thus causing an alignment issue. As seen in the "Alpakka"-ad, the market icon is pushed slightly to the left.

ninarg:
Try recreating the snapshot with iPhone 12 Pro, iOS 14.5 if you didn’t use that one already. That’s where the tests are run in CircleCI 😅

ninarg:
> Probably not in the scope of this PR, but is the price tag added even if the string is empty? Seems like the empty string is rendered and thus causing an alignment issue. As seen in the "Alpakka"-ad, the market icon is pushed slightly to the left.

Regarding this, I think we actually had a task ages ago to "center" the icon in the price tag if the price string isn't there. So If that's a quick change @magnustviberg , then you could fix it - or we could create another task for it

magnustviberg:
Finally got the tests to run successfully. And centered the icon when no price is displayed @ninarg @hakloev

magnustviberg:
Sorry for the messy PR but the tests passed locally but not in circleCI 🤷‍♂️

FinniversKit/Sources/Recycling/GridViews/AdRecommendations/Cell/StandardAdRecommendationCell.swift
```diff
@@ -372,6 +394,15 @@ public class StandardAdRecommendationCell: UICollectionViewCell, AdRecommendatio
             performViewChanges()
         }
     }
+    
+    private func centerIconInContainer(_ shouldCenter: Bool) {
+        guard let leadingConstraint = self.iconImageLeadingConstraint,
+              let centerXConstraint = self.iconCenterXConstraint else {
+            return
+        }
+        leadingConstraint.priority = shouldCenter ? .defaultLow : .defaultHigh
```
ninarg, line 403: I have one idea. Would it work to just remove the trailing padding for the price text label if there's no price (that would be change `imageDescriptionView.trailingAnchor` spacing to 0)? Then we don't have to switch between these constraints, and the left padding would always be the same, so it looks nicer in my opinion. (now the left padding looks bigger for the centered icons, because of the bigger width).

If I'm not wrong, then the icon would look centered, because it would have `StandardAdRecommendationCell.margin` on both leading and trailing 🤔  Correct me if I'm wrong!

ninarg (approved):
Great work 💯

PR #1050, opened 2021-08-23
Redesign of MarketsGridView

magnustviberg:
# Why?

The marketlist is redesigned. This includes both the grid view and the compact view.

# What?

This is the first of two PRs and contains the implementation of the new design for the MarketsGridView. Its is now scrollable and the cells have shadow. 

_The front page is going to get a new background color which will make the shadow look better than what it currently does on a white background_

_The colors should be updated when the work on the new color palette  in Fabric is done_

# UI Changes

| Before | After |
| --- | --- |
| Simulator Screen Shot - iPhone 12 Pro - 2021-08-18 at 11 09 52 | Simulator Screen Shot - iPhone 12 Pro - 2021-08-23 at 16 10 12 |

### GIF
Simulator Screen Recording - iPhone 12 Pro - 2021-08-23 at 16 01 31


vikram-86 (approved):
👌🏿

PR #1051, opened 2021-08-25
Feature: Request access to existing contract

robinsalehjan:
# Why?

- To remove confusion between buyer/seller on which contract to use use.
- If the seller has created a contract (will be bundled with the ad purchase flow). We will present this view to let the buyer request access to the sellers contract.

# What?

- Add the `contract` icon to `Assets`
- Add description label below the subtitle in the `ContractActionView` component
- Add `description` property to `ContractActionViewModel`
- Make top and trailing padding of the image customisable in the public `configure` method

# Version Change

- Minor

# GIF

Aug-25-2021 14-19-12


FinniversKit/Sources/Components/ContractActionView/ContractActionView.swift
```diff
@@ -89,8 +97,12 @@ public class ContractActionView: UIView {
     public func configure(
         with viewModel: ContractActionViewModel,
         trailingImage: UIImage? = nil,
+        trailingImageTopConstant: CGFloat = 0,
```
robinsalehjan, line 100: If we rename `trailingImage` to `image`. We can drop the `trailing` prefix in the parameters below, that would however break the API. 

Let me know if I should rename the parameters, I don't feel any particular way about it.

magnustviberg: It would be nice to drop it, but I think it should be ok to just keep it and optionally create a small refactoring task on it in Jira or something. Because other than that I think it looks good

magnustviberg (approved):

PR #1052, opened 2021-08-30
Redesign of CompactMarketsView

magnustviberg:
# Why?

The marketlist is redesigned. This includes both the grid view and the compact view.

# What?

This is the second PR and contains the implementation of the new design for the CompactMarketsView. It now has a drop shadow instead of a border and has newly designed cells which no longer contain the icon.

_The colors should be updated when the work on the new color palette  in Fabric is done_

# UI Changes

| Before | After |
| --- | --- |
| Simulator Screen Shot - iPhone 12 Pro - 2021-08-18 at 11 09 43 | Simulator Screen Shot - iPhone 12 Pro - 2021-08-30 at 13 01 24 |

### GIF
Simulator Screen Recording - iPhone 12 Pro - 2021-08-30 at 13 01 39


FinniversKit/Sources/Recycling/GridViews/Markets/CompactView/CompactMarketsViewCell.swift
```diff
@@ -147,10 +123,14 @@ extension CompactMarketsViewCell {
 // MARK: - Private extensions
 
 private extension UIColor {
-    class var externalLinkColor: UIColor {
-        dynamicColorIfAvailable(defaultColor: .sardine, darkModeColor: .darkSardine)
-    }
-}
+     class var externalLinkColor: UIColor {
+         return .blueGray600
```
ninarg, line 127: does it look nice with dark mode too?

magnustviberg: Betzy told me to use the same color for both light and dark mode

PR #1057, opened 2021-09-09
Changed minimum version to iOS 13

osanoj:
# Why?
We only support iOS 13 and above.

# What?
- Changed minimum version to iOS 13
- Removed checks for iOS 13 or above in code. Also removed one iOS 12 check.
- Fixed some SwiftLint warnings.

# Version Change
Major


ninarg (approved):
Awesome 👏

PR #1058, opened 2021-09-10
Feature: HelthjemView

robinsalehjan:
# Why?

- To give more exposure to the `HeltHjem` integration we have for sending packages. We currently show the `InfoboxView` on the object page to highlight this integration (not ideal)

- This view will replace the `InfoboxView` on the object page

# What?

- Create `HelthjemView`
- Add snapshot test 

# Version Change

- Minor

# UI Changes

| | Light | Dark |
| --- | --- | --- |
| **iPhone** | Simulator Screen Shot - iPhone 12 Pro Max (14 1) - 2021-09-10 at 09 27 28 |  Simulator Screen Shot - iPhone 12 Pro Max (14 1) - 2021-09-10 at 09 27 37 |
| **iPad** | Simulator Screen Shot - iPad Pro (12 9-inch) (14 1) - 2021-09-10 at 09 30 32 | Simulator Screen Shot - iPad Pro (12 9-inch) (14 1) - 2021-09-10 at 09 30 21 |


FinniversKit/Sources/Components/HelthjemView/HelthjemView.swift
```diff
@@ -0,0 +1,219 @@
...
+    func helthjemViewDidSelectSecondaryButton(_ view: HelthjemView)
+}
+
+public final class HelthjemView: UIView {
+    public weak var delegate: HelthjemViewDelegate?
+
+    // MARK: - Subviews
+
+    private lazy var imageView: UIImageView = {
+        let image = Config.userInterfaceStyleSupport == .forceLight ?
```
ninarg, line 14: hmm, this will fail for `.dynamic`, because we'd always use dark mode icon even if its was dynamic but light 🤔

ninarg:
Approve the rest, great work 🏅 
I think this view would be a good candidate to be put in FinnUI instead of FinniversKit, since it has a very specific and non-reusable use case. Unless the TJT package needs to display it somehow, since that package probably only depends on FinniversKit, not FinnUI, but looks like this will only be shown from object page?
Or what do you think? 😄

robinsalehjan:
> Approve the rest, great work 🏅
> I think this view would be a good candidate to be put in FinnUI instead of FinniversKit, since it has a very specific and non-reusable use case. Unless the TJT package needs to display it somehow, since that package probably only depends on FinniversKit, not FinnUI, but looks like this will only be shown from object page?
> Or what do you think? 😄

Yes, I can move it into `FinnUI`. It's not something the TJT package depends as of now, but that may change in the future (then I may move this view into the package if necessary) 👍

ninarg:
> Yes, I can move it into `FinnUI`. It's not something the TJT package depends as of now, but that may change in the future (then I may move this view into the package if necessary) 👍

Sounds good 😄

FinniversKit/Sources/Components/HelthjemView/HelthjemView.swift
robinsalehjan: Nice catch, pushed up a change to cater for `.dynamic` type.

FinniversKit/Sources/Components/HelthjemView/HelthjemView.swift
```diff
@@ -0,0 +1,239 @@
...
+    public weak var delegate: HelthjemViewDelegate?
+
+    // MARK: - Subviews
+
+    private lazy var helthjemImage: UIImage = {
+        let lightImage = UIImage.init(named: .shipWithHelthjem)
+        let darkImage = UIImage.init(named: .shipWithHelthjemDarkmode)
+
+        if #available(iOS 13.0, *) {
+            switch Config.userInterfaceStyleSupport {
```
ninarg, line 18: I think this can be simplified to just check `traitCollection.userInterfaceStyle` , if it's `.dark` or `.light`. 
And, if you want the image to update to the correct color if the user switched between light and dark mode while being on the object page, you have to `override traitCollectionDidChange` to update the image (but it might be acceptable to keep the image as is, since it's just an illustration, it's not super bad if it doesn't update 😛 )

robinsalehjan:
[Component has been moved to `FinnUI`](https://github.com/finn-no/finnui-ios/pull/26)

PR #1061, opened 2021-09-29
Tweaks to FeedbackView size and presentation

ninarg:
# Why?

I'm gonna insert `FeedbackView` into the new search result cells. Since the cells are self-sizing, I have to make some adjustments to this view to make it layout correctly.

# What?

Be explicit about what kind of presentation to use for the cell (grid or list), so I added an option to set a fixed presentation, instead of using the automatic one (which determines display type based on the view's dimensions).

Set a `heightAnchor` for the title label, so self-sizing cells don't give it zero height.
Also set a max height for image in list presentation - since self-sizing cells allowed it to be much larger than how it was designed.

A small dark mode fix.

# Version Change

Patch.

# UI Changes

There shouldn't be any UI changes.


bstien (approved):

ag12 (approved):
LGTM

PR #1062, opened 2021-09-29
Fix: Remove responsibility from UserAdManagementUserActionCell

bstien:
# Why?
`UserAdManagementUserActionCell`'s viewmodel, `AdManagementActionCellModel`, had too much responsibility. One of its properties, `AdManagementActionType`, is a decodable model that we use in AdInsertion and is formed after the response we get from the old ad-management endpoint, which will no longer be the case.

# What?
There is a difference in cell height between the old and this solution.
The previous solution had constraints between the top/bottom of `titlelabel` and `contentView`, and `titleLabel` is smaller than the icon we present on the left.
This setup sets the constraint between top/bottom `contentStackView` and `contentView`, which makes the cells ~5px taller. I don't think this should be an issue.

- Refactor `AdManagementActionCellModel` to be given an image, instead of knowing which one to present based on the parsed response from the old endpoint.
- Refactor `UserAdManagementUserActionCell` to use stackviews and an optional trailing item.
- Update demo.
  - Icons previously used have been moved to the demo.

# Version Change
Breaking.

# UI Changes

| Before | After |
| --- | --- |
| Screenshot 2021-09-29 at 14 12 11 | Screenshot 2021-09-29 at 15 39 39 |


bstien:
I will check if we use the icons I moved to the demo anywhere else. If not, I'll remove them.

Demo/Sources/Recycling/ListViews/AdManagement/AdManagementDemoView.swift
```diff
@@ -23,34 +21,35 @@ public class AdManagementDemoView: UIView, Tweakable {
...
         [
-            AdManagementActionCellModel(actionType: .rentalApplications, title: "Gå til enkel utleie dashboard", showExternalIcon: true)
+            AdManagementActionCellModel(title: "Gå til enkel utleie dashboard", iconImage: .stakeholder, trailingItem: .external)
         ],
         [
-            AdManagementActionCellModel(actionType: .delete, title: "Slett annonsen"),
-            AdManagementActionCellModel(actionType: .stop, title: "Skjul annonsen midlertidig", description: "Annonsen blir skjult fra FINNs søkeresultater")
+            AdManagementActionCellModel(title: "Slett annonsen", iconImage: .adManagementTrashcan),
+            AdManagementActionCellModel(title: "Skjul annonsen midlertidig", description: "Annonsen blir skjult fra FINNs søkeresultater", iconImage: .eyeHide),
+            AdManagementActionCellModel(title: "Flott togle", iconImage: .checkCircle, trailingItem: .toggle),
```
ninarg, line 37: toggle 😇

FinniversKit/Sources/Recycling/ListViews/AdManagement/Cells/UserAdManagementUserActionCell.swift
```diff
@@ -90,135 +81,89 @@ public class UserAdManagementUserActionCell: UITableViewCell {
...
+        switch model.trailingItem {
+        case .none:
+            break
+        case .chevron:
+            trailingImageView.image = UIImage(named: .arrowRight).withRenderingMode(.alwaysTemplate)
+            trailingImageView.isHidden = false
+        case .external:
+            trailingImageView.image = UIImage(named: .webview).withRenderingMode(.alwaysTemplate)
+            trailingImageView.isHidden = false
+            contentStackViewTrailingConstraint.constant = -24
```
ninarg, line 141: `.spacingL`

ninarg (approved):
Great improvement 👏

PR #1068, opened 2021-11-02
Feature/recently favorited shelf

vikram-86:
# Why?

We are going to add a shelf on front page 
which contains recently favorited items.. this implements the components according to design
Skjermbilde 2021-11-02 kl  15 11 32

# What?

Adds new cell and a collection view that uses compositional layout

# Version Change

minor

# UI Changes

| Before | After |

Simulator Screen Shot - iPhone 12 Pro - 2021-11-02 at 15 12 08


FinniversKit/Sources/Cells/RecentlyFavoritedShelfCell/RecentlyFavoritedShelfCell.swift
```diff
@@ -0,0 +1,275 @@
...
+        label.translatesAutoresizingMaskIntoConstraints = false
+        label.backgroundColor = .clear
+        label.textAlignment = .center
+        
+        return label
+    }()
+    
+    private lazy var priceBackground: UIView = {
+        let background = UIVisualEffectView(withAutoLayout: true)
+        if #available(iOS 13.0, *) {
```
osanoj, line 46: We only support iOS 13 and above now, so this check can be removed 🙂

FinniversKit/Sources/Cells/RecentlyFavoritedShelfCell/RecentlyFavoritedShelfCell.swift
```diff
@@ -0,0 +1,275 @@
...
+            background.effect = UIBlurEffect(style: .systemThinMaterialDark)
+        } else {
+            background.effect = nil
+            background.backgroundColor = UIColor.priceLabelBackgroundColor.withAlphaComponent(0.8)
+        }
+        background.alpha = 1.0
+        background.layer.cornerRadius = RecentlyFavoritedShelfCell.priceTagHeight / 2
+        background.clipsToBounds = true
+        
+        background.contentView.addSubview(priceLabel)
```
osanoj, line 56: Could you move this and the setup of the constraints into the `setup()` func? To gather all the view configurations there if possible.

FinniversKit/Sources/Cells/RecentlyFavoritedShelfCell/RecentlyFavoritedShelfCell.swift
vikram-86: Thats right! totally forgot that part, thanks ! ☺️

FinniversKit/Sources/Cells/RecentlyFavoritedShelfCell/RecentlyFavoritedShelfCell.swift
vikram-86: will do! 😁

FinniversKit/Sources/Cells/RecentlyFavoritedShelfCell/RecentlyFavoritedShelfCell.swift
```diff
@@ -0,0 +1,275 @@
...
+    private var isFavorited: Bool = true {
+        didSet {
+            favoriteButton.isToggled = isFavorited
+        }
+    }
+    
+    private var task: URLSessionDataTask?
+    
+    private lazy var imageView: UIImageView = {
+        let imageView = UIImageView()
```
osanoj, line 24: I think using RemoteImageView should be better here, then you do not need any `URLSessionDataTask` in this class and can use the image loading method provided

osanoj (approved):
LGTM 👍  But I think CircleCI wants a new snapshot image 😄

PR #1070, opened 2021-11-08
implementing shelf component on frontpage

vikram-86:
# Why?
Added shelf on front page

# What?

using the front-page shelf component, this PR implements that component into front page

# Version Change

 minor

# UI Changes
Simulator Screen Shot - iPhone 12 Pro - 2021-11-08 at 14 59 24


FinniversKit/Sources/Fullscreen/FrontPage/FrontPageView.swift
```diff
@@ -269,6 +297,17 @@ public final class FrontPageView: UIView, BasicFrontPageView {
         
     }
     
+    private func addFrontPageShelf() {
+        guard shelfViewModel != nil else { return }
```
ninarg, line 301: This is called every time we change `shelfViewModel` from the app. Just in case we do it twice, we could add a `frontPageShelfView?.removeFromSuperview()`, to make it impossible to add multiple.

FinniversKit/Sources/Fullscreen/FrontPage/FrontPageView.swift
```diff
@@ -343,3 +382,75 @@ extension FrontPageView: MarketsViewDelegate {
...
+    }
+}
+
+//MARK: - RemoteImageViewDataSource
+extension FrontPageView: RemoteImageViewDataSource {
+    public func remoteImageView(_ view: RemoteImageView, cachedImageWithPath imagePath: String, imageWidth: CGFloat) -> UIImage? {
+        return nil
+    }
+    
+    public func remoteImageView(_ view: RemoteImageView, loadImageWithPath imagePath: String, imageWidth: CGFloat, completion: @escaping ((UIImage?) -> Void)) {
```
ninarg, line 437: We already have a `CDNRemoteImageViewDataSource` in the app to handle image fetching. So you can add `remoteImageViewDataSource` to the init of this class, and use that one instead.

ninarg (approved):
🥇

FinniversKit/Sources/Fullscreen/FrontPage/FrontPageView.swift
```diff
@@ -269,6 +298,18 @@ public final class FrontPageView: UIView, BasicFrontPageView {
         
     }
     
+    private func addFrontPageShelf() {
```
ninarg, line 301: I noticed we call this function every time the view model changes (and thus remove the full view hierarchy and add it again). We should reload the shelf collection view instead, less expensive 😄

PR #1071, opened 2021-11-10
Fix: Make it possible to set text of TextField

bstien:
# Why?
When using this view I noticed that it's not possible to set the text manually, which I need for a form I'm adding.

# What?
- Add `set` block for `TextField`'s `text`-property.

# Version Change
- Minor.

# UI Changes
_No UI._

### DIFF
--- a/FinniversKit/Sources/Components/TextField/TextField.swift
+++ b/FinniversKit/Sources/Components/TextField/TextField.swift
@@ -149,7 +149,13 @@ public class TextField: UIView {
     }
 
     public var text: String? {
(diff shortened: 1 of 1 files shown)


osanoj (approved):
👍 

I guess you could have used `textField.textField.text = ""`, but I agree it makes sense to be able to use `textField.text = ""`

bstien:
> I guess you could have used `textField.textField.text = ""`, but I agree it makes sense to be able to use `textField.text = ""`

Yes, but I want the validation to occur when I set it. I figured this was the easier way :)

PR #1072, opened 2021-11-10
Support recommendation cells without icon

osanoj:
# Why?
Because we will be removing the market icons in the cells.

# What?
I thought it best to keep the code for having icons a little longer incase we decide to go back on this.

- Reworked support for configuring the cells to hide the image overlay completely, since the old way felt a bit error prone and could override my first attempt at hiding the overlay whenever the icon or image text was not set.
- Made the overlay in the cells handle either missing icon or missing text (using a stackview).

# Version Change
Major change since it changes the public api 😞 

# UI Changes
Added a couple of cells to demo how it looks without icon and when hiding the overlay completely:
Simulator Screen Shot - iPhone 12 Pro - 2021-11-10 at 11 28 21


ag12 (approved):
👍

PR #1078, opened 2021-12-14
Fix: Make TextView (a bit) more customizable

bstien:
# Why?
For the realestate sold state I need to remove the underline and change it's backgroundColor, neither of which are possible now.

We also need to limit the number of characters the user is allowed to enter, so I've proxied the delegate method `textView(_:shouldChangeTextIn:replacementText:)` through `TextViewDelegate`.

# What?
- Added two methods for configuring the view: removing the underline and changing the background color.
- Proxy the delegate method `UITextViewDelegate.textView(_:shouldChangeTextIn:replacementText:)`.

# Version Change
Patch/minor. The delegate method has a default implementation and is optional.

# UI Changes

| Before | Changed bgColor | Changed bgColor and removed underline |
| --- | --- | --- |
| Screenshot 2021-12-14 at 17 57 28 | Screenshot 2021-12-14 at 17 58 24 | Screenshot 2021-12-14 at 17 58 27 |


ninarg (approved):
Nice 👍

PR #1080, opened 2021-12-17
Remove traces of AB-test with promotion front page

ninarg:
# Why?

This spring we ran some AB-tests with a separate front page for the app. It's a while since we stopped the experiment, and I don't think we're planning to take the alternative front page back. 
So I will remove traces of the experiment, to keep our front page clean. If we want to bring it back some day, we can just check the history 😄 

# What?

Remove the promo box on our current front page - as we don't use it anymore.

Delete 
- `PromotionFrontPageView`
- `PromoSlidesView` + all the different slides it contained
- `PromoLinkView`
- Demos and tests related to it

# Version Change

Minor since I remove all usages in the app first, _then_ merge this.


bstien (approved):
Awesome! Nice to remove so much stuff😁✨

PR #1082, opened 2021-12-20
Fix: LinkButtonListView - override color for external icon

bstien:
# Why?
External links for the sold state should have external icons, but we need to override their colors.

# What?
- Add optional property to viewModel and view to override the color.
  - If nil it will use the default color, aka. sardine.

# Version Change
Patch

# UI Changes
Screenshot 2021-12-20 at 17 41 19


ninarg (approved):
👍  Might be that you have to update your snapshot with Xcode 13 etc.

PR #1084, opened 2022-01-13
Update: Add caption label to ObjectPageTitleView

bstien:
# Why?
We need a label for realestate to present the address.

# What?
- Added caption label + style to `ObjectPageTitleView`.

# Version Change
- Minor. Both properties are optional.

# UI Changes
Screenshot 2022-01-13 at 13 05 55


ag12 (approved):
LGTM

PR #1085, opened 2022-01-20
Component: OverflowCollectionView

bstien:
# Why?
In FinnUI we added a view that had a "tag"-like collectionView that pushes elements that doesn't fit on the current row onto the next. I need a view like this in another component, so I'm pulling this out into a generic reusable view and placing it here for others to use.

## How to use
Since we want to reuse this view with different cells, and therefor also using different models, I created the protocol `OverflowCollectionViewCell` that needs to be implemented by the cell type. Because of this, `OverflowCollectionView` is only capable of using a single cell-type.

Because of the generic used in the protocol, `OverflowCollectionView` can use each of the models it receives and ask the cell to calculate the size using that model.

I also added `OverflowCollectionView.CellSpacing` which let's us change the spacing between each element on a row, and the spacing between rows.

# What?
- Added `OverflowCollectionView`.

# Version Change
Minor. Only a new view added.

# UI Changes
Screenshot 2022-01-20 at 13 51 31


bstien:
Also, this view fixes an issue in the original collectionView where the height would not be updated when/if the data changes. Now the collectionView will resize itself to account for more/fewer elements :)

ninarg (approved):
💯

PR #1087, opened 2022-01-20
Fix: Remove dark mode override for demo

bstien:
# Why?
The manual override of darkmode is annoyance that prevents us from checking how our views react when the `userInterfaceStyle` changes.

# What?
- Remove all traces of forced `userInterfaceStyle` override from demo.
- Fix correct appearance for navigationBar in demo `UINavigationController`.
- Cleanup of misc. demo related files.
  - There's still more to cleanup, but no need to clutter this PR with that.

# Version Change
Major. Both the app and FinnUI depends on some of the models removed.

# UI Changes
Simulator Screen Recording - iPhone 13 Pro - 2022-01-20 at 18 37 00


bstien:
I had an issue in `ButtonDemoView` where all the section titles would say `Unknown`, instead of the current style. I've seen it before when running in simulator, but it would be gone on the next run. Now I would get the same when running snapshot tests.

The reason might be that `UIColor` is a reference type, and the styles have somehow gotten the same instance in some cases and a different one in others.

bstien:
It seems a lot of our libs were using `dynamicColorIfAvailable(defaultColor:darkModeColor:)`, so I've re-added it with a deprecation notice. It now simply calls `dynamicColor(defaultColor:darkModeColor:)`, so no functionality is broken.

PR #1089, opened 2022-01-25
Fix: Make it possible for buttons to have multiple lines in ObjectPagePriceView

bstien:
# Why?
We need to support longer titles for the link buttons in `ObjectPagePriceView`. We seem to already have support by setting `button.titleLabel?.numberOfLines` to `0`, but the extra lines are not accounted for when `Button` calculates its `intrinsicContentSize` so the buttons end up looking pretty cramped because of the missing space. See the 'Before'-image below, where I've set `numberOfLines = 0`.

I found a solution thanks to [this answer](https://stackoverflow.com/a/34228276), which has a pretty good explanation on how to solve this issue and why it exists to begin with.

# What?
- Add ad-hoc `UIButton`-subclass for price link buttons.
- Cleanup stackView usages.

# Version Change
Patch. Only changes within the view itself.

# UI Changes
| Before | After |
| --- | --- |
| Screenshot 2022-01-25 at 17 26 45 | Screenshot 2022-01-25 at 16 47 04 |

### DIFF
--- a/FinniversKit/Sources/Components/ObjectPagePriceView/Subviews/PriceLinkButtonView.swift
+++ b/FinniversKit/Sources/Components/ObjectPagePriceView/Subviews/PriceLinkButtonView.swift
@@ -17,7 +17,6 @@ class PriceLinkButtonView: UIView {
     // MARK: - Private properties
 
     private let viewModel: PriceLinkButtonViewModel
-    private let linkButtonStyle = Button.Style.link.overrideStyle(smallFont: .body)
     private lazy var fillerView = UIView(withAutoLayout: true)
     private lazy var externalImage = UIImage(named: .webview).withRenderingMode(.alwaysTemplate)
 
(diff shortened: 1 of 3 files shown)


ag12 (approved):
LGTM.

FinniversKit/Sources/Components/ObjectPagePriceView/Subviews/PriceLinkButtonView.swift
```diff
@@ -143,8 +142,52 @@ class PriceLinkButtonView: UIView {
...
 }
+
+private class PriceLinkButton: UIButton {
+    override var intrinsicContentSize: CGSize {
+        /// We need to override this to get the correct height for button with multiple lines of text.
+        guard let titleSize = titleLabel?.intrinsicContentSize else { return .zero }
+
+        return CGSize(
+            width: titleSize.width,
+            height: titleSize.height + 10 // 2 * .spacingXS + 2 (aka. magic number).
```
ag12, line 158: 🥇

PR #1083, opened 2022-01-12
Feature/shelf fix

vikram-86:
# Why?
curshed some bugs in frontshelfView

_A short description on what's the reasoning was behind these changes._
there were some UI and logic bugs that needed to be crushed

PR #1090, opened 2022-01-27
Front page shelves bug fixes

ninarg:
# Why?

We have a bug where the shelf would sometimes be empty:
It was a bit difficult to reproduce, it happened randomly now and then.

empty-shelf

# What?

I think I found the cause behind this bug, as least I haven't been able to reproduce it anymore after this fix.
The problem was that the cells were never created (`cellForItemAt` never called) since the frame of the shelf container ended up being squeezed by the container above and getting 0 height. And cells are never created if the height is 0, so it never ended up "claiming" the space that it needed.
I added a height anchor:
`frontPageShelfView?.heightAnchor.constraint(greaterThanOrEqualToConstant: 100)`
that makes sure the view always has a minimum height.  The constant 100 is a bit random, the shelf will always end up with a taller height anyways.

**Other bug fix**
We used to `setupFrames()` only when setting up the view for the first time. This caused the height to be wrong if one of the shelves disappeared or reappeared. We should always update the height, since the state could have changed.

# Version Change

None, merging this into Sutha's PR #1083


vikram-86:
I've tested this and indeed it fixes the bug 🔥  Great Catch ! 👍

PR #1091, opened 2022-02-01
Improve height calculations for front page shelves, and add more section spacing

ninarg:
# Why?

I noticed the height calculations for front page shelves were wrong, it doesn't match the actual height of the shelf, so we ended up with a "random" spacing in the bottom of it.

# What?

- Make static vars for the heights, so we don't need to keep a copy of the same height in two locations (where the height is calculated, and where the constraint is set), this makes it easier to change a height later
- Remove `topPadding` from height calculations, there wasn't any top padding used in the view
- Add `interSectionSpacing` for shelves, since they should have a bit more space
- Fix some warnings

# Version Change

Patch.

# UI Changes
A bit more spacing.

Skjermbilde 2022-02-01 kl  15 36 56


bstien (approved):
Good fix! 🥇

PR #1093, opened 2022-02-11
Detail callout

ninarg:
# Why?

We need a new kind of callout to highlight new verticals in the app (Næringstorget).

# What?

We already have one kind of callout, called `CalloutView`. 
The new callout is called `DetailCalloutView`, since it's smaller and has detail font.

`DetailCalloutView` is mostly a copy of `CalloutView`, but some things changed: 
- Less spacing
- Different font (.detail)
- The arrow can point to the left or right (instead of top or bottom)
- It's simpler - no functions to animate in & out, it's just configured with a text

# Version Change

Minor.

# UI Changes
Skjermbilde 2022-02-11 kl  13 41 54


magnustviberg (approved):
Looks good👏

PR #1094, opened 2022-02-11
Fix: Support for snapshot tests on Intel and M1

bstien:
# Why?
As mentioned in the iOS meeting yesterday the snapshot tests would not run successfully on M1, since there's a small color difference between the architectures when rendering shadows and cornerRadius.

The issue is described here:
https://github.com/pointfreeco/swift-snapshot-testing/issues/313
https://github.com/pointfreeco/swift-snapshot-testing/issues/424

As we discussed [there's an open PR](https://github.com/pointfreeco/swift-snapshot-testing/pull/571) that masks this issue by allowing pixels to have a slightly different value, which will work for us until we're all on M1 sometime™. Since it's not merged yet we will point to @pimms fork (thank you!) until it is or we figure out another solution. 

`subpixelThreshold` will make sure that the color channel values (RGBA) of all pixels are within the threshold. I.e. the snapshot tests won't fail if the reference pixel is #000000FF and the produced pixel is #040404FF.

## PNG issue
After setting this up yesterday some tests were still failing, and I noticed that there is a _tiiiiiny_ difference in placement when it comes to rendering some PNG images. 

https://user-images.githubusercontent.com/1901556/153592207-d0c560aa-79d7-4f10-9d30-33faf6b6731e.mp4

I'm not sure of the reason and if this applies to all PNGs, or just those with different sizes _(@1x, @2x, @3x)_, but I reached out to Sanjin and got an updated icon in PDF to use when we need a placeholder.

# What?
- Point `SnapshotTesting` to https://github.com/pimms/swift-snapshot-testing
- Set `subpixelThreshold` to 5.
- Replace PNG image `.noImage` with an updated PDF.
- Retake snapshots for all views using this 👆 

# Version Change
Minor. No breaking changes.

# UI Changes
| Old icon | New icon |
| --- | --- |
|  |  |


pimms:
I just fixed a bug, you may want to use commit ebe5672 instead.

osanoj (approved):

PR #1095, opened 2022-02-16
Component: MultilineButton

bstien:
# Why?
Our `Button`-class isn't capable of presenting a multiline string properly, which I now need in two different places. I added an ad-hoc class in #1089 for `ObjectPagePriceView`, but now I need something similar in `LinkButtonListView` as well. The difference is that here it's possible to override the button style. 

`Button` seems to be capable of presenting the title on multiple lines when there's a newline present, but not without it. I've tried to make `Button` take longer text into account when calculating its `intrinsicContentSize`, like the implementation I've added to `MultilineButton`, but it breaks almost all of the snapshot tests since there are either a few pixels less or a few pixels extra in either height or width.

So to get a working implementation, and not break everything else, I've added `MultilineButton` which subclasses `Button`.

# What?
- Add `MultilineButton`.
- Use this in `ObjectPagePriceView` and `LinkButtonListView`.

# Version Change
Minor. Only a new view added.

# UI Changes

| `Button` with a long text | `Button` with text containing newlines | `MultilineButton` |
| --- | --- | --- |
| Screenshot 2022-02-16 at 09 35 02 | Screenshot 2022-02-16 at 09 34 31 | Screenshot 2022-02-16 at 09 34 08 |


FinniversKit/Sources/Components/Button/Button/MultilineButton.swift
```diff
@@ -0,0 +1,39 @@
...
+
+    public override var intrinsicContentSize: CGSize {
+        guard let titleSize = titleLabel?.intrinsicContentSize else { return .zero }
+
+        let paddings = style.paddings(forSize: size)
+        let imageSize = imageView?.image?.size ?? .zero
+
+        return CGSize(
+            width: ceil(titleSize.width) + style.margins.left + style.margins.right + paddings.left + paddings.right + imageSize.width,
+            height: ceil(titleSize.height) + style.margins.top + style.margins.bottom + paddings.top + paddings.bottom + 2
```
bstien, line 15: 🧙‍♂️🪄

osanoj (approved):

PR #1097, opened 2022-03-14
KeyValueGridView: Use word wrapping

bstien:
# Why?
The `lineBreakMode` set for key/value labels in `KeyValueGridView` was `byCharWrapping`, which is probably not what we wanted when implementing this view.

# What?
- Use `byWordWrapping` instead.

# Version Change
Patch.

# UI Changes
| Before | After |
| --- | --- |
| Screenshot 2022-03-14 at 15 27 04 | Screenshot 2022-03-14 at 15 26 03 |


ninarg (approved):
✂️ 🚫

PR #1099, opened 2022-03-16
Improve release script

osanoj:
# Why?
A while back I was looking into automating building new Demo versions, but for that to work on the CI system we would need some changes to how certificates and prov profiles are managed (automatic by Xcode doesn't work on CI).

I did actually make some improvements to the release script and a lane to archive a version of the Demo app to be uploaded to TestFlight manually. 

# What?
- Improvements to our release script, added better version management so it will now set "Xcode" version properly in the projects.
- Added fastlane lane to build a FinniversKit Demo app, which can then manually be distributed from Xcode Organizer. It is not ideal, but it will enable us to quickly make new TestFlight version for internal testing.
- Fixed some listing issues.
- Updated gems.

# Version Change
Minor

# UI Changes
No UI changes


bstien (approved):
LGTM!

PR #1101, opened 2022-03-23
Tjt 908 make text field component customizable

thorbjxrn:
# Why?

These changes allows TextFields to be customized with a dynamic border, needed to comply with a feature design

# What?

An additional configureBorder() method was added, that takes a Boolean for enabling the variation. A default border color variable was added to keep the configureBorder method call color, as the other variable is used to change the border color dynamically

# Version Change

minor

# UI Changes

https://user-images.githubusercontent.com/17337441/159730283-6e3f9441-a949-42d4-a6ad-4ebeb54a6d2d.mp4

(although the preview is not included in the PR, just the component)

### DIFF
--- a/FinniversKit/Sources/Components/TextField/TextField.swift
+++ b/FinniversKit/Sources/Components/TextField/TextField.swift
@@ -328,8 +328,8 @@ public class TextField: UIView {
         textFieldBorderColor = color
         textFieldBackgroundView.layer.cornerRadius = radius
         textFieldBackgroundView.layer.borderWidth = width
-        setNeedsLayout()
         transition(to: state)
+        setNeedsLayout()
     }
 
     // MARK: - Actions
--- a/FinniversKit/Sources/Components/TextField/TextField.swift
+++ b/FinniversKit/Sources/Components/TextField/TextField.swift
@@ -55,7 +55,9 @@ public class TextField: UIView {
     private var textFieldBorderColor: UIColor?
     private var textFieldDefaultBorderColor: UIColor?
     private var textFieldDynamicBorder: Bool?
-
+    
+    public var textRegex: String?
+    
     private var underlineHeightConstraint: NSLayoutConstraint?
     private var helpTextLabelLeadingConstraint: NSLayoutConstraint?
 
@@ -185,7 +187,7 @@ public class TextField: UIView {
         case .phoneNumber:
             isValidByInputType = isValidPhoneNumber(text)
         case .normal, .multiline:
-            isValidByInputType = true
+            isValidByInputType = isValidRegEx(text)
         }
 
         if isValidByInputType, let customValidator = customValidator {
@@ -376,6 +378,13 @@ public class TextField: UIView {
 
     private func isValidPassword(_ password: String) -> Bool {
         return !password.isEmpty
+    }
+    
+    private func isValidRegEx(_ text: String) -> Bool {
+        guard let regex = textRegex else {
+            return true
+        }
+        return evaluate(regex, with: text)
     }
 
     private func isHelpTextForErrors() -> Bool {
--- a/FinniversKit/Sources/Components/TextField/TextField.swift
+++ b/FinniversKit/Sources/Components/TextField/TextField.swift
@@ -314,8 +314,8 @@ public class TextField: UIView {
         setNeedsLayout()
     }
     
-    public func configureFullBorder(radius: CGFloat, width: CGFloat, color: UIColor) {
-        textFieldFullBorder = true
+    public func configureBorder(radius: CGFloat, width: CGFloat, color: UIColor, dynamicBorder: Bool) {
+        textFieldFullBorder = dynamicBorder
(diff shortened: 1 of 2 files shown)


FinniversKit/Sources/Components/TextField/TextField.swift
```diff
@@ -312,6 +314,17 @@ public class TextField: UIView {
         transition(to: state)
         setNeedsLayout()
     }
+    
+    public func configureBorder(radius: CGFloat, width: CGFloat, color: UIColor, dynamicBorder: Bool) {
```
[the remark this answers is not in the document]
robinsalehjan: Setting `dynamicborder: Bool = false` should be sufficient.

FinniversKit/Sources/Components/TextField/TextField.swift
```diff
@@ -394,8 +407,20 @@ public class TextField: UIView {
 
     private func transition(to state: State) {
         layoutIfNeeded()
-        underlineHeightConstraint?.constant = state.underlineHeight
-
+        
+        if let dynamicBorder = self.textFieldDynamicBorder, dynamicBorder == true {
+            if state != .normal {
```
robinsalehjan, line 412: What if we use a `switch` statement here instead?

```
switch state {
    case .normal: ...
    default: ...
}
self.layoutSubviews()
...
```

hakloev: FYI: Actually, we shouldn’t call `layoutSubviews` here. According to Apple we should never call it ourselves, but rather call `setNeedsLayout` or/and `layoutIfNeeded`. Details [here](https://developer.apple.com/documentation/uikit/uiview/1622482-layoutsubviews)

FinniversKit/Sources/Components/TextField/TextField.swift
```diff
@@ -183,7 +187,7 @@ public class TextField: UIView {
         case .phoneNumber:
             isValidByInputType = isValidPhoneNumber(text)
         case .normal, .multiline:
-            isValidByInputType = true
+            isValidByInputType = isValidRegEx(text)
```
bstien, line 190: Could we move the implementation of the method here? 🙂 The name of the method is somewhat misleading since it seems it checks the validity of the regex and not if the text matches it.

I.e.
```swift
if let textRegex = textRegex else {
    isValidByInputType = evaluate(textRegex, with: text)
} else {
    isValidByInputType = true
}
```

FinniversKit/Sources/Components/TextField/TextField.swift
```diff
@@ -53,7 +53,13 @@ public class TextField: UIView {
     private let errorIconWidth: CGFloat = 16
     private var textFieldBackgroundColorOverride: UIColor?
     private var textFieldBorderColor: UIColor?
-
+    private var textFieldDefaultBorderColor: UIColor?
+    private var textFieldDynamicBorder: Bool?
+    
+    private var untouched = true
```
robinsalehjan, line 59: What if we rename it to: `displayHelpTextOnFirstRender = false`

FinniversKit/Sources/Components/TextField/TextField.swift
```diff
@@ -385,7 +412,10 @@ public class TextField: UIView {
     }
 
     private func evaluateCurrentTextState() {
-        if let text = text, !text.isEmpty, !isValid {
+        if textRegex != nil, !untouched, let text = text, text.isEmpty {
```
robinsalehjan, line 415: And change it here to:

```
if textRegex != nil, displayHelpTextOnFirstRender, let text = text, text.isEmpty
```

FinniversKit/Sources/Components/TextField/TextField.swift
```diff
@@ -140,6 +146,7 @@ public class TextField: UIView {
     public let inputType: InputType
     public let emailRegEx = "[A-Z0-9a-z._%+-]+@[A-Za-z0-9.-]+\\.[A-Za-z]{2,64}"
 
+    @available(*, message: "deprecated in favor of TextField+InputType.swift phoneNumber(String)")
```
robinsalehjan, line 149: This can be removed.

FinniversKit/Sources/Components/TextField/TextField.swift
```diff
@@ -410,10 +451,11 @@ public class TextField: UIView {
         UIView.animate(withDuration: animationDuration) {
             self.layoutIfNeeded()
             self.underline.backgroundColor = state.underlineColor
+
             self.textFieldBackgroundView.backgroundColor = self.textFieldBackgroundColorOverride ?? state.textFieldBackgroundColor
             self.typeLabel.textColor = state.accessoryLabelTextColor
             self.helpTextLabel.textColor = state.accessoryLabelTextColor
-
+
```
robinsalehjan, line 458:
```suggestion

```
I guess we don't need this blank line? 😄

robinsalehjan (approved):
🚢 🚢

PR #1102, opened 2022-03-25
Make Christmas promotion flexible, so it can be used in other contexts

ninarg:
# Why?

We already have support for a `ChristmasPromotion` on the front page, the same space is going to be used for other promos like hjerterom for refugees. I want to make it more general, so it can be reused in different contexts. This means also removing a bunch of fixed sizing etc.

# What?

ℹ️  I renamed `ChristmasPromotionView/ChristmasPromotionView.swift` to `PromotionView/PromotionView.swift`, but decided to temporarily keep the old name while this PR is reviewed because then you can actually see a diff (git didn’t recognise it as the same file after rename). Will name it back after it has been approved. 

- Remove all Christmas prefixes, now it’s just Promo
- Make buttons optional, hide them if not configured with a text
- Add another optional text label
- Optional background color for image
- Had to do some layout adjustments to make the layout look good for both kinds of promos
    - Now the text takes 60% of the vertical space and the image up to 40%
    - The height of the promo view follows the text content’s height, so the image adjusts to that
    - Configure image with an alignment - the icon looks best filling the space, while the Christmas image looks best trailing
- Instead of having to set a fixed height, now the header frame is resized according to the size of the promo
- Make entire promo view clickable
- Use auto layout for front page `headerView` - a bunch of layout conflicts was printed out when not because the view had a temporary width of 0

# Version Change

Breaking

# UI Changes

Skjermbilde 2022-03-25 kl  11 02 52

Skjermbilde 2022-03-25 kl  11 02 41
hjerterom-ipad


osanoj (approved):
LGTM 👍

bstien (approved):
Great job 👏 🔥

PR #1103, opened 2022-03-31
Updated API so that the callout font can be updated from the outside

magnustviberg:
# Why?

The design for TJT Shipping opt-in tooltip uses a different font than the default one

# What?

Updated API so that the callout font can be updated from the outside using a setFont() function

# Version Change

minor


robinsalehjan (approved):
💯

PR #1104, opened 2022-04-04
Component/sidescrollable view

AtishPapa:
# Why?

We need a side scrollable view to show "tabs" on the user ads view.

Screenshot 2022-04-04 at 15 48 19

Currently it's not customisable since the only use case for it is the user ads view.

# What?

Add `SideScrollableView`.

# Version Change

Minor: A new view was added.

# UI Changes

| Static | Moving |
| --- | --- |
| Screenshot 2022-09-29 at 14 39 43 | RocketSim_Recording_iPhone_13_Pro_2022-09-29_14 39 51 |


FinniversKit/Sources/Components/SideScrollableView/SideScrollableView.swift
```diff
@@ -0,0 +1,188 @@
...
+    
+    public override var intrinsicContentSize: CGSize {
+        let height = SideScrollableView.contentInset.top +
+                     labelHeight +
+                     SideScrollableView.contentInset.bottom
+        return CGSize(width: UIView.noIntrinsicMetric,
+                      height: height)
+    }
+
+    public weak var delegate: SideScrollableViewDelegate?
```
bstien, line 86: Public properties should go at the top of the class for visibility

FinniversKit/Sources/Components/SideScrollableView/SideScrollableView.swift
```diff
@@ -0,0 +1,188 @@
...
+        // The current implementation always makes the first item as selected
+        for (index, itemName) in itemTitles.enumerated() {
+            let button = Button.makeSideScrollableButton(withTitle: itemName)
+            contentView.addArrangedSubview(button)
+            buttons.append(ButtonItem(title: itemName,
+                                      button: button))
+
+            button
+                .publisher(for: .touchUpInside)
+                .sink { [weak self, index] _ in
```
bstien, line 155: `index` is an Int so it should already be captured by this closure without the placing it in this list, right?

FinniversKit/Sources/Components/SideScrollableView/SideScrollableView.swift
```diff
@@ -0,0 +1,188 @@
...
+        indicatorView.backgroundColor = .primaryBlue
+        indicatorView.heightAnchor.constraint(equalToConstant: 4).isActive = true
+        return indicatorView
+    }()
+    
+    private lazy var labelHeight: CGFloat = {
+        "I".height(withConstrainedWidth: .greatestFiniteMagnitude, font: .captionStrong)
+    }()
+    
+    private var buttons: [ButtonItem] = []
```
bstien, line 69: Maybe this can be renamed to `buttonItems`?

FinniversKit/Sources/Extensions/UIKit/UIControlExtensions.swift
```diff
@@ -0,0 +1,60 @@
+//
+//  Copyright © 2022 FINN AS. All rights reserved.
+//
+
+import Combine
+import UIKit
+
+/// A custom subscription to capture UIControl target events.
```
bstien, line 8: Maybe we should add a credit/link to the article where we found this solution?😄

FinniversKit/Sources/Components/SideScrollableView/SideScrollableView.swift
```diff
@@ -0,0 +1,188 @@
...
+import UIKit
+
+public struct SideScrollableItemModel {
+    
+    let items: [String]
+    
+    public init(items: [String]) {
+        self.items = items
+    }
+}
```
bstien, line 15: I think this could be pulled out into another file, just to keep models and the view separated :)

FinniversKit/Sources/Components/SideScrollableView/SideScrollableView.swift
```diff
@@ -0,0 +1,188 @@
...
+
+public protocol SideScrollableViewDelegate: AnyObject {
+
```
bstien, line 18: This newline should be removed

FinniversKit/Sources/Components/SideScrollableView/SideScrollableView.swift
```diff
@@ -0,0 +1,188 @@
...
+            button
+                .publisher(for: .touchUpInside)
+                .sink { [weak self, index] _ in
+                    self?.selectedItemPublisher.send(index)
+                }
+                .store(in: &cancellables)
+        }
+
+        contentView.layoutIfNeeded()
+        indicatorViewWidthConstraint.constant = buttons[0].button.frame.width
```
bstien, line 162: I don't think we have any guarantee about the number of items we receive, so this should probably rather be

```swift
indicatorViewWidthConstraint.constant = buttons.first?.button.frame.width ?? 0
```

FinniversKit/Sources/Extensions/UIKit/UIControlExtensions.swift
```diff
@@ -0,0 +1,60 @@
...
+}
+
+/// Extending the `UIControl` types to be able to produce a `UIControl.Event` publisher.
+protocol CombineCompatible { }
+extension UIControl: CombineCompatible { }
+extension CombineCompatible where Self: UIControl {
+    func publisher(for events: UIControl.Event) -> UIControlPublisher<Self> {
+        return UIControlPublisher(control: self, events: events)
+    }
+}
```
bstien, line 60: These extensions could potentially grow long over time, and `CombineCompatible` can be used in combination with other publishers, so maybe we should place these in a separate file?

bstien (commented):
Great job😄 🎉 This seems to be a good example of where we can utilize Combine😁

The code itself looks good to me, but I left a few nitpicking comments about style/formatting

FinniversKit/Sources/Components/SideScrollableView/SideScrollableView.swift
```diff
@@ -0,0 +1,188 @@
...
+        let title: String
+        let button: Button
+    }
+    
+    private lazy var scrollView: UIScrollView = {
+        let scrollView = UIScrollView(withAutoLayout: true)
+        scrollView.addSubview(contentView)
+        scrollView.showsHorizontalScrollIndicator = false
+        scrollView.contentInset = .init(vertical: 0,
+                                        horizontal: SideScrollableView.horizontalInset)
```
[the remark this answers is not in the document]
AtishPapa: Okay, I'll keep this in mind 😉

FinniversKit/Sources/Components/SideScrollableView/SideScrollableView.swift
```diff
@@ -0,0 +1,188 @@
...
+        let stackView = UIStackView(withAutoLayout: true)
+        stackView.axis = .horizontal
+        stackView.spacing = SideScrollableView.itemSpacing
+        return stackView
+    }()
+    
+    private lazy var indicatorView: UIView = {
+        let indicatorView = UIView(withAutoLayout: true)
+        indicatorView.backgroundColor = .primaryBlue
+        indicatorView.heightAnchor.constraint(equalToConstant: 4).isActive = true
```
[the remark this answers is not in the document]
AtishPapa: Yup, good remark 👍

FinniversKit/Sources/Components/SideScrollableView/SideScrollableView.swift
```diff
@@ -0,0 +1,188 @@
...
+        for (index, itemName) in itemTitles.enumerated() {
+            let button = Button.makeSideScrollableButton(withTitle: itemName)
+            contentView.addArrangedSubview(button)
+            buttons.append(ButtonItem(title: itemName,
+                                      button: button))
+
+            button
+                .publisher(for: .touchUpInside)
+                .sink { [weak self, index] _ in
+                    self?.selectedItemPublisher.send(index)
```
hakloev, line 156: I think I would argue that we actually don't need this `selectedItemPublisher` as it's not published out of the view. Also we could reduce the number of subscriptions by one. 

My proposal is to simply remove it, and instead rename `subscribeToButtonTap` to `handleButtonTap(_ index: Int)` and call it directly from this sink. That way we also remove the need for weak assigning in `subscribeToButtonTap`.

```
button
  .publisher(for: .touchUpInside)
  .sink { [weak self, index] _ in 
    self?.handleButtonTap(index, button)
  }
  .store(in: &cancellables)
``` 

and

```
private func handleButtonTap(_ index: Int, _ button: UIButton) {
  let selectedItem = buttons[selectedIndex]
  scrollView.scrollRectToVisible(button.frame, animated: true)

  indicatorViewWidthConstraint.constant = button.frame.width
  indicatorViewLeadingConstraint.constant = button.frame.
  
  UIView.animate(withDuration: 0.2, delay: 0, options: .curveEaseOut) {
    self?.layoutIfNeeded()
  }

  delegate?.sidescrollableViewDidTapItem(self, item: selectedItem.title, itemIndex: index)
}
```

AtishPapa: That's great, thanks 🙌 Somehow I felt that the second publisher is kind of unnecessary 😄

hakloev:
I'm a bit unsure about the name `SideScrollableView`. My immediate thought is that it's a horizontally scrollable scroll view, but in fact it's a tab view or a more fancy `UISegmentedControl`. Maybe we should consider naming it something else? I don't have a good example, but wanted to put the question out there to see if you agree?

AtishPapa:
Definitely agree with you :D Couldn't come up with a nice name for this yet either.

bstien:
I've done a lot of changes in this PR/view, so I'd love for someone to do a new review of this :)

And I'd love some feedback on what we can call this view😄

thorbjxrn (approved):
Looks clean to me! I remember there were some good namings yesterday, but I think ScrollableTabView is fine

PR #1105, opened 2022-04-08
OverflowCollectionView: Offer delegate to configure cell, if needed

bstien:
# Why?
I need to re-use one cell for `OverflowCollectionView` in multiple places, but I need to configure the label dynamically. This isn't possible today, so I'm adding an optional method to the delegate protocol for this which will be called after the cell has been dequeued and configured by the view.

# What?
- Added method `overflowCollectionView<Cell>(_:didConfigureCell:atIndex:)`.

# Version Change
Minor. There is a default implementation for the delegate method.

# UI Changes
_No UI._


bstien:
Closing this PR. This method is redundant, since the cells should instead define a view model which contains the needed configuration.

PR #1106, opened 2022-04-26
Fix: Specify Bundle when registering cells

bstien:
# Why?
We have a few NIBs in AdIn, and I really want to use these extensions to clean up some code, but the default parameter for bundle is `nil`.

# What?
- Add parameter to `UITableView` and `UICollectionView` extension methods for specifying bundle.

# Version Change
Minor.

# UI Changes
_No UI._


ag12 (approved):
GJ.

hakloev (approved):

PR #1107, opened 2022-05-02
Component: SearchView

bstien:
# Why?
This view was originally in AdInsertion, but we need to move it out + use it from TJT seller.

# What?
- Added `SearchView`.
- Added demo + snapshot tests.
- Added methods to `TextField` to simplify making it the first responder.

# Version Change
Minor. Only a new view added.

# UI Changes


robinsalehjan (approved):
LGTM! 👍

PR #1108, opened 2022-05-02
MonthAndYearPickerView

hakloev:
# Why?

We need a picker view for selecting month and year in the job profile. The `UIDatePicker` provided  in `UIKit` doesn't have a `datePIckerMode` that supports only month and year. The closest alternative includes the day. 

# What?

Introduces `MonthAndYearPickerView`, a subclass of `UIControl` to be used as an inline view or as an input view for textfields. It supports setting the `locale` and `timeZone` to the internal `Calendar` instance, but defaults to `.autoupdatingCurrent`. The minimum year for when generating the possible years is set to 1900, as I thought it might be a reasonable value. Also there's support for setting a `minimumDate` and `maximumDate`. 

* Add snapshot tests

# Version Change

Minor

# UI Changes

| Before | After |
| --- | --- |
| N/A | Simulator Screen Shot - iPhone 13 Pro - 2022-05-02 at 15 33 46 |


Demo/SnapshotTests/ComponentViewTests.swift
```diff
@@ -294,4 +294,8 @@ class ComponentViewTests: XCTestCase {
     func testDetailCallout() {
         snapshot(.detailCallout)
     }
+
+    func testMonthAndYearPickerView() {
+        snapshot(.monthAndYearPickerView)
```
bstien, line 299: The snapshot image looks spicy😄 🌶️ It seems like it could fail later if it lays out properly before the test is run, so maybe we should we introduce a small delay to this test, just in case it should fail inconsistently?

Screenshot 2022-05-02 at 15 42 13

FinniversKit/Sources/Components/MonthAndYearPickerView/MonthAndYearPickerView.swift
```diff
@@ -0,0 +1,190 @@
...
+        setup()
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    private func setup() {
+        calendar.locale = locale
+        calendar.timeZone = timeZone
```
bstien, line 83: These won't be set if the `init()` is used, since `setup()` isn't called from there. Maybe `calendar` should be refactored into a lazy var instead and move these there?

bstien: Same with the lines adding and filling `pickerView` as a subview below. Do we need the `init()` at all?

bstien (approved):
LGTM!🎉 Left some comments that we should resolve before merging :)

Demo/SnapshotTests/ComponentViewTests.swift
hakloev: Yes, good point. I'll look into it :)

hakloev: No delay seems to work, tried multiple different time intervals, the snapshot still always look the same 🤷‍♂️

FinniversKit/Sources/Components/MonthAndYearPickerView/MonthAndYearPickerView.swift
hakloev: `setup` is called. The convenience initializer calls `self.init(frame:)`, not `super.init(frame:)` which in turn calls `setup` :)

FinniversKit/Sources/Components/MonthAndYearPickerView/MonthAndYearPickerView.swift
```diff
@@ -0,0 +1,190 @@
...
+        case .month:
+            return months[row]
+        case .year:
+            return String(years[row])
+        }
+    }
+
+    public func pickerView(_ pickerView: UIPickerView, didSelectRow row: Int, inComponent component: Int) {
+        let month = pickerView.selectedRow(inComponent: .month) + 1
+        let year = years[pickerView.selectedRow(inComponent: .year)]
```
robinsalehjan, line 166: What if we use the `safe` subscript here? 

```suggestion
        guard let year = years[safe: pickerView.selectedRow(inComponent: .year)] else { return }
```

robinsalehjan (approved):
A part from the minor comment, this looks good to me!

FinniversKit/Sources/Components/MonthAndYearPickerView/MonthAndYearPickerView.swift
bstien: My bad, didn't notice it was calling `self`😄

PR #1109, opened 2022-05-03
Component: CheckmarkListView

bstien:
# Why?
We need a new view for the user to opt-in to shipping.

The view can handle 1-n items and the description can be either plain text or an `NSAttributedString`.

# What?
- Added `CheckmarkListView`.
- Added demo + snapshot tests.

# Version Change
Minor. Only new view added.

# UI Changes
| 1 item | 2 items | 3 items |
| --- | --- | --- |
| Simulator Screen Recording - iPhone 13 Pro - 2022-05-03 at 12 43 06 | Simulator Screen Recording - iPhone 13 Pro - 2022-05-03 at 12 42 54 | Simulator Screen Recording - iPhone 13 Pro - 2022-05-03 at 12 42 41 |


bstien:
The snapshot image doesn't look correct, since it has all corners rounded for the first and last element. After some testing and googling I found that this [could to be an issue with the snapshotting library](https://github.com/pointfreeco/swift-snapshot-testing/issues/358).

AtishPapa (approved):
Awesome job and perfectly usable in SwiftUI! 👏 🎉

PR #1110, opened 2022-05-06
SelectionListView: Add support for radio buttons

bstien:
# Why?
In #1109 I added a view that supports a list of checkboxes for TJT shipping, but now we also need a similar view with radio buttons. Instead of creating a new view I modified the existing one.

# What?
- Rename `CheckmarkListView` to `SelectionListView`.
- Added enum to configure `SelectionListView`.
  - `.checkboxes` or `.radioButtons` are supported.
- Add optional property `identifier` for each item.
- Update delegate method to pass identifier.
- Added support for custom titles for demo items.
- Added new demo for the radiobuttons.
  - So we can have snapshot tests.

# Version Change
Breaking. View is renamed.

# UI Changes

| Radiobuttons | Checkboxes |
| :---: | :---: |
| Screenshot 2022-05-06 at 12 50 19 | Screenshot 2022-05-06 at 12 50 22 |
| Simulator Screen Recording - iPhone 13 Pro - 2022-05-06 at 12 47 10 | Simulator Screen Recording - iPhone 13 Pro - 2022-05-06 at 12 47 30 |


ninarg (approved):
LGTM! Smart to reuse it like this 👏

PR #1112, opened 2022-05-19
View: AddressComponentView

bstien:
# Why?
This view was originally in AdIn, but now we need it elsewhere.

# What?
- Added `AddressComponentView`.
  - Handles a group of field views.
- Added `AddressComponentFieldView`.
  - Shows regular items.
- Added `AddressComponentPostalFieldView`.
  - Shows postal code and place.
- Added models and icon.

`AddressComponentFieldView` and `AddressComponentPostalFieldView` are made public so I can use them in the old solution in AdIn, which handles these in another way.

# Version Change
Minor. Only new view added.

# UI Changes

Screenshot 2022-05-19 at 15 39 46


magnustviberg (approved):
LGTM 🫀

PR #1115, opened 2022-05-25
Fix: Make it possible to enable/disable button in FooterButtonView

bstien:
# Why?
I want to reuse this view, but I need to disable the button in a few scenarios.

# What?
- Added property `isEnabled` to `FooterButtonView`.

# Version Change
Patch.


AtishPapa (approved):
Awesome 🤩

PR #1116, opened 2022-05-25
Changed font size for titleLabel in PromotionView

MarenZR:
# Why?

The title on the front page promotion banner was supposed to be bigger. 

# What?

Font size of the banner title is increased to match Android and design specs. 

# Version Change

Minor: changed font size for a label.

# UI Changes

| Before | After |
| --- | --- |
|  before| Simulator Screen Shot - iPhone 13 Pro - 2022-05-25 at 14 39 51|


robinsalehjan (approved):
Awesome ❤️

PR #1117, opened 2022-05-27
Fix: Add detail items list to SelectionListView

bstien:
# Why?
The design requires an optional list of detail/checkmark items within the `SelectionListView`.

# What?
- Added subview `CheckmarkItemDetailView`.
- Added optional property to `SelectionItemModel`.
- Added checkmark icon.

# Version Change
Minor. The property added to `SelectionItemModel` is optional and has a default `nil` value in the constructor.

# UI Changes
Screenshot 2022-05-27 at 11 51 44


FinniversKit/Sources/Assets/ImageAsset.swift
```diff
@@ -56,6 +56,7 @@ enum ImageAsset: String {
     case checkCircleFilled
     case checkCircleFilledMini
     case checkmarkBig
+    case checmark
```
MarenZR, line 59: Just a spelling-error: should be checkmark ✔

MarenZR (approved):
Looks good 👍🏻

FinniversKit/Sources/Assets/ImageAsset.swift
bstien: Good catch😄 thanks!

PR #1118, opened 2022-06-02
Combine extension

AtishPapa:
# Why?

We don't have Combine extensions for UIControls. 

# What?

If someone is using Combine and wants to leverage the advantages they can subscribe to UIControl events.

```
anyButton
    .publisher(for: .touchUpInside)
    .sink { [weak self] _ in
        self?.doSomethingAwesome()
    }
    .store(in: &cancellables)
```

# Todo

There's something similar in `job-profile-ios` it might worth to merge the two somehow.

# Version Change

Minor: Only an extension to UIControl.


FinniversKit/FinniversKit.xcodeproj/project.pbxproj
```diff
@@ -2948,14 +2954,6 @@
...
 			sourceTree = "<group>";
 		};
-		DA5A858A24617DE200AC0DFE /* SwiftUI */ = {
-			isa = PBXGroup;
-			children = (
-				DA5A858B24617DF000AC0DFE /* ImageExtensions.swift */,
-			);
-			path = SwiftUI;
-			sourceTree = "<group>";
-		};
```
bstien, line 2958: Is this folder already removed from the project, or was it removed unintentionally?

bstien: Sorry, nevermind! Just noticed the file `ImageExtensions.swift` was moved in this PR 😄

bstien (approved):

FinniversKit/Sources/Extensions/Combine/CombineCompatible.swift
```diff
@@ -0,0 +1,11 @@
+// Found on https://www.avanderlee.com/swift/custom-combine-publisher.
+import Combine
+
+/// Extending the `UIControl` types to be able to produce a `UIControl.Event` publisher.
+public protocol CombineCompatible { }
+extension UIControl: CombineCompatible { }
```
robinsalehjan, line 6: Can we have some spacing here please? 😄

robinsalehjan, line 6: Can we have some spacing here please? 😄

AtishPapa: Oops 🤭 Definitely.

robinsalehjan (approved):
This is a great addition! to the library 

We take a look at the implementation in job profile and see if it makes sense to use this extension istead. I think the need and solution may be similar 😄

PR #1120, opened 2022-06-13
Component: BadgeView used for fiks ferdig promotion

ninarg:
# Why?

We already have this `BadgeView` attached to shippable items in search quest - but moving it over to FinniversKit since we want to use the same view in collections as well.

# What?

Copy + paste `BadgeView` + view model from SearchQuest. 
Make them `public`.

Add demo for it. 
Record snapshot.

# Version Change

Minor. 

# UI Changes

Skjermbilde 2022-06-13 kl  10 33 23


FinniversKit/Sources/Components/BadgeView/BadgeView.swift
```diff
@@ -0,0 +1,76 @@
+import Foundation
+import UIKit
+
+public class BadgeView: UIView {
```
BjornRuud, line 4: Can this be `final class` since it's not open for subclassing outside the framework?

FinniversKit/Sources/Components/BadgeView/BadgeView.swift
```diff
@@ -0,0 +1,76 @@
...
+    private func setup() {
+        translatesAutoresizingMaskIntoConstraints = false
+        backgroundColor = .banana
+        layer.cornerRadius = .spacingS
+        layer.maskedCorners = [.layerMaxXMaxYCorner]
+
+        addSubview(iconImageView)
+        addSubview(textLabel)
+
+        let iconSize: CGFloat = 16
```
robinsalehjan, line 39: `.spacingM` here instead?

robinsalehjan (approved):
LGTM - a part from minor comments ⭐ ⭐

FinniversKit/Sources/Components/BadgeView/BadgeView.swift
ninarg: Sure! Done ✅

PR #1124, opened 2022-06-23
Fiks ferdig promo redesign

MarenZR:
# Why?

There is a new design for the frontpage promotion banner with a background image and different constraints for the top-image (Fiks ferdig logo) depending on screen width. 

# What?

Added option for a background image to PromotionViewModel and PromotionView. Also added a case for image placement called dynamic, that changes according to screen width available. 

# Version Change

Minor.

# UI Changes

QuickTime movie

### DIFF
--- a/FinniversKit/Sources/Components/PromotionView/PromotionView.swift
+++ b/FinniversKit/Sources/Components/PromotionView/PromotionView.swift
@@ -95,15 +95,15 @@ public class PromotionView: UIView {
         imageView.bottomAnchor.constraint(equalTo: imageContainer.bottomAnchor, constant: -.spacingXXL),
         imageView.topAnchor.constraint(equalTo: imageContainer.topAnchor),
         imageView.leadingAnchor.constraint(equalTo: backgroundImageContainer.leadingAnchor, constant: .spacingM),
-        imageView.widthAnchor.constraint(lessThanOrEqualTo: imageView.heightAnchor, multiplier: imageRatio ?? 1),
-        ]
+        imageView.widthAnchor.constraint(lessThanOrEqualTo: imageView.heightAnchor, multiplier: viewModel.imageRatio),
+    ]
     private lazy var regularDynamicConstraint: [NSLayoutConstraint] = [
         imageView.bottomAnchor.constraint(equalTo: imageContainer.bottomAnchor),
         imageView.topAnchor.constraint(equalTo: imageContainer.topAnchor),
         imageView.leadingAnchor.constraint(equalTo: backgroundImageContainer.leadingAnchor, constant: .spacingM),
-        imageView.widthAnchor.constraint(lessThanOrEqualTo: imageView.heightAnchor, multiplier: imageRatio ?? 1),
-        ]
-    private var imageRatio: CGFloat?
+        imageView.widthAnchor.constraint(lessThanOrEqualTo: imageView.heightAnchor, multiplier: viewModel.imageRatio),
+    ]
+    private var viewModel: PromotionViewModel
 
     // MARK: - Public properties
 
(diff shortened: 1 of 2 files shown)


FinniversKit/Sources/Components/PromotionView/PromotionView.swift
```diff
@@ -107,33 +130,72 @@ public class PromotionView: UIView {
...
+        if previousTraitCollection?.horizontalSizeClass != traitCollection.horizontalSizeClass {
+            if isCompactScreen() {
+                activateCompactConstraints()
+            } else {
+                activateRegularConstraints()
+            }
+        }
+    }
+
+    private func isCompactScreen() -> Bool {
```
ninarg, line 196: We already have a property for this that you can use, in `UITraitCollectionExtensions.swift`. We can check `isHorizontalSizeClassRegular` instead

FinniversKit/Sources/Components/PromotionView/PromotionView.swift
```diff
@@ -107,33 +130,72 @@ public class PromotionView: UIView {
...
+
         imageContainer.addSubview(imageView)
         imageView.image = viewModel.image
 
         switch viewModel.imageAlignment {
         case .fullWidth:
             imageView.fillInSuperview(insets: UIEdgeInsets(top: .spacingM, leading: .spacingS, bottom: -.spacingM, trailing: -.spacingS))
         case .trailing:
-            let imageRatio = viewModel.image.size.width / viewModel.image.size.height
+            guard let imageRatio = imageRatio else { return }
```
ninarg, line 154: It's a bit "dangerous" this LOC depends on us remembering to set the `imageRatio` above ☝️
It used to be required and now it's optional. 
What if we make image ratio an extension of the view model, like

```
private extension PromotionViewModel {
     var imageRatio: CGFloat {
          image.size.width / image.size.height
```
The constraints above can also use the view model property, like `viewModel?.imageRatio ?? 1`, as long as we save the view model when configuring it 😄  
```
    func configure(with viewModel: PromotionViewModel) {
         self.viewModel = viewModel
```

FinniversKit/Sources/Components/PromotionView/PromotionView.swift
```diff
@@ -73,15 +73,38 @@ public class PromotionView: UIView {
...
         stackView.alignment = .leading
         return stackView
     }()
 
+    private lazy var compactDynamicConstraints: [NSLayoutConstraint] = [
+        imageView.bottomAnchor.constraint(equalTo: imageContainer.bottomAnchor, constant: -.spacingXXL),
+        imageView.topAnchor.constraint(equalTo: imageContainer.topAnchor),
+        imageView.leadingAnchor.constraint(equalTo: backgroundImageContainer.leadingAnchor, constant: .spacingM),
+        imageView.widthAnchor.constraint(lessThanOrEqualTo: imageView.heightAnchor, multiplier: imageRatio ?? 1),
+        ]
```
ninarg, line 99: too much indentation on` ]`
same below 👇

ninarg (commented):
Great work 🥇 Added a few comments.

FinniversKit/Sources/Components/PromotionView/PromotionView.swift
MarenZR: Nice to know :)

FinniversKit/Sources/Components/PromotionView/PromotionView.swift
MarenZR: Cool 👍

MarenZR: Actually the viewModel can be saved on init of the view, so it doesn't even have to be optional. Much better.

ninarg: perfect ❤️

ninarg (approved):
Great work 🔥

PR #1125, opened 2022-07-01
Fix: Size bug with image gallery

bstien:
# Why?
The full screen image gallery had an issue where the image you're swiping to would "jump" into the correct size when the swipe is finished. This is mostly visible on iPad, especially when multitasking.

The issue seems to be that the size calculation of the view happens before it's frame is set properly, so it takes the whole width of the screen before "jumping" into the correct size once the frame is set and the size of the view is recalculated.

# What?
- Pass the size of the container to the VC handling each full screen image.

# Version Change
Patch. No API changes.

# UI Changes
Sorry for the 2 FPS gifs 😅 The issue is still visible, even though it hurts to watch.

| Before | After |
| --- | --- |
| Simulator Screen Recording - iPad Air (5th generation) - 2022-07-01 at 14 33 48 | Simulator Screen Recording - iPad Air (5th generation) - 2022-07-01 at 14 32 57 |


thorbjxrn (approved):
Good stuff 🖼

MarenZR (approved):
LGTM

PR #1128, opened 2022-07-05
Fix: `FooterButtonView` button init issue

bstien:
# Why?
Thanks to @MarenZR who, unfortunately, experienced the issue of being stuck on the shipping page we managed to track down the root issue.

The button within `FooterButtonView` was initialized as a `let` and not as a `lazy var`. Not sure how this was missed earlier, and why Xcode doesn't provide a warning, but AppCode managed to provide it.

Screenshot 2022-07-05 at 11 52 40

# What?
- Replace `let` with `lazy var`, since we use a reference to `self` for adding a target.

# Version Change
Patch.


MarenZR (approved):
🎖😄

PR #1131, opened 2022-07-11
ObjectPageTitleView: Set accessibility elements

bstien:
# Why?
This view has a bunch of labels, but it's not as easy as it should be to find this view when using VoiceOver.

# What?
- Define the list/order of `accessibilityElements`, having the `titleLabel` first and giving it the `header` trait.
  - I placed `ribbonView` last in this list, I think this makes sense.
- Set all other labels to `staticText`.

# Version Change
Patch. No API changes.

# UI Changes
| Before | After |
| --- | --- |
| Screenshot 2022-07-11 at 12 12 06 | Screenshot 2022-07-11 at 12 11 15 |


bstien:
The `titleLabel` and `subtitleLabel` had also the trait of `button` before, since we add a `UILongPressGestureRecognizer` to it to make the text copyable.

I'm not sure if this is needed or not, but what do you think? We have the option to add it back.

robinsalehjan (approved):

PR #1138, opened 2022-08-18
Tjt object page

AtishPapa:
# Why?

The TJT related object page changes required parsing `Label` text from HTML.

# What?

- The class `Label` was extended with `setText(fromHTMLString...` method.
- new label style was added 
- new named colors were added

# Version Change

Minor


FinniversKit/Sources/Components/Label/Label.swift
```diff
@@ -74,3 +74,28 @@ extension Label {
         UIMenuController.shared.setMenuVisible(true, animated: true)
     }
 }
+
+// MARK: - Support HTML
+
+public extension Label {
+    func setText(fromHTMLString htmlString: String, style: [String: String] = [:]) {
+        var styledText = htmlString
+        for (itemIndex, styleItem) in style {
```
bstien, line 83: `itemIndex` is not an index, so for clarity I think this could be renamed to i.e. `stylePlaceholder`, `styleIdentifier` or something similar

FinniversKit/Sources/Components/Label/Label.swift
```diff
@@ -74,3 +74,28 @@ extension Label {
...
+            styledText = styledText.replacingOccurrences(of: itemIndex, with: styleItem)
+        }
+        let htmlTemplate = "<span style=\"font-family: \(font!.fontName); font-size: \(font!.pointSize); color: \(textColor.hexString)\">\(styledText)</span>"
```
bstien, line 86: We shouldn't need to force unwrap `font` for `Label`/`UILabel`, so I think this should be removed

bstien (approved):
LGTM!

PR #1141, opened 2022-08-25
Feature/tmp 1058 update tjm entrypoint

thorbjxrn:
# Why?

A new design is given for the ContractActionView / MotorTransactionJourney entry point

> PR removed and re-added as I've done a slightly odd way of rebasing

# What?

- Remove ContractActionView video
- Make bullet points optional
- New layout

# Version Change

minor

# UI Changes

| Before | After |
| --- | --- |
| Screenshot 2022-08-18 at 16 16 03 Screenshot 2022-08-18 at 16 14 13 |   Screenshot 2022-08-18 at 16 10 11  |

New look as seen in the Finn app:
Screenshot 2022-08-18 at 16 09 20


FinniversKit/Sources/Assets/Icons.xcassets/MotorTransaction/carFront.imageset/Contents.json
```diff
@@ -0,0 +1,22 @@
...
+      "filename" : "Size=32.svg",
+      "idiom" : "universal",
+      "scale" : "2x"
+    },
+    {
+      "idiom" : "universal",
+      "scale" : "3x"
+    }
+  ],
+  "info" : {
```
bstien, line 18: I'm not completely sure, but it doesn't seem like this image _(added to FinniversKit and not the demo)_ is used anywhere so it can probably be removed :)

bstien (approved):
Great job😄🎉 I know we use this view in other places also (i.e. realestate-rental) and it seems this view will still support that with these changes 🔥

PR #1144, opened 2022-09-26
Support Posten opt-in

AtishPapa:
# Why?

We need to support multiple selection of shipping providers during ad-in opt-in.

# What?

- The selection list can display dynamically sized icons.
- a `HTMLLabel` class was added
- added a basic `HTMLStyler` implementation for supporting custom tags (at the moment it only supports `UIColor` values)
- added `ViewModelCoordinator` classes to help separating the navigation from the view models

# Version Change

Breaking: `setText(fromHTMLString...` in `Label` was removed.

# UI Changes
| Fixed images | Dynamic images |
| --- | --- |
| Screenshot 2022-09-26 at 17 17 03 | Screenshot 2022-09-26 at 17 15 03 |


FinniversKit/Sources/Base Components/ViewModelCoordinator/AnyViewModelCoordinator.swift
```diff
@@ -0,0 +1,31 @@
...
+    }
+}
+
+// MARK: - AnyViewModelCoordinator
+
+public final class AnyViewModelCoordinator<Route>: ViewModelCoordinator {
+
+    // MARK: - Properties
+
+    internal let _handle: (Route) -> ()
```
BjornRuud, line 15: The access specifier should be enough for access semantics, there is no need for underscore prefix on the name.

AtishPapa: hehe, yeah 😉 it's a copy-pasted code 😛

FinniversKit/Sources/Base Components/ViewModelCoordinator/UnownedViewModelCoordinator.swift
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public func eraseToUnownedCoordinator<TargetRoute>(transform: @escaping (TargetRoute) -> Route) -> UnownedViewModelCoordinator<TargetRoute> {
+        return .init(coordinator: self, transform: transform)
+    }
+}
+
+// MARK: - UnownedCoordinator
+
+public final class UnownedViewModelCoordinator<Route>: ViewModelCoordinator {
```
BjornRuud, line 15: When is the unowned version needed? How is the lifetime of the wrapped coordinator ensured to be as long as the unowned coordinator?

AtishPapa: It depends on the use case. I didn't needed this, but there might be cases later when you want to avoid circular dependencies and this could help someone. 😉

FinniversKit/Sources/Components/HTMLLabel/HTMLLabel.swift
```diff
@@ -0,0 +1,63 @@
+public class HTMLLabel: Label {
```
BjornRuud, line 1: Nice! I need this for my branch. 😄

FinniversKit/Sources/Components/HTMLLabel/HTMLLabel.swift
```diff
@@ -0,0 +1,63 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    public func setHTMLText(_ htmlText: String, with additionalStyleMap: HTMLStyler.StyleMap = [:]) {
+        self.htmlText = htmlText
+        self.additionalStyleMap = additionalStyleMap
+        setAttributedString(from: htmlText)
+    }
+
+    public override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
```
BjornRuud, line 27: Could you add a comment why this override is needed?

FinniversKit/Sources/Components/HTMLLabel/HTMLLabel.swift
```diff
@@ -0,0 +1,66 @@
...
+    public override var textColor: UIColor! {
+        didSet {
+            styleMap[.textColor] = textColor
+        }
+    }
+
+    public init(style: Style, styleMap: HTMLStyler.StyleMap = [:], withAutoLayout: Bool = false) {
+        self.styleMap = styleMap
+        super.init(style: style, withAutoLayout: withAutoLayout)
+    }
```
bstien, line 15: You set `numberOfLines` to `0` everywhere this view is used. Should it maybe just be set here for this view?

AtishPapa: Not sure about it 🤔 It's still a `Label` and we don't set it there either...

bstien: That's true. But now that we have this init, how about adding it as an param here instead with a default value of either 0 or 1, then?

```swift
public init(
    style: Style,
    styleMap: HTMLStyler.StyleMap = [:],
    numberOfLines: Int = 0,
    withAutoLayout: Bool = false
) {
    self.styleMap = styleMap
    super.init(style: style, withAutoLayout: withAutoLayout)
    self.numberOfLines = numberOfLines
}
```

FinniversKit/Sources/Components/SelectionListView/SelectionListView.swift
```diff
@@ -20,6 +20,14 @@ public class SelectionListView: UIView {
         itemViews.contains(where: { $0.isSelected })
     }
 
+    public var selectedItems: [(index: Int, identifier: String?)] {
+        itemViews.enumerated().map { index, itemView in
+            guard itemView.isSelected else { return nil }
+
+            return (index: index, identifier: itemView.model.identifier)
+        }.compactMap { $0 }
```
bstien, line 28: This could be simplified:

```swift
itemViews.enumerated().compactMap { index, itemView in
    guard itemView.isSelected else { return nil }

    return (index: index, identifier: itemView.model.identifier)
}
```

FinniversKit/Sources/Components/Label/Label.swift
```diff
@@ -74,28 +74,3 @@ extension Label {
         UIMenuController.shared.setMenuVisible(true, animated: true)
     }
 }
-
-// MARK: - Support HTML
-
-public extension Label {
-    func setText(fromHTMLString htmlString: String, style: [String: String] = [:]) {
-        var styledText = htmlString
```
bstien, line 82: Just a heads up: removing this method makes this a breaking change, and not a minor bump.

PR #1149, opened 2022-10-12
Component: MyAdsListView & rewrite ScrollableTabView

bstien:
# Why?
We have a redesign ongoing for the list of user's ads.

# What?
I also needed/wanted to see how the table would behave with the diffable data source, since it's one of the first times I've implemented it, so I've added a protocol to the demo that lets demo views provide `UITabBarItem`s to `DemoViewController`. These will be presented as long as it's contained within a `UINavigationController`.

- Add `MyAdsListView` + subviews.
- Add `MyAdModel`.
- Add optional parameters to `Label` init.
- Set content compression resistance to `.required` on `RibbonView`.
- Add protocol `BarButtonProvider` to demo.
- Rewrite `ScrollableTabView` to use `UIScrollView` instead of `UIStackView`.

# Version Change
Minor. Only new view and optional parameters with default values added.

# UI Changes

| iPhone | iPad |
| --- | --- |
| Screenshot 2022-10-12 at 15 29 23 | Screenshot 2022-10-12 at 15 28 01 |

| Moving/reordering items | Loading cell |
| --- | --- |
| RocketSim_Recording_iPhone_13_Pro_2022-10-12_15 29 29 | RocketSim_Recording_iPhone_13_Pro_2022-10-17_12 59 36 |

### DIFF
--- a/Demo/SnapshotTests/Info.plist
+++ b/Demo/SnapshotTests/Info.plist
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
 	<key>CFBundleShortVersionString</key>
-	<string>95.0.0</string>
+	<string>95.0.3</string>
 	<key>CFBundleVersion</key>
 	<string>1</string>
 	<key>NSPrincipalClass</key>
--- a/Demo/Sources/Info.plist
+++ b/Demo/Sources/Info.plist
@@ -21,7 +21,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>95.0.0</string>
+	<string>95.0.3</string>
 	<key>CFBundleVersion</key>
 	<string>1</string>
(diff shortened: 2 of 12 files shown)


FinniversKit/Sources/Components/MyAdsList/Models/MyAdModel.swift
```diff
@@ -0,0 +1,36 @@
...
+        self.subtitle = subtitle
+        self.imageUrl = imageUrl
+        self.expires = expires
+        self.numFavorites = numFavorites
+        self.numViews = numViews
+        self.ribbon = ribbon
+    }
+
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(adId)
```
BjornRuud, line 34: This will lead to hash collisions if other types of models also only hash the adId and they have the same adId. I don't know if that has any consequence for this use case, or is intended behaviour?

bstien: Yes, it's intended 🙂 Ads will only be represented once in the list of models we provide this view. If we need to pass other models, or if there'll be duplicates, we'll have to change this.

PR #1153, opened 2022-11-03
Fix: Avoid duplicate labels on maps (from Apple map) when building with Xcode 14

ninarg:
# Why?

When building the project with Xcode 14, the Apple map labels suddenly started overlapping our own tiles, thus we ended up with duplicate labels.

# What?

Specifying the tile should be placed `.aboveLabels` fixes the issue.

# Version Change

Patch

# UI Changes

For instance, see extra "Lillestrøm" label. Double "Gjerdrum" label.

| Before | After |
| --- | --- |
| Simulator Screen Shot - iPhone 13 Pro - 2022-11-03 at 10 34 53 | Simulator Screen Shot - iPhone 13 Pro - 2022-11-03 at 10 40 13 |


osanoj (approved):
🎉

PR #1154, opened 2022-11-04
Fix: Make radius overlay on map visible when building with Xcode 14

ninarg:
# Why?

@MarenZR found another bug introduced when building with Xcode 14. (Great catch 💯) 
Suddenly, the map radius overlay for addresses is not visible anymore.

# What?

When zooming in and out, we can cleary see the radius is placed below the map tiles.
Since tiles are added `.aboveLabels`, it can be fixed by also adding the radius overlay `.aboveLabels`. 

# Version Change

Patch

# UI Changes

| Before | After |
| --- | --- |
| Simulator Screen Shot - iPhone 13 Pro - 2022-11-04 at 09 17 20 | Simulator Screen Shot - iPhone 13 Pro - 2022-11-04 at 09 19 40 |


AtishPapa (approved):
LGTM 👍

PR #1158, opened 2022-11-09
Update snapshot library and fix snapshots that fail on Intel

osanoj:
# Why?
There was a couple of snapshots that crashed on Intel machines when building with Xcode 14.

# What?
- Updated snapshot lib version, this did not seem to help, but is still something we want to do 😄 
- Fix the already existing fixes for map snapshots being inconsistent, there was an issue related to alpha values in the crash so also hiding the mapView instead of just covering it seemed to fix that
- Specify to always use Norwegian locale for snapshot target, to avoid inconsistencies in month+year picker 🤞 

Also not that this is a PR to another PR #1157 

# Version Change
No changes for anyone using FinniversKit library.


ninarg (approved):
LGTM! Great work 🥇  Hopefully this also works on M1 🤞

osanoj:
I'll merge this into the #1157 PR now @ninarg so we can continue with this there if there are any more changes needed 🙂

PR #1159, opened 2022-11-10
Update cat icon

ninarg:
# Why?

Improving the cat icon by adding scaled images.

# What?

Upload icons in 3 different sizes.

# Version Change

Patch.

# UI Changes

Should barely be visible.


BjornRuud (approved):
👍

PR #1167, opened 2022-12-13
My ads: Improve accessibility

bstien:
# Why?
The accessibility support in my ads was pretty bad, so this is an attempt to improve it.

- The buttons within the top bar that let's you select facet were just a list of strings and they didn't mention which one was selected.
- The list of my ads only presented the value for each statistic (clicks, favorites) and not what they mean.
- The label containing number of days left on the ad didn't read out the number of days.

# What?
Fix these 👆 

# Version Change
Breaking. Changing the properties contained within `MyAdModel`.

# UI Changes
| What | Before | After |
| :---: | --- | --- |
| My ads | Screenshot 2022-12-13 at 12 34 59 | Screenshot 2022-12-13 at 12 32 48 |
| Selection | Screenshot 2022-12-13 at 12 34 46 | Screenshot 2022-12-13 at 12 33 52 |


BjornRuud (approved):
Nice work! 👍 I'm never quite sure how to handle items with many labels. Should they be separate or should some be joined together for voiceover? Feedback will tell.

bstien:
> Nice work! 👍 I'm never quite sure how to handle items with many labels. Should they be separate or should some be joined together for voiceover? Feedback will tell.

Thanks! I'm not really sure myself 🤔  but I think it could make sense to break them up so you can scroll to the item you want instead of having to wait for it to read out what you're looking for.

I think we could work together with the UU group to get some insight into this and to get our questions answered.

PR #1170, opened 2022-12-22
Apps 4580 braze cc torget

MarenZR:
# Why?

There were some suggestions for code refactoring for my last braze content card PR which I did not have time to implement, these are the changes based on those suggestions. Also marketing wanted some spacing changes.

# What?

Cleanup and refactoring.

# Version Change
Patch

# UI Changes

| Before | After |
| --- | --- |
| _Simulator Screen Shot - iPhone 14 Pro - 2022-11-24 at 13 38 06_ | _Simulator Screen Shot - iPhone 14 Pro - 2022-12-22 at 10 10 56_ |


FinniversKit/Sources/Components/BrazePromotionView/BrazePromotionView.swift
```diff
@@ -151,29 +151,33 @@ public class BrazePromotionView: UIView {
     }
 
     private func loadImage() {
-        guard let imageUrl = viewModel.image else {
-            remoteImageView.removeFromSuperview()
-            NSLayoutConstraint.deactivate(stackViewConstraintsImage)
-            NSLayoutConstraint.activate(stackViewConstraintsNoImage)
+        if let imageUrl = viewModel.image {
+            closeButton.setImage(UIImage(named: .cross), for: .normal)
+            closeButton.backgroundColor = UIColor.black.withAlphaComponent(0.6)
```
AndreyMomot, line 156: Is dark mode tested with this?

AndreyMomot (approved):
LGTM except small clarification

FinniversKit/Sources/Components/BrazePromotionView/BrazePromotionView.swift
MarenZR: Yes, the color is only black when there is an image there, so that's why the color is the same for light and dark mode here. Good observation 😊

PR #1172, opened 2022-12-22
Improve accessibility support in frontpage

osanoj:
# Why?
We should get better at accessibility. This PR mainly focuses on VoiceOver for the frontpage.

# What?
- Added .header trait on the labels that are headers on the frontpage.
- Added an accessibility only header for the markets grid.
- Added button/link trait on the markets.
- Improved recommendation cells so the favorite button is there for VoiceOver users.

# Version Change
This is a major change since I added the need for a header text for the markets grid.

# UI Changes
It should not be a visual change, but for VoiceOver it will work like this (audio on to hear VoiceOver):
https://user-images.githubusercontent.com/3169203/209131864-8346f52b-5bc4-4650-ae06-0cc3c9874400.mov


FinniversKit/Sources/Recycling/GridViews/AdRecommendations/Cell/JobAdRecommendationCell.swift
```diff
@@ -34,6 +34,12 @@ public class JobAdRecommendationCell: UICollectionViewCell, AdRecommendationCell
 
     private lazy var defaultImage: UIImage = UIImage(named: .noImage)
 
+    /// Extra container for accessibility issues. The cell should have "all content" and favoriteButton as accessibilty elements but it get confused if favoriteButton is a subview of the other accessibility element (so contentView can not be one of the accessibility elements
+    private lazy var containerView: UIView = {
```
BjornRuud, line 38: Is it possible to drop this if you define your own accessibility elements for the view?

BjornRuud (approved):
Good stuff! 👍 Great use of headers.

PR #1173, opened 2022-12-22
Update colors to latest Fabric version.

BjornRuud:
# Why?

The Fabric color scheme has been updated.

Note: The snapshot tests have not been updated yet to keep the amount of changes small for review.

# What?

- Added all Fabric colors.
- Updated Finn colors to use the mappings from Fabric.
- Added missing colors in various extensions.

# Version Change

Minor. Will break snapshot tests as some colors are slightly different.


BjornRuud:
It appears the snapshot tests ran fine on CircleCI, but many fail locally for me. Looks like a problem on my end.

MarenZR (approved):
It makes me so happy to see this all organized and alphabetized 🤩 🥇 Great job!

PR #1174, opened 2022-12-29
Feature/apps 4687 give feedback

AndreyMomot:
# Why?

- Deliver Trust NMP FeedbackInput component which uses TextView and it should be UI-configured from inside Trust SDK
- only Trust SDK uses StepIndicator & StepDots, so no need to store it in FinniverseKit

# What?

- Make TextView UI-configurable, some small fixes
- Move StepIndicator & StepDots out from FinniverseKit

# Version Change

minor


ag12 (approved):
👍

PR #1175, opened 2023-01-03
Fix crash if components.count < 3

AndreyMomot:
# Why?

There is a crash if components.count < 3

# What?

Safe extract values from array

# Version Change

minor

# UI Changes

| Before | After |
| --- | --- |
| crash| no_crash|


BjornRuud:
How can there be less than 3 components when using RGB color space? 🤔

AndreyMomot:
Screenshot 2023-01-03 at 14 22 28
Good question... 
Try to open a torget ad in debug with id: 279680157

BjornRuud (approved):

PR #1178, opened 2023-01-04
Small changes supporting the disclaimer update

AtishPapa:
# Why?

Some changes were necessary to make the dynamic disclaimer feature possible.

# What?

- a `struct` was converted to `class`.
- added `text` to `HyperLinkTextView`.

# Version Change

Breaking: a `struct` was converted to a `class`.


BjornRuud (approved):
👍

PR #1179, opened 2023-01-05
Missing dark bgcolor

BjornRuud:
# Why?

Some components were using Bluegray700 as dark background color, but it was defined as a custom color with a different value than the one in Fabric. When colors were updated to match Fabric these components got a slightly lighter background. After disussion with the design team it was decided to use the original color which corresponds with `darkBgPrimaryProminent` on Android.

# What?

Changed instances of `blueGray700` to `darkBgPrimaryProminent` in custom dark colors in components.

# Version Change

Patch.

# UI Changes

| Before | After |
| --- | --- |
| Simulator Screen Shot - iPhone 14 Pro - 2023-01-05 at 13 40 31 | Simulator Screen Shot - iPhone 14 Pro - 2023-01-05 at 13 39 42 |

### DIFF
--- a/FinniversKit/Sources/Components/BrazePromotionView/BrazePromotionView.swift
+++ b/FinniversKit/Sources/Components/BrazePromotionView/BrazePromotionView.swift
@@ -232,7 +232,7 @@ private extension UIColor {
     }
 
     static var bgColor: UIColor {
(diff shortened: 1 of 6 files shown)


ninarg (approved):
Great work 👍 💯

PR #1183, opened 2023-01-12
Accessibility improvements

thorbjxrn:
# Why?

Accessibility changes so low vision users will have an easier time.

# What?

Single column for frontpage scrolling, and adjusted label sizes. Double column for iPad, scaled down from the 3 default columns

# Version Change

minor

# UI Changes

this video shows old and new layouts. (Includes changes to FinnUI as well.)

https://user-images.githubusercontent.com/17337441/212094921-de66406a-fee4-4e92-b167-70d151b8ef90.mov


BjornRuud (approved):
Nice improvement! 👍 We need to test a lot to ensure no unintended layout side effects.

osanoj:
Sory for being late to the party, @thorbjxrn I think you should check the `Config.isDynamicTypeEnabled` before doing this, Otherwise you will get the column change without getting bigger text 😅

PR #1184, opened 2023-01-13
Accessibility improvements

thorbjxrn:
# Why?

Additional check for Config.isDynamicTypeEnabled, so we dont get a weird half-change when the option is disabled in config.

this is a addon to Accessibility improvements #1183

# What?

added a conditional value for Config.isDynamicTypeEnabled

# Version Change

patch

### DIFF
--- a/FinniversKit/Sources/Recycling/GridViews/AdRecommendations/Cell/StandardAdRecommendationCell.swift
+++ b/FinniversKit/Sources/Recycling/GridViews/AdRecommendations/Cell/StandardAdRecommendationCell.swift
@@ -213,24 +213,6 @@ public class StandardAdRecommendationCell: UICollectionViewCell, AdRecommendatio
         imageHeightMinimumConstraint.priority = .defaultHigh
 
         containerView.fillInSuperview()
-        let traitCollectionContentSize = {
-            switch self.traitCollection.preferredContentSizeCategory {
-            case UIContentSizeCategory.accessibilityExtraExtraExtraLarge:
-                return 2.5
-            case UIContentSizeCategory.accessibilityExtraExtraLarge:
-                return 2.25
-            case UIContentSizeCategory.accessibilityExtraLarge:
-                return 2.0
-            case UIContentSizeCategory.accessibilityLarge:
-                return 1.75
-            case UIContentSizeCategory.accessibilityMedium:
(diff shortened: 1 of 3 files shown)


FinniversKit/Sources/Recycling/GridViews/AdRecommendations/Cell/StandardAdRecommendationCell.swift
```diff
@@ -165,7 +165,7 @@ public class StandardAdRecommendationCell: UICollectionViewCell, AdRecommendatio
 
     private func setup() {
         isAccessibilityElement = true
-        let accessibilityMultiplier: CGFloat = {
+        let accessibilityMultiplier: CGFloat = Config.isDynamicTypeEnabled ? {
```
osanoj, line 168: I think it might be a bit to complicated to read like this 😅 

How about simplifying with something like this
```
        var accessibilityMultiplier = 0.0
        if Config.isDynamicTypeEnabled {
            ...
        }
```

FinniversKit/Sources/Recycling/GridViews/AdRecommendations/GridView/AdRecommendationsGridView.swift
```diff
@@ -234,7 +234,7 @@ extension AdRecommendationsGridView: AdRecommendationsGridViewLayoutDelegate {
         case .columns(let columns) where columns > 1 && columns <= 3:
             return columns
         default:
-            if traitCollection.preferredContentSizeCategory.isAccessibilityCategory {
+            if traitCollection.preferredContentSizeCategory.isAccessibilityCategory && Config.isDynamicTypeEnabled {
```
osanoj, line 237: Hmm, what do you think about a rewrite like this?
```
default:
   var columns = traitCollection.horizontalSizeClass == .regular ? 3 : 2
   if traitCollection.preferredContentSizeCategory.isAccessibilityCategory && Config.isDynamicTypeEnabled {
      columns -= 1
   }
   return columns
```

thorbjxrn: Looks smooth 👍

osanoj (approved):

PR #1185, opened 2023-01-16
Component: ViewingsRedesignView + refactor current view

bstien:
# Why?
For the realestate homes redesign we have a new design for the viewings component. I'm not sure, but I think the plan is to replace all usages of the existing view with this one going forward.

# What?
This view will reuse the views for the viewing items, so I've refactored this to work in both contexts.

- Add `ViewingsRedesignView`.
- Rename `ViewingsView` to `ViewingsListView`.
- Refactor `ViewingItemView` to work for both current view and redesign.
- Add optional init params for `UIStackView`.

# Version Change
Breaking. Renaming view.

# UI Changes

| Dark mode | Light mode |
| --- | --- |
| Screenshot 2023-01-13 at 17 43 09 | Screenshot 2023-01-13 at 17 43 07 |


AndreyMomot (approved):
Tests :heart:

PR #1188, opened 2023-01-19
Delete old search views

ninarg:
# Why?

We have deleted old search, so these views aren't used anymore.

# What?

- Delete SearchResultMapView
- Delete SearchListEmptyView
- Delete SaveSearchPromptView
- Delete icons used in them
the latter two don't have a demo for some reason, so didn't have to delete those.

# Version Change

No need to bump version, since they're not used anymore in the main app.


ag12 (approved):
👍

PR #1189, opened 2023-01-20
Support badges in recommendations

ninarg:
# Why?

We want to show "Fiks ferdig" badge in recommendations.

# What?

Add `BadgeView` to `StandardAdRecommendationCell` and view model.
Attach it to top leading anchor in image.
Add it to the demo as well, rerecorded snapshots.

# Version Change

Breaking, new field in view model protocol.

# UI Changes

| Front page recommendations | Torget recommendations |
| --- | --- |
| Simulator Screen Shot - iPhone 14 Pro - 2023-01-20 at 14 09 37 | Simulator Screen Shot - iPhone 14 Pro - 2023-01-20 at 14 09 53 |


Demo/Sources/Recycling/GridViews/AdRecommendations/AdRecommendationsGridViewDemoViewHelpers.swift
```diff
@@ -110,7 +112,9 @@ struct AdFactory {
                 scaleImageToFillView: scaleImageToFillView,
                 adType: .normal,
                 sponsoredAdData: index % 4 == 0 ? sponsoredAdData : nil,
-                favoriteButtonAccessibilityLabel: "Sett annonsen som favoritt")
+                badgeViewModel: index == 1 ? badgeViewModel : nil,
```
BjornRuud, line 115: Is it possible to not reference the numerical index? Seems fragile. I would also recommend moving the check to where the model is created to avoid allocating a model when not necessary.

BjornRuud (approved):
Minor comment, looks good! 👍

Demo/Sources/Recycling/GridViews/AdRecommendations/AdRecommendationsGridViewDemoViewHelpers.swift
ninarg: I moved the view model out so we create it only once. ✅ 
When it comes to the index, this is demo code and we already make many assumptions about indexes in this code. So it is possible, but maybe not necessary. Then we'd have to change how the entire demo is built. 😄

PR #1190, opened 2023-01-23
Refactor HTMLLabel

AtishPapa:
# Why?

The current implementation of `HTMLLabel` depends on a solution which dispatches to WebKit when creating the HTML string, causing crashes and layout issues.

# What?

The inner part of the label was remade to use the lexer, also the UIKit based parser was refactored to return an attributed string. Since the basic HTML elements should be the same, currently the custom styling is narrowed to the `style` tag.

# Version Change

Breaking: The styling interface of HTMLLabel changed.


FinniversKit/Sources/Components/HTMLLabel/HTMLLabel.swift
```diff
@@ -1,26 +1,31 @@
 public class HTMLLabel: Label {
     private var htmlText: String?
-    private var styleMap: HTMLStyler.StyleMap
-    private var additionalStyleMap: HTMLStyler.StyleMap?
+    private let spanMapper: HTMLStringUIKitStyleTranslator.SpanMapper
```
BjornRuud, line 3: In the future this class should probably be modified to accept custom mapping for any element. We have no idea what kinds of HTML the various backends want to send.

BjornRuud: There are various weights of bold, and in CSS you can style elements however you want regardless of their original intent. We need a default for the common semantic tags though.

BjornRuud (approved):
LGTM! Nice tests for attributed string output. 👍

FinniversKit/Sources/Components/HTMLLabel/HTMLLabel.swift
AtishPapa: That's fine, but until we don't get that sophisticated HTML syntax from backend, I think it's simpler to only open up `span` at the moment.

PR #1195, opened 2023-02-01
Refactor: Prepare for realestate redesign

bstien:
# Why?
There are a few changes that needs to be done in order to implement the new redesign.

# What?
- Add new properties to `CollapsibleContentView.Style`.
- Make it possible to do simple styling for items in `KeyValueGridView`.

# Version Change
Patch. Public types have optional parameters.


FinniversKit/Sources/Components/KeyValueGridView/KeyValuePair.swift
```diff
@@ -1,11 +1,32 @@
...
+    public let title: String
+    public let value: String
+    public let accessibilityLabel: String?
+    public let valueStyle: Style?
 
-    public init(title: String, value: String, accessibilityLabel: String? = nil) {
+    public init(
+        title: String,
+        value: String,
+        accessibilityLabel: String? = nil,
```
MarenZR, line 10: Are we supposed to get this from backend? Could we potentially default to something rather than nil?

MarenZR (approved):
LGTM

PR #1196, opened 2023-02-01
[Don't merge] Snapshot issues

bstien:
# Why?
I'm using this branch to test snapshot issues in #1195.

# What's tested?
| What | Does test succeed? |
| --- | :---: |
| Remove snapshot reference image | ❌ |
| Add a fullscreen black cover to `BadgeDemoView` | ✅ |
| Retake snapshot on iPhone 13 Max Pro, instead of iPhone X | ❌ |


bstien:
The issue is identified and fixed in #1197

PR #1197, opened 2023-02-02
Revert to fork of `swift-snapshot-testing`

bstien:
# Why?
CircleCI doesn't have any GPUs, which is necessary for the `perceptualPrecision` diffing that was [added to the lib late last year](https://github.com/pointfreeco/swift-snapshot-testing/pull/628) since that method uses Metal.

# What?
- Revert to @pimms [fork of `swift-snapshot-testing`](https://github.com/pimms/swift-snapshot-testing)
- Use `subpixelThreshold`.
- Retake all snapshot images.

# Version Change
Minor. No public changes.

# UI Changes
_None._


bstien:
There is an open PR that adds support for non-Metal perceptual diffing that we can follow.

https://github.com/pointfreeco/swift-snapshot-testing/pull/666

osanoj (approved):
I have compared the new snapshots pixel by pixel and there were no unexpected changes 😉

adriansergheev:
https://github.com/RevenueCat/purchases-ios/pull/3526
looks like CircleCI has added m1 instances

bstien:
Yes, but we need to pay for those. If we get it approved to pay for it, and we also have access to GPU, we can try to switch.

That won't help us on Travis, though, where I believe we don't have any GPUs available.

PR #1198, opened 2023-02-02
AddressMapView: Overlay polygons above labels

bstien:
# Why?
In order to show an area outline (i.e. for postcodes) we need to add the polygons above labels.

# Version Change
Patch.


ninarg (approved):
👌

PR #1194, opened 2023-02-01
SwiftUI Toasts

ninarg:
# Why?

We need to be able to show toasts from our SwiftUI views ✨ 

# What?

- Renamed Demo project's `State` struct to `DemoState`, since it was conflicting with SwiftUI `@State `
- Implement a new SwiftUI version of our already existing UIKit `ToastView`, with the same capabilities
   - With or without action buttons
   - Success and error styles
   - Present from top or bottom of screen
   - Swipe up or down to dismiss
   - Dismisses itself after a timeout, default 5 seconds
   - It's possible to present them from any SwiftUI view using a view modifier 

# Version Change

Minor.

# UI Changes

Screen Recording 2023-02-01 at 12 59 03

PR #1199, opened 2023-02-03
Use view model as binding for showing toasts, instead of an extra isShowing bool

ninarg:
# Why?

Made an improvement to the toast view modifier, where we use a binding for the view model instead of an extra boolean indicating whether the toast should be shown.

# What?

When you set a `Toast.ViewModel` on the view, the toast will show up, when it's set to nil, it will disappear. It gives us much more flexibility than adding a bunch of required view models where we control whether they're shown with booleans. 
Remove `isShowing` boolean.

# Version Change

Merging this into #1194


BjornRuud (approved):
Clean! 👍

PR #1200, opened 2023-02-06
SwiftUI loading view

BjornRuud:
# Why?

Ongoing work of creating SwiftUI versions of existing basic UIKit views.

# What?

SwiftUI implementation of LoadingView. It has fullscreen and boxed mode, can be used as a standalone view, and has a view modifier to overlay it on any view.

# Version Change

Minor, just addition.

# UI Changes

| | Fullscreen | Boxed |
| --- | --- | --- |
| Light | Simulator Screen Shot - iPhone 14 Pro - 2023-02-06 at 14 45 38 | Simulator Screen Shot - iPhone 14 Pro - 2023-02-06 at 14 45 47 |
| Dark | Simulator Screen Shot - iPhone 14 Pro - 2023-02-06 at 14 46 01 | Simulator Screen Shot - iPhone 14 Pro - 2023-02-06 at 14 46 09 |


ninarg (approved):
Awesome, great work 🎉 
The snapshot doesn't actually capture the loading view. But guess it's difficult to snapshot anyways, since it's animating 🤔

BjornRuud:
> Awesome, great work 🎉 The snapshot doesn't actually capture the loading view. But guess it's difficult to snapshot anyways, since it's animating 🤔

Yes I don't know how to snapshot this with correct content, and we need the snapshot to avoid the snapshot missing error. Any thoughts @osanoj @bstien ?

bstien:
> > Awesome, great work 🎉 The snapshot doesn't actually capture the loading view. But guess it's difficult to snapshot anyways, since it's animating 🤔
> 
> Yes I don't know how to snapshot this with correct content, and we need the snapshot to avoid the snapshot missing error. Any thoughts @osanoj @bstien ?

If we could snapshot this without it animating, like we have today, that would be the best. In the current demo we do it like shown below. Do we have that option with this view?
| Delay the animation | Snapshot |
| --- | --- |
| `delayedLoadingIndicatorView.startAnimating(after: 2.0)` | testLoadingIndicator iPhone |

If that's not possible I think it'll be hard to snapshot this, since we're not guaranteed to capture it consistently at the exact same time within the animation loop.

BjornRuud:
> > > Awesome, great work 🎉 The snapshot doesn't actually capture the loading view. But guess it's difficult to snapshot anyways, since it's animating 🤔
> > 
> > 
> > Yes I don't know how to snapshot this with correct content, and we need the snapshot to avoid the snapshot missing error. Any thoughts @osanoj @bstien ?
> 
> If we could snapshot this without it animating, like we have today, that would be the best. In the current demo we do it like shown below. Do we have that option with this view?
> 
> If that's not possible I think it'll be hard to snapshot this, since we're not guaranteed to capture it consistently at the exact same time within the animation loop.

The animation is controlled by SwiftUI and tied to view visibility so not sure how to achieve this. I don't want to contort the code too much either just to support snapshots. :D I'll investigate.

PR #1202, opened 2023-02-07
Fix: Statistics design changes

bstien:
# Why?
The icon for clicks should be changed, the title of the value should be made bold and the icon should be 24x24.

# What?
- 👆 Fix these.
- Remove unneeded public access from `StatisticsItemView`.

# Version Change
Breaking. Making public view internal.

# UI Changes
| Before | After |
| --- | --- |
| Screenshot 2023-02-07 at 11 44 22 | Screenshot 2023-02-07 at 11 55 26 |


FinniversKit/Sources/Components/StatisticsView/StatisticsItemView.swift
```diff
@@ -4,37 +4,38 @@
 
 import UIKit
 
-public class StatisticsItemView: UIView {
+class StatisticsItemView: UIView {
```
BjornRuud, line 7: Is the class used in inheritance or can it be final?

FinniversKit/Sources/Assets/Icons.xcassets/AdManagement/statsClick.imageset/Contents.json
```diff
@@ -0,0 +1,15 @@
+{
+  "images" : [
+    {
+      "filename" : "statsClick.pdf",
```
BjornRuud, line 4: Is this icon available as SVG? I think we should prefer SVG when possible as it is text-based. It should also have the preserve vector data flag set.

BjornRuud (approved):
Looks good! 👍

FinniversKit/Sources/Components/StatisticsView/StatisticsItemView.swift
bstien: Good catch – adding

PR #1206, opened 2023-02-13
Cleanup: Old user ads list views

bstien:
# Why?
Last year we redesigned the screen with the list of user ads to a new endpoint, but there wasn't any time to clean it up back then. This PR removes views related to the old solution.

# What?
- Delete `UserAdsRatingView`.
- Delete `UserAdEmphasizedActionTableViewCell`.

# Version Change
Breaking. Removing protocols and properties in another protocol.

# UI Changes
_No UI._


osanoj (approved):
✂️

PR #1207, opened 2023-02-17
Add flag to selection list for separated items design

Breiby:
# Why?

The current list design looks like it's based on web design. This is a more... appy alternative.

# What?

Add a flag that changes the list design to add some space between the list elements to make them look more like individually tappable elements.

# Version Change

Minor

# UI Changes

| Without flag | With flag |
| --- | --- |
|simulator_screenshot_D20307E0-7079-4592-A9D0-6D607A6276A8|simulator_screenshot_34CBA31A-1405-47A5-B3D2-860FB6677B45|


FinniversKit/Sources/Components/SelectionListView/SelectionListView.swift
```diff
@@ -15,6 +15,7 @@ public class SelectionListView: UIView {
 
     public let presentation: Presentation
     public weak var delegate: SelectionListViewDelegate?
+    let isSeparatedListElementDesign: Bool
```
Breiby, line 18: Open to suggestions for a better name.

Breiby:
Extra space being added on top/bottom of list. Need to fix.

FinniversKit/Sources/Components/SelectionListView/SelectionListView.swift
```diff
@@ -27,22 +28,31 @@ public class SelectionListView: UIView {
...
 
-    public init(presentation: Presentation, withAutoLayout: Bool) {
+    /// Create a selection list view
+    /// - Parameters:
+    ///   - presentation: Determines if list elements should be radio buttons or checkboxes
+    ///   - withAutoLayout: If true, sets `translatesAutoresizingMaskIntoConstraints` to false
+    ///   - isSeparatedListElementDesign: If true, list elements are rendered as separate elements with vertical spacing.
+    public init(presentation: Presentation, withAutoLayout: Bool, isSeparatedListElementDesign: Bool = false) {
         self.presentation = presentation
+        self.isSeparatedListElementDesign = isSeparatedListElementDesign
```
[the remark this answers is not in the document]
BjornRuud: Since type of buttons and how the element appears have various combinations maybe we need a Style struct or something?

FinniversKit/Sources/Components/SelectionListView/SelectionListView.swift
BjornRuud: What about baking this into configurable spacing between elements? If there is no spacing then the original design is used. Maybe not explicit enough?

BjornRuud:
I have no objections to this change but is it part of a design? If so, isn't the non-separate style a component of the design system we're using? I just want to make sure we are consistent in how we use our design components, but I do know designers go off the rails sometimes. :)

FinniversKit/Sources/Components/SelectionListView/SelectionListView.swift
Breiby: I prefer being more explicit with this, but a separate parameter/flag might be overkill

FinniversKit/Sources/Components/SelectionListView/SelectionListView.swift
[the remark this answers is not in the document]
Breiby: I agree, but that would potentially make the change breaking and require changes in torget and wherever else the component is used

FinniversKit/Sources/Components/SelectionListView/SelectionListView.swift
BjornRuud: Separate types of lists is also an option, but the differences are not enough yet for that I think.

FinniversKit/Sources/Components/SelectionListView/SelectionListView.swift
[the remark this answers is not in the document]
BjornRuud: Yup, lots of PRs! 😄 Not necessary for now though.

Breiby:
> I have no objections to this change but is it part of a design? 

Yes, it's part of a design for job profile onboarding.

> If so, isn't the non-separate style a component of the design system we're using? I just want to make sure we are consistent in how we use our design components, but I do know designers go off the rails sometimes. :)

I thought we didn't have a design system for apps yet? As far as I know, the selection list as is just comes from the transaction journey sketches, not a design system or component library.

BjornRuud:
> I thought we didn't have a design system for apps yet? As far as I know, the selection list as is just comes from the transaction journey sketches, not a design system or component library.

Probably right, I haven't seen anything formal, just some common components in Figma. With NMP the verticals will largely do their own thing on top of the NMP design system anyway.

BjornRuud (approved):
👍

PR #1208, opened 2023-02-22
Fix keyboard intersection height with Prefer Cross-Fade Transitions

Breiby:
# Why?

When Reduce Motion and Prefer Cross-Fade Transition are enabled, the keyboard notification returns an empty end frame. This causes us to calculate an incorrect intersection height, which can break views that depend on this value.

# What?

Return 0 if keyboard notification end frame is empty

# Version Change

Patch


MarenZR (approved):
Good fint 🤩

bstien (approved):

PR #1213, opened 2023-02-28
Wishlist - Matvii Support files

sakhnenkoff:
# Why?

In order to mark all necessary support files files available in the main application for the **onboarding Wishlist task**

# What?

- added generic **AsyncImageView** built with SwiftUI 
- added **RibbonViewRepresentable** view to use with SwiftUI
- added some needed for the main application **protocol conformances** to some of the types

# Version Change

Minor

# Should not be merged

### DIFF
--- a/FinniversKit/Sources/Components/Ribbon/Models/RibbonViewModel.swift
+++ b/FinniversKit/Sources/Components/Ribbon/Models/RibbonViewModel.swift
@@ -4,12 +4,29 @@
 
 import UIKit
 
-public struct RibbonViewModel: Hashable {
+public struct RibbonViewModel: Hashable, Codable {
     public let style: RibbonView.Style
     public let title: String
-
+    
     public init(style: RibbonView.Style, title: String) {
         self.style = style
         self.title = title
+    }
+    
+    private enum CodingKeys: String, CodingKey {
+        case style
+        case title
+    }
+    
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        style = try container.decode(RibbonView.Style.self, forKey: .style)
+        title = try container.decode(String.self, forKey: .title)
+    }
+    
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(style, forKey: .style)
+        try container.encode(title, forKey: .title)
     }
 }
--- a/FinniversKit/Sources/Components/Ribbon/Models/RibbonViewModel.swift
+++ b/FinniversKit/Sources/Components/Ribbon/Models/RibbonViewModel.swift
@@ -4,29 +4,12 @@
 
 import UIKit
 
-public struct RibbonViewModel: Hashable, Codable {
+public struct RibbonViewModel: Hashable {
     public let style: RibbonView.Style
     public let title: String
     
     public init(style: RibbonView.Style, title: String) {
         self.style = style
         self.title = title
-    }
-    
-    private enum CodingKeys: String, CodingKey {
-        case style
-        case title
-    }
-    
-    public init(from decoder: Decoder) throws {
-        let container = try decoder.container(keyedBy: CodingKeys.self)
-        style = try container.decode(RibbonView.Style.self, forKey: .style)
-        title = try container.decode(String.self, forKey: .title)
-    }
-    
-    public func encode(to encoder: Encoder) throws {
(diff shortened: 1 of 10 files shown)


FinniversKit/Sources/Components/Ribbon/Models/RibbonViewModel.swift
```diff
@@ -4,12 +4,29 @@
 
 import UIKit
 
-public struct RibbonViewModel: Hashable {
+public struct RibbonViewModel: Hashable, Codable {
```
AndreyMomot, line 7: Currently one of main concern for FinniverseKit SDK is everything related to UI.
All Codable models should be located inside FinnClient SDK as part of API.

If you need that model for UI as well as API, best option is to create 2 models and convert they if needed

FinniversKit/Sources/Components/Ribbon/Ribbon+Style.swift
```diff
@@ -5,7 +5,7 @@
 import Foundation
 
 public extension RibbonView {
-    enum Style: Hashable {
+    enum Style: Hashable, Codable {
```
AndreyMomot, line 8: Same here, no API related stuff inside UI SDK pls )

FinniversKit/Sources/Components/Ribbon/Models/RibbonViewModel.swift
```diff
@@ -4,12 +4,29 @@
...
         self.style = style
         self.title = title
     }
+    
+    private enum CodingKeys: String, CodingKey {
+        case style
+        case title
+    }
+    
+    public init(from decoder: Decoder) throws {
```
AndreyMomot, line 21: Do we actually need custom decoder here? Seems like default provided by using Codable implementation is enough here

FinniversKit/Sources/Components/Ribbon/RibbonViewRepresentable.swift
```diff
@@ -0,0 +1,18 @@
+import SwiftUI
+
+public struct RibbonViewRepresentable: UIViewRepresentable {
```
AndreyMomot, line 3: Looks like you decided to choose UIViewRepresentable because of using `overlay` inside SwiftUI, but why you need it?
Make things simple, just put it inside HStack with all other labels, but set property for your RibbonView 
`fixedSize` or `.layoutPriority(1)` (alternative of `.setContentHuggingPriority(.required, for: .horizontal)` in UIKit as far as I remember)

FinniversKit/Sources/Components/Ribbon/RibbonViewRepresentable.swift
```diff
@@ -0,0 +1,18 @@
...
+        guard let ribbonViewModel else { return nil }
+        self.ribbonViewModel = ribbonViewModel
+    }
+    
+    public func makeUIView(context: Context) -> RibbonView {
+        let view = RibbonView(viewModel: ribbonViewModel)
+        return view
+    }
+    
+    public func updateUIView(_ uiView: RibbonView, context: Context) {
```
AndreyMomot, line 16: While conforming to any Protocols try to avoid braking of SOLID's Interface Segregation Principle principle if it possible (clients should not be forced to implement interfaces they don’t use)

AndreyMomot (commented):
Looks good, but added some comments.
May be I'm bit out of context so some of them are irrelevant.
Additional, would be nice to see visuals of newly added GenericAsyncImageView here inside description of PR

FinniversKit/Sources/Components/Ribbon/Models/RibbonViewModel.swift
sakhnenkoff: indeed, removed

FinniversKit/Sources/Components/Ribbon/Models/RibbonViewModel.swift
sakhnenkoff: noted

FinniversKit/Sources/Components/Ribbon/Ribbon+Style.swift
sakhnenkoff: noted

FinniversKit/Sources/Components/Ribbon/RibbonViewRepresentable.swift
sakhnenkoff: noted

PR #1214, opened 2023-03-02
Recommendation image mixup fix 🤞 

MarenZR:
We suspect this is causing the image to sometimes be the incorrect due some cancellation of imageloadings not functioning as intended.

# Why?

We have had a long-living bug where some images get mixed up in the recommendation-feed, this may be the reason for it. If the imageLoading didn't cancel loading properly, it may have been inserted after prepareForReuse has been called, and the imageView.image would not be nil.

# What?

Remove check for imageView.image == nil in StandardAdRecommendationCell.

# Version Change

Patch


osanoj (approved):
Let's hope it helps 🤞

PR #1215, opened 2023-03-07
Maximise the image quality in the fullscreen view

thorbjxrn:
# Why?

Previously the fullscreen view images only loaded to fill the screen. But the users might want to zoom into the image and take a closer look. So therefore we here bump the image resolution in fullscreen gallery view to get the maximum available.

# What?

Here the image width is set to 3840, aka 4k, instead of the screens width. However the maximum image width in the backend is 1600 so this is then fetched.

# Version Change

minor

# UI Changes

| Before | After |
| --- | --- |
| Screenshot 2023-03-07 at 13 29 58 | Screenshot 2023-03-07 at 13 26 09 |
| Screenshot 2023-03-07 at 13 29 19 | Screenshot 2023-03-07 at 13 25 43 |

### DIFF
--- a/FinniversKit/Sources/Fullscreen/FullscreenGallery/FullscreenGalleryViewController.swift
+++ b/FinniversKit/Sources/Fullscreen/FullscreenGallery/FullscreenGalleryViewController.swift
@@ -268,7 +268,7 @@ extension FullscreenGalleryViewController: FullscreenImageViewControllerDataSour
         }
 
         let url = viewModel.imageUrls[vc.imageIndex]
-        let imageWidth = min(view.bounds.width, view.bounds.height)
+        let imageWidth = CGFloat(3840)
 
         galleryDataSource.fullscreenGalleryViewController(self, imageForUrlString: url, width: imageWidth, completionHandler: { (_, image, _) in
             dataCallback(image)


FinniversKit/Sources/Fullscreen/FullscreenGallery/FullscreenGalleryViewController.swift
```diff
@@ -268,7 +268,7 @@ extension FullscreenGalleryViewController: FullscreenImageViewControllerDataSour
         }
 
         let url = viewModel.imageUrls[vc.imageIndex]
-        let imageWidth = min(view.bounds.width, view.bounds.height)
+        let imageWidth = CGFloat(3840)
```
BjornRuud, line 271: What's stopping us from just using the original image size in fullscreen? Is the size increase so large for some images that it's an issue for us?

thorbjxrn: I went for that first, but since there is no ready method for fetching the original CDN image afaik and that the urls come in two different formats (CDN and AppsGw) I pondered that perhaps this change would come with the same result without the hassle of the contingencies.

PR #1216, opened 2023-03-08
Update UI of Checkmark item details

msaadq:
# Why?

[TJT-1630](https://finn-jira.atlassian.net/browse/TJT-1630)

# What?

- Update UI of Checkmark item details
- Change alignment of the checkmark

# Version Change

Minor


osanoj:
> These changes will affect all users of this view. Are you sure it will not have unintended design changes elsewhere?

Yes, that is a very valid point. Is there a design or image or something so we can see the change proposed?

BjornRuud (changes requested):
- Needs demo.
- Did you investigate if this style change can be global or is local for TJT for this component?
- If local, maybe we need something more flexible, like applying a custom style.
- Item icons (checkmarks) need first line alignment option.

BjornRuud (approved):
👍

PR #1220, opened 2023-03-27
Fix config file

AndreyMomot:
# Why?

CircleCI check failed

# What?

Fix config.yml

# Version Change

Patch

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
 jobs:
   build-and-test:
     macos:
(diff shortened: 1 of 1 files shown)


BjornRuud:
The reason for the change was that the swiftlint in the docker image is really old, and it's not getting updated. The homebrew solution is the one recommended by CircleCI, but apparently some more configuration is needed.

BjornRuud:
I'll experiment a bit on this branch. 🧐

AndreyMomot:
Deal!
And nice that we found this

BjornRuud (approved):
Fixed swiftlint crash. The issue was that swiftlint does not run without a full Xcode installation, the CLI tools are not enough.

PR #1221, opened 2023-04-27
TJT-1921 Address Opt In

msaadq:
# Why?

https://finn-jira.atlassian.net/browse/TJT-1921

# What?

* [Improve the alignment of Checkbox by reducing the extra paddings from the base view](https://github.com/finn-no/FinniversKit/commit/c5fde0dba06f56ebf3a1b7c8cd3c2abe3fcf1282)
* [Modify the BlockUserView which depends on Checkbox and Update Snapshots](https://github.com/finn-no/FinniversKit/commit/355aa20f6c87d99b3a92dd5b4fdff5b878cc2477)

# Version Change

minor


FinniversKit/Sources/Components/SelectionBoxes/Selectionbox.swift
```diff
@@ -149,18 +155,24 @@ extension Selectionbox {
...
-        addSubview(stack)
+        
+        addSubview(containerStack)
+        
+        containerStack.addArrangedSubviews([
+            titleLabel,
+            stack
+        ])
+        
+        if titleLabel.text == nil {
```
BjornRuud, line 166: If you set the label to hidden the stack view will hide it and adjust spacing automatically.

FinniversKit/Sources/Components/SelectionBoxes/Selectionbox.swift
```diff
@@ -78,7 +78,10 @@ public class Selectionbox: UIView {
 
     public var title: String? {
         get { return titleLabel.text }
-        set { titleLabel.text = newValue }
+        set {
+            titleLabel.text = newValue
+            setupSubviews()
```
BjornRuud, line 83: Not necessary to do a new setup each time the title is set.

BjornRuud (changes requested):
Some comments. Since this is FinniversKit the PR should also be reviewed by someone outside TJT.

FinniversKit/Sources/Components/SelectionBoxes/Selectionbox.swift
msaadq: We need to hide the title if the label text is set to nil

BjornRuud: Yes but the label can just toggle isHidden based on presence of the text.

BjornRuud (approved):
👍

PR #1226, opened 2023-05-12
Customizable SwiftUI radio button and checkbox

BjornRuud:
# Why?

Make offer v2 design uses a slightly different radio button and checkbox design compared to Fabric.

# What?

- Made existing SwiftUI implementation of radio button and checkbox customizable.
- Updated radio button inner circle size to Fabric reference size.

# Version Change

Minor. Fully backwards compatible, existing design is the default.


AtishPapa (approved):
LGTM 👍

PR #1227, opened 2023-05-19
My ads tjt overview cell

thorbjxrn:
# Why?

A little Tableview cell with an icon used for showing a TJT status sheet from my ads overview

# What?

This is a new UITableViewCell.

# Version Change

minor

# UI Changes
Screenshot 2023-05-19 at 15 12 37

This is a new cell that will be shown for TJT enabled ads that are public
Screenshot 2023-05-04 at 16 57 08


bstien:
This cell doesn't seem to do much else than the cells we already have for ad management. Can any of those be reused instead?

FinniversKit/Sources/Recycling/ListViews/AdManagement/Cells/UserAdManagementTransactionJourneyStatusCell.swift
```diff
@@ -0,0 +1,47 @@
...
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    // MARK: - Setup
+
+    public func configure(title: String, iconImage: UIImage) {
+        super.configure(with: AdManagementActionCellModel(title: title, iconImage: iconImage))
+
+        showSeparator(false)
```
bstien, line 35: This is not needed – it's already set appropriately when dequeing the cell in AdIn🙂

FinniversKit/Sources/Recycling/ListViews/AdManagement/Cells/UserAdManagementTransactionJourneyStatusCell.swift
```diff
@@ -0,0 +1,47 @@
...
+        contentView.addGestureRecognizer(contentViewTapRecognizer)
```
bstien, line 36: This should be moved to i.e. `setup()`, which you then need to call from `init(style:reuseIdentifier:)`

bstien (approved):
LGTM! Great job 🔥

PR #1233, opened 2023-07-04
Semantic Color Provision : add accentPrimaryBlue and re-order

thorbjxrn:
# Why?

To make the AdInsertion colors compatible with the NMP (see https://github.schibsted.io/finn/ad-insertion-ios/pull/394) I found it reasonable to add a missing color .accentPrimaryBlue to match the existing .accentSecondaryBlue

However, these colors are likely not going to be _blue_ in all NMP marketplaces so they should be re-named but that can wait until the new color semantics are in place.

# What?

A missing part of the logical semantic colorscheme is added, and the order of colours is slightly less random

# Version Change

breaking

# UI Changes

Nothing direct, everything looks like before but you can use .accendPrimaryBlue instead of .btnPrimary when it is logical.

### DIFF
--- a/Demo/SnapshotTests/Info.plist
+++ b/Demo/SnapshotTests/Info.plist
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
 	<key>CFBundleShortVersionString</key>
-	<string>112.1.0</string>
+	<string>113.0.0</string>
 	<key>CFBundleVersion</key>
 	<string>1</string>
 	<key>NSPrincipalClass</key>
--- a/Demo/Sources/Info.plist
+++ b/Demo/Sources/Info.plist
@@ -21,7 +21,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>112.1.0</string>
+	<string>113.0.0</string>
 	<key>CFBundleVersion</key>
 	<string>1</string>
 	<key>ITSAppUsesNonExemptEncryption</key>
--- a/FinniversKit.podspec
+++ b/FinniversKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = 'FinniversKit'
-  s.version      = '112.1.0'
+  s.version      = '113.0.0'
   s.summary      = "FINN's iOS Components"
   s.author       = 'FINN.no'
   s.homepage     = 'https://schibsted.frontify.com/d/oCLrx0cypXJM/design-system'
--- a/FinniversKit/Sources/Components/RemoteImageView/RemoteImageView.swift
+++ b/FinniversKit/Sources/Components/RemoteImageView/RemoteImageView.swift
@@ -5,7 +5,7 @@
 import UIKit
 
 public protocol RemoteImageViewDataSource: AnyObject {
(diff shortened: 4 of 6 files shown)


FinniversKit/Sources/DNA/Color/ColorProvider.swift
```diff
@@ -69,6 +71,10 @@ public struct DefaultColorProvider: ColorProvider {
         .dynamicColor(defaultColor: .aqua50, darkModeColor: .darkIce)
     }
 
+    public var bgInfoHeader: UIColor {
+        .dynamicColor(defaultColor: .accentToothpaste, darkModeColor: .primaryBlue)
```
BjornRuud, line 75: If possible these should be Fabric colors.

FinniversKit/Sources/DNA/Color/UIColor+FinniversKit.swift
```diff
@@ -7,17 +7,19 @@ import UIKit
...
     public class var bgCritical: UIColor { Config.colorProvider.bgCritical }
+    public class var bgSuccess: UIColor { Config.colorProvider.bgSuccess }
     public class var bgInfo: UIColor { Config.colorProvider.bgInfo }
+    public class var bgInfoHeader: UIColor { Config.colorProvider.bgInfoHeader }
     public class var bgPrimary: UIColor { Config.colorProvider.bgPrimary }
-    public class var bgQuaternary: UIColor { Config.colorProvider.bgQuaternary }
     public class var bgSecondary: UIColor { Config.colorProvider.bgSecondary }
-    public class var bgSuccess: UIColor { Config.colorProvider.bgSuccess }
     public class var bgTertiary: UIColor { Config.colorProvider.bgTertiary }
+    public class var bgQuaternary: UIColor { Config.colorProvider.bgQuaternary }
```
BjornRuud, line 22: These were ordered alphabetically so that we don't have to maintain grouping across UIColor, CGColor and Color. 😄

BjornRuud:
Are the new colors from frameworks or completely new? Also, remember to keep colors up-to-date for UIColor, CGColor and Color.

thorbjxrn:
> Are the new colors from frameworks or completely new? Also, remember to keep colors up-to-date for UIColor, CGColor and Color.

The .accentPrimaryBlue / Blue600 is a Fabric main component, same as .btnDefault

The compounded .bgInfoHeader is extrapolated from the adinsertion purchase package cards. It could possibly be in adInsertion extensions, but I found 2 use cases for it and I recon it's a logical component of the new apps color system

ag12 (approved):

osanoj:
This is actually a breaking change, so prepare in main repo before merging

FinniversKit/Sources/DNA/Color/ColorProvider.swift
```diff
@@ -69,6 +71,10 @@ public struct DefaultColorProvider: ColorProvider {
         .dynamicColor(defaultColor: .aqua50, darkModeColor: .darkIce)
     }
 
+    public var bgInfoHeader: UIColor {
+        .dynamicColor(defaultColor: .accentToothpaste, darkModeColor: .accentPrimaryBlue)
```
osanoj, line 75: To make things extra complicated, when you define the semantic color here you should probably rather not use other semantic colors, so here I would rather go for the values those colors have (`.aqua200` and `.blue600` I think)

ag12:
would also be nice *if we could* see the color preview in the DNA-> Color screen. But than you have to make a new screenshot for the tests I guess. In that way, if anyone change something it will break. But tbh I'm not sure if everybody else has followed this pattern.

PR #1234, opened 2023-07-05
Fix: RemoteImageView image loading changes

bstien:
# Why?
We've experienced some stuttering when it comes to image loading after migrating the image loading stack.

# What?
- Bind method on `RemoteImageViewDataSource` to `MainActor`.
- Add property to control when we cancel image loading.

# Version Change
Breaking.

# UI Changes
_No UI._

### DIFF
--- a/FinniversKit/Sources/Components/RemoteImageView/RemoteImageView.swift
+++ b/FinniversKit/Sources/Components/RemoteImageView/RemoteImageView.swift
@@ -25,6 +25,7 @@ public class RemoteImageView: UIImageView {
 
     private var imagePath: String?
     private var imageWidth: CGFloat?
+    private var isLoadingImage = false
 
     // MARK: - Public methods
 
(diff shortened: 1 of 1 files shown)


BjornRuud:
Not sure I understand what caused the stuttering. Was it because `cachedImageWithPath` was somehow async or because `cancelLoadingImageWithPath` was called multiple times?

bstien:
> Not sure I understand what caused the stuttering. Was it because `cachedImageWithPath` was somehow async or because `cancelLoadingImageWithPath` was called multiple times?

The implementation of the dataSource that will ask for cached images will use a few methods that are already bound to `MainActor`, so binding this method here seemed like the smallest amount of code to make that work properly.

BjornRuud:
> > Not sure I understand what caused the stuttering. Was it because `cachedImageWithPath` was somehow async or because `cancelLoadingImageWithPath` was called multiple times?
> 
> The implementation of the dataSource that will ask for cached images will use a few methods that are already bound to `MainActor`, so binding this method here seemed like the smallest amount of code to make that work properly.
> 
> I wrote the same in the PR for the app, but I'm not exactly why the stuttering was happening. However, providing cached images and limiting the number of calls to `remoteImageView(_:cancelLoadingImageWithPath:imageWidth:)` seemed to fix it.

OK! Ideally the caller of an API like this should not have to worry about execution context, but that's a refinement for another day. 😄

BjornRuud (approved):
Though I don't quite understand why this fixes the stuttering, it's a fix. 👍

PR #1236, opened 2023-07-18
marketplace-navigation-bar-colour

adriansergheev:
"marketplaceNavigationBarIcon" property as a "temporary" solution until design team figures out how these colours should be structured.

### DIFF
--- a/FinniversKit/Sources/DNA/Color/ColorProvider.swift
+++ b/FinniversKit/Sources/DNA/Color/ColorProvider.swift
@@ -38,6 +38,7 @@ public protocol ColorProvider {
     var iconSecondary: UIColor { get }
     var iconTertiary: UIColor { get }
     var borderDefault: UIColor { get }
+    var marketplaceNavigationBarIcon: UIColor { get }
 }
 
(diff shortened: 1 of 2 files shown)


FinniversKit/Sources/DNA/Color/ColorProvider.swift
```diff
@@ -174,4 +175,8 @@ public struct DefaultColorProvider: ColorProvider {
     public var borderDefault: UIColor {
         return .dynamicColor(defaultColor: .blueGray300, darkModeColor: .darkSardine)
     }
+
+    public var marketplaceNavigationBarIcon: UIColor {
+        .aqua400
+    }
```
Breiby, line 181: No dark mode?

Breiby (approved):
We love temporary 😄

FinniversKit/Sources/DNA/Color/ColorProvider.swift
adriansergheev: image
image
seems to be the same in both dark/light mode? (top icons, i.e "torget", "jobb" etc)

PR #1237, opened 2023-07-27
Refactor: Use DemoKit

bstien:
# Why?
In order to not duplicate our efforts when it comes to create component showcase apps and/or snapshot tests, I've created a new library that'll help kickstart these two. In good Apple spirit I've named it [DemoKit](https://github.com/finn-no/DemoKit-iOS).

I'll write some proper documentation in that repo's `README`, but I'll start here to explain the differences.

And I'm really sorry for the size of this PR, and since GitHub seems to struggle to show the diff the best way to view this is probably to clone it and open it in Xcode to see the difference.

## Demos
Most of the demo apps we have are based on what we have here in FinniversKit, so I've pulled most of that setup out and tried to make it generic and easily adaptable into what we already have in our repos.

Like FinniversKit has had, DemoKit will also remember both the last demo and tweak you used on a view/demo and reopen this when you re-run the demo in the simulator.

The big difference between the old demo and DemoKit is that anything can be demoed/snapshotted, as long as it implements `Demoable` and you can create a `UIViewController` to visually present it.

### Protocols
DemoKit adds a few new protocols that are to be used/implemented in your target demo app.

**[`Demoable`](https://github.com/finn-no/DemoKit-iOS/blob/406cd504e9b3f955e8f343d985067d3c07d5fba9/Sources/DemoKit/Public/Protocols/Demoable.swift)**
Something that can be demoed and/or snapshot tested. As long as you can create a viewController to contain your `Demoable`, then it can be used. The currently supported types are `UIView`, `UIViewController` _(obviously)_, `View` and `PreviewProvider`. The plan is to open DemoKit to let users provide a `UIViewController` for a given `Demoable`.

Any `Demoable` can decide itself how it'll be presented and how it should be dismissed, as well as some other properties.

**[`TweakableDemo`](https://github.com/finn-no/DemoKit-iOS/blob/406cd504e9b3f955e8f343d985067d3c07d5fba9/Sources/DemoKit/Public/Protocols/TweakableDemo.swift)**
Extends `Demoable` and opens for the `Demoable` to provide tweaks to configure itself. These are used both in the demo app and in snapshot tests.

**[`TweakingOption`](https://github.com/finn-no/DemoKit-iOS/blob/406cd504e9b3f955e8f343d985067d3c07d5fba9/Sources/DemoKit/Public/Protocols/TweakingOption.swift)**
A protocol to be implemented that represents a tweak/state for a `TweakableDemo`. I.e. one tweak would be a view with content, and another tweak would be the same view in an empty state.

**[`DemoGroup`](https://github.com/finn-no/DemoKit-iOS/blob/406cd504e9b3f955e8f343d985067d3c07d5fba9/Sources/DemoKit/Public/Protocols/DemoGroup.swift)**
A group that provides one or more `Demoable`, like we have grouped our demos today within 'Components', 'Fullscreen', etc.
`DemoGroup` can be implemented by both `enum`s and `struct`s.

**[`DemoGroupItem`](https://github.com/finn-no/DemoKit-iOS/blob/406cd504e9b3f955e8f343d985067d3c07d5fba9/Sources/DemoKit/Public/Protocols/DemoGroupItem.swift)**
A lightweight group item existing in a `DemoGroup` that provides a title and an identifier for the list of `Demoable`s within a `DemoGroup`. This had to be created so we don't create/reference views when we need to get a title for the selection list.

## Snapshots
This repo has, until now, only taken snapshots of our views on iPhone in light mode. Several of our demos have tweaks, but they are also disregarded in the snapshots.

In DemoKit there are methods on `XCTestCase` that'll snapshot `DemoGroup` or `Demoable`. All snapshots taken will be done on both iPhone and iPad, as well as both light and dark mode. If the `Demoable` also conforms to `TweakableDemo` then all of its tweaks will be snapshotted, as long as `TweakableDemo.shouldSnapshotTweaks` is `true`.

This will create a lot more snapshots, but it's something we should have in order to make sure changes in PRs doesn't change a view unintentionally.

# What?
- Remove all traces of our old demo.
- Use DemoKit.
- Delete old snapshots and take new ones.

# Version Change
Patch. No API changes to FinniversKit itself.

# UI Changes
_The compression is pretty rough, so colors are not shown properly in the demo._

https://github.com/finn-no/FinniversKit/assets/1901556/0a48ed58-4fcd-472a-a814-e1123fa8dbc6


BjornRuud:
🤩

BjornRuud:
Will the support for PreviewProviders work with Travis? What type of build does Travis use for the snapshots tests? PreviewProviders are stripped from release builds. 🤔

bstien:
> Will the support for PreviewProviders work with Travis? What type of build does Travis use for the snapshots tests? PreviewProviders are stripped from release builds. 🤔

Fudge, that's a good point 🤦‍♂️ I haven't changed which views are snapshotted, but I'll look through the SwiftUI ones to test and see

BjornRuud (approved):
Fantastic work! 🔥 👏 A shared demo solution was needed, we have a lot of demo apps now. I haven't done a superdetailed review of this PR and DemoKit but overall looks good! I think it's best to just start using it and fix anything that crops up as we go, this is for internal use after all.

ag12:
Love it

ag12 (approved):
🧯

PR #1240, opened 2023-08-10
Make GenericErrorViewModel as Equatable

AndreyMomot:
# Why?

Need to compare GenericErrorViewModel outside

# What?

Make GenericErrorViewModel as Equatable

# Version Change

patch

### DIFF
--- a/FinniversKit/Sources/Fullscreen/EmptyView/GenericErrorSwiftUIView.swift
+++ b/FinniversKit/Sources/Fullscreen/EmptyView/GenericErrorSwiftUIView.swift
@@ -1,6 +1,12 @@
 import SwiftUI
 
-public final class GenericErrorViewModel {
+public final class GenericErrorViewModel: Equatable {
+    public static func == (lhs: GenericErrorViewModel, rhs: GenericErrorViewModel) -> Bool {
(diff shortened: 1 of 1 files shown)


FinniversKit/Sources/Fullscreen/EmptyView/GenericErrorSwiftUIView.swift
```diff
@@ -1,6 +1,12 @@
 import SwiftUI
 
-public final class GenericErrorViewModel {
+public final class GenericErrorViewModel: Equatable {
+    public static func == (lhs: GenericErrorViewModel, rhs: GenericErrorViewModel) -> Bool {
```
sakhnenkoff, line 4: I think It might be a bit nicer to move it to a separate extension below the class, but maybe I am just nitpicking here 😅

sakhnenkoff (approved):

FinniversKit/Sources/Fullscreen/EmptyView/GenericErrorSwiftUIView.swift
AndreyMomot: Make sense, updated 👍

PR #1241, opened 2023-08-15
RefreshControl - hide default icon on first refresh

ninarg:
# Why?

We have a FINN-styled `RefreshControl`. We set the default `UIRefreshControl` icon to `.clear` to hide it and insert our own. This is for some reason not working on the first refresh, so you get duplicate icons (default and FINNs).

# What?

Made a fix where we set the alpha to 0 as well. This hides it on first refresh.

# Version Change

Patch.

# UI Changes

| Before | After |
| --- | --- |
| Simulator Screenshot - iPhone 14 Pro - 2023-08-15 at 10 44 53 | Simulator Screenshot - iPhone 14 Pro - 2023-08-15 at 14 50 56 |


ag12 (approved):
alpha

PR #1242, opened 2023-08-24
Add SwiftUI version of the IconButton

Breiby:
# Why?

Enable showing favoritable items in SwiftUI.

# What?

A SwiftUI version of the IconButton.

# Version Change

Minor

# UI Changes

Screenshot 2023-08-24 at 11 43 22

### DIFF
--- a/FinniversKit/Sources/SwiftUI Components/Button/SwiftUIIconButton.swift
+++ b/FinniversKit/Sources/SwiftUI Components/Button/SwiftUIIconButton.swift
@@ -16,9 +16,6 @@ public struct SwiftUIIconButton: View {
 
     public var body: some View {
         Image.init(isToggled ? style.iconToggled : style.icon)
-            .onTapGesture {
-                isToggled.toggle()
-            }
             .accessibility(addTraits: isToggled ? [.isSelected] : [])
             .opacity(isToggled ? 0.8 : 1)
     }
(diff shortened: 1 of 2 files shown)


robinsalehjan (approved):
⭐

FinniversKit/Sources/SwiftUI Components/Button/SwiftUIIconButton.swift
```diff
@@ -0,0 +1,40 @@
...
+    private let style: Style
+
+    public init(style: Style, isToggled: Binding<Bool>) {
+        self.style = style
+        self._isToggled = isToggled
+    }
+
+    public var body: some View {
+        Image.init(isToggled ? style.iconToggled : style.icon)
+            .accessibility(addTraits: isToggled ? [.isSelected] : [])
```
bstien, line 19: A bit late to the party, since this is merged, but I'll leave this either way😄

- [`accessibility(addtraits:)` is deprecated](https://developer.apple.com/documentation/swiftui/view/accessibility(addtraits:)#).
    - We should use [`accessibilityAddTraits(_:)`](https://developer.apple.com/documentation/swiftui/view/accessibilityaddtraits(_:)) instead.
- I think we should add `.isButton` to the list of traits as well, since all we have here is an `Image`.
    - `accessibilityAddTraits(isToggled ? [.isSelected, .isButton] : [.isButton])`

PR #1244, opened 2023-08-28
Update SwiftUI icon button accessibility traits

Breiby:
# Why?

The button is being read as an image.

# What?

Remove the image trait and add the button trait.

# Version Change

Patch

### DIFF
--- a/FinniversKit/Sources/SwiftUI Components/Button/SwiftUIIconButton.swift
+++ b/FinniversKit/Sources/SwiftUI Components/Button/SwiftUIIconButton.swift
@@ -16,7 +16,8 @@ public struct SwiftUIIconButton: View {
 
     public var body: some View {
         Image.init(isToggled ? style.iconToggled : style.icon)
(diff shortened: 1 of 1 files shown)


FinniversKit/Sources/SwiftUI Components/Button/SwiftUIIconButton.swift
```diff
@@ -16,7 +16,8 @@ public struct SwiftUIIconButton: View {
 
     public var body: some View {
         Image.init(isToggled ? style.iconToggled : style.icon)
-            .accessibility(addTraits: isToggled ? [.isSelected] : [])
+            .accessibilityRemoveTraits(.isImage)
+            .accessibility(addTraits: isToggled ? [.isButton, .isSelected] : [.isButton])
```
bstien, line 20: This method is deprecated. We should use [`accessibilityAddTraits(_:)`](https://developer.apple.com/documentation/swiftui/view/accessibilityaddtraits(_:)) instead

bstien (approved):

PR #1249, opened 2023-09-14
Nmp styling image provider

ChristianTherkelsen:
# Why?

For NMP we need to style all components to the brands. This PR handles such styling for images where we can not use a tintColor programatically.

# What?

A new ImageProvider protocol is added. This ensures we can provide brand specific versions of images where multiple colors are needed and therefor not able to programatically tint. In this case its the following:

- Checkbox (animated)
- Radiobutton (animated)
- Checkmark 
- Favourite image (non-added and added state)
- SliderThumb 

The protocol is made with exact same approach as the FontProvider and ColorProvider. It has a default implementation and extensions are added to images can be dotted when used (ex `.brandCheckmark`)

# Version Change

This is a minor release and its adds a new protocol. No breaking changes are added as the new protocol is provided a default implementation that delivers the same images as before.

# UI Changes

No visual UI changes. 

Examples from NMP Tori (with new assets provided) 
Screenshot 2023-09-14 at 11 11 12
Screenshot 2023-09-14 at 11 11 01
Screenshot 2023-09-14 at 11 10 51


FinniversKit/Sources/SwiftUI Components/Button/SwiftUIIconButton.swift
```diff
@@ -2,8 +2,8 @@ import SwiftUI
 
 public struct SwiftUIIconButton: View {
     public struct Style {
-        let icon: ImageAsset
-        let iconToggled: ImageAsset
+        let icon: UIImage
```
ChristianTherkelsen, line 5: I found it necessary to change this style lets to use `UIImage` instead of `ImageAsset`.  Not sure if this is ideal as ImageAssets are commonly used to render images. I could only find the `.favourites` style used with this component where we actually need brand images. 

I could not find any use of this component where the public extension SwiftUIIconButton.Style was not used, so I dont see any breaking changes as a result of the let change 🤔

ChristianTherkelsen:
Investigating CircleCI 👀

ChristianTherkelsen:
I just now realised that we are actually suppose to have to versions of the favourite button image. That is what was causing the test fails. It was working as intended. One for when on top of an image and one that is not. I will revert the snapshots I took an expand the protocol to support both versions of the image 

Screenshot 2023-09-15 at 07 16 37

FinniversKit/Sources/SwiftUI Components/Button/SwiftUIIconButton.swift
bstien: The only usage of this struct seems to be the property `favorite` on line 26 in this file, so we're good :) Also, the properties are internal so this is not a breaking change 👌

bstien (approved):
Great job! Nice to get these images "brandified"😄 🎉

FinniversKit/Sources/SwiftUI Components/Button/SwiftUIIconButton.swift
```diff
@@ -15,7 +15,7 @@ public struct SwiftUIIconButton: View {
     }
 
     public var body: some View {
-        Image.init(isToggled ? style.iconToggled : style.icon)
+        Image.init(uiImage: isToggled ? style.icon : style.iconToggled)
```
Breiby, line 18: Hmm 😅

ChristianTherkelsen: Oh how did I miss that 😲

PR #1250, opened 2023-09-14
iOS 15 minimum version

BjornRuud:
# Why?

The app requires iOS 15, and so should FinniversKit.

# What?

- Set minimum deployment version to iOS 15.
- Fix warnings. Only normal warnings were fixed, the deprecation warnings will be fixed in a separate PR.

# Version Change

Major.


bstien (approved):
LGTM! Nice to get this bumped😄

BjornRuud:
> LGTM! Nice to get this bumped😄

Indeed! And now I need to fix the deprecation warnings, of which there are many! 😅

PR #1251, opened 2023-09-14
Fix deprecations

BjornRuud:
# Why?

Setting iOS 15 as deployment target birthed lots of deprecation warnings.

# What?

Fixed deprecation warnings. I have not fixed (most of) the UIButton deprecation warnings for APIs that were deprecated in iOS 15. Fixing those require moving to UIButtonConfiguration for buttons and that will impact a large part of the app and require a lot of not only visual testing but also behavioural. I'm thinking we do it piecemeal when we have opportunities to change some buttons.

# Version Change

Patch.


osanoj (approved):
Nice to get this fixed!

Issue #1255, opened 2023-09-19
XCode 15 related issues


vishkaischibsted:
To whom it may concern,
Hi, 
There are some pre-generated keys from XCode that has conflict with the current asset keys `ImageAsset`,
They will prevent consumer from compiling this lib 🦀 .

Hopefully these name conflicts can be solved with a small change!

Known keys that has conflicts:

```
1. remove
2. view
3. checkmarkBig
4. ratingFaceDissatisfied
5. magnifyingGlass
```

# Fix proposal:

1. Prepend root type `ImageAsset`.
2. Disable asset symbol generation `in build setting set “Generate Asset Symbols” (ASSETCATALOG_COMPILER_GENERATE_ASSET_SYMBOLS) to NO.`

PR #1258, opened 2023-09-21
Support/xcode 15 symbol generation second try

vishkaischibsted:
# Fix:
Prepend root type to image names that had conflict.

# Re-open reason:
CC @emel.elias@schibsted.com @adriansergheev 

- We tried creating a sample project and adding the `FinnverseKit` what we understood is the asset generation is depending on the project where the packages is added not on the package's project setting itself.

- the ambiguity error from our understanding is because in iOS 17 the compiler generates string enums also, this can be only prevented adding this namespace in code.

Fixes for #1255


ag12:
Will this work?

emelschibsted:
Hopefully it will atleast in the testproject it seems to work 🤞

ag12:
> Hopefully it will atleast in the testproject it seems to work 🤞

👀

PR #1261, opened 2023-09-26
Make `MyAdCollectionViewCell` public

bstien:
# Why?
This cell is needed for the new drafts and markets view for NMP.

# What?
- Make `MyAdCollectionViewCell` public.

# Version Change
Patch.

# UI Changes
_None._


ag12 (approved):
GO

osanoj:
We should really move these away from FinniversKit sometime soon

bstien:
> We should really move these away from FinniversKit sometime soon

Agree. I can take some time to move AdIn specific stuff into that repo after the deadline 👌

PR #1266, opened 2023-09-27
Fix favorite icon being opposite of toggle state

Breiby:
# Why?

There was a regression where the logic was flipped.

# What?

Flip it back.

# Version Change

Patch


BjornRuud (approved):
Flippin' ay.

ChristianTherkelsen (approved):
Good catch! ✅

PR #1272, opened 2023-10-05
Update jobs favourite button icon and color

Breiby:
# Why?

It was using the assets meant for overlaying on an image, but the view the button is used in has a solid color background.

# What?

Change asset and set colors based on toggle.

# Version Change

Patch


robinsalehjan (approved):
LGTM - I guess the snapshot has to be updated as well?

ChristianTherkelsen (commented):
Nice to have the tintColor added here 🙌 I know that Thuy (UX designer) has expressed that the favouriteIcon should be watermelon red for Tori. This would change with this PR to being Petrolium as that is the .btnAction is blue600 for FINN and Petrolium for Tori 🤔

I can see that this is used for Jobs only so far? Maybe lets go with this change and see if we need to revise it later if we need SwiftUIIconButtons with different tint once we get the implement WARP

Breiby:
> Nice to have the tintColor added here 🙌 I know that Thuy (UX designer) has expressed that the favouriteIcon should be watermelon red for Tori. This would change with this PR to being Petrolium as that is the .btnAction is blue600 for FINN and Petrolium for Tori 🤔
> 
> I can see that this is used for Jobs only so far? Maybe lets go with this change and see if we need to revise it later if we need SwiftUIIconButtons with different tint once we get the implement WARP

Haha, yeah, it's kind of the wild west until we get WARP 🤠 I went for btnAction because I saw it was used in `FavoriteButtonView`. Not sure whether the brand colours will do as well in all dark mode contexts? A problem for another day, I guess...

ChristianTherkelsen:
Wild west indeed! 😂 Ah you are right that its used in there. That is the one place I know where it differs because thats the one for the objectpage under the images. For lists and search results its actually watermelon everywhere mainly via using a tintColor or the branded images versions because it exists with a border for the favourite icon. Anywho, Ive applied this other places 👇 Maybe we can do the same here? Thats blue600 for FINN and watermelon for tori 🙂 Would that work for you?

favoriteButton.tintColor = isFavorited ? .nmpBrandColorPrimary : .iconSecondary

Breiby:
> Maybe we can do the same here? Thats blue600 for FINN and watermelon for tori 🙂 Would that work for you?
> 
> favoriteButton.tintColor = isFavorited ? .nmpBrandColorPrimary : .iconSecondary

`blue600` will not provide enough contrast in dark mode in the contexts that this icon is used, so I'd prefer not to as long as `nmpBrandColorPrimary` doesn't have a dark mode color.

ChristianTherkelsen:
Thats a good point @Breiby! I added the . nmpBrandColorPrimary but didnt provide the darkmode color for it. I think it should probably have that. It should probably just be the .darkPrimaryBlue as that is used otherplace as the darkmode support for the blue600. 

Then the .nmpBrandColorPrimary implemetation (for FINN) would be 

```
public var nmpBrandColorPrimary: UIColor {
        .dynamicColor(defaultColor: .blue600, darkModeColor: .darkPrimaryBlue)
    }
```

Breiby:
> Thats a good point @Breiby! I added the . nmpBrandColorPrimary but didnt provide the darkmode color for it. I think it should probably have that. It should probably just be the .darkPrimaryBlue as that is used otherplace as the darkmode support for the blue600.
> 
> Then the .nmpBrandColorPrimary implemetation (for FINN) would be
> 
> ```
> public var nmpBrandColorPrimary: UIColor {
>         .dynamicColor(defaultColor: .blue600, darkModeColor: .darkPrimaryBlue)
>     }
> ```

That would work for me, but you might need to clear it with design. I believe `darkPrimaryBlue` is an "unofficial" color (as in not part of any design systems), so it's possible they don't want that included in these brand color things? I'm a bit out of the loop here, so just want to be careful 😬

Breiby:
> We can also leave it like you made it here and then see what is decided once we start verifying the look of the Tori app. Then I guess someone will have an opinion and it can be decided then

I changed it to `btnPrimary`, which even as a dark mode color for Tori. What do you think?
`.dynamicColor(defaultColor: .blue600, darkModeColor: .darkCallToAction)`
`.dynamicColor(defaultColor: .toriWatermelon, darkModeColor: .nmpDarkmodeBrandBtnPrimary)`

ChristianTherkelsen:
Perfect! 🙌

ChristianTherkelsen (approved):
Ship it :shipit: 🙌

PR #1273, opened 2023-10-12
HostingContentConfiguration fix

BjornRuud:
# Why?

Cells using HostingContentConfiguration could in some cases be blank before refreshed, and could have layout issues. This was caused by the child view controller being added to the parent too early so the parent view controller couldn't be found.

# What?

Manage child view controller after view has moved in/out of the parent view hierarchy.

# Version Change

Patch.


Demo/Sources/Cells/HostingConfigurationCell/HostingConfigurationCell.swift
```diff
@@ -3,9 +3,12 @@ import FinniversKit
 import SwiftUI
 
 final class HostingContentConfigurationCell: UITableViewCell {
+    var onUpdateCellsConfiguredCount: (() -> Void)?
```
AtishPapa, line 6: IMHO introducing this closure just for debug / demo purposes could lead to confusion (at least it did for me 😆). 

Wouldn't it be cleaner to make a specific subclass in the demo target? 🤔

BjornRuud: It is a subclass in the demo target. 😄 But the name should include demo in it, it's a bit confusing.

AtishPapa: Ohh man, I should get new glasses 👴 😆

robinsalehjan (approved):
LGTM 🥇

PR #1277, opened 2023-10-19
TMP-1975 Customizable button padding

BjornRuud:
# Why?

Inner button padding wasn't customizable on SwiftUI button styles so adjusting minimum height was difficult.

# What?

Customizable inner padding for buttons.

# Version Change

Minor. Should be fully compatible and not change anything that uses the button styles.


msaadq (approved):
LGTM 👍

PR #1279, opened 2023-10-30
Swiftlint: Align rules with main repo and fix linting issues

osanoj:
# Why?
There was a lot of warnings (400-ish), many of them from rules we do not really care about.

# What?
- Updated swiftlint rules
- Added a swiftlint autocorrect rule file and ran it (fixing whitespace etc)
- Added "in code" disabling of existing force unwrap, new ones will produce an error!
- Fixed a lot of lint warnings

There are still 50-ish warning, mostly about deprecations - my goal was to quickly get it down to a level we can live with.

# Version Change
Should not make any changes to the end product, so this merge in itself does not require a new version.

# UI Changes
None


Demo/SnapshotTests/BroadcastMessageTests.swift
```diff
@@ -8,7 +8,7 @@ import XCTest
 class BroadcastMessageTests: XCTestCase {
     func testBroadcastModelCanReplaceHTMLLinkWithAttributedStringWithLinkAttrbutes() {
         // Given
-        let linkURL = URL(string: "https://www.finn.no/")!
+        let linkURL = URL(string: "https://www.finn.no/")! // swiftlint:disable:this force_unwrapping
```
BjornRuud, line 11: Instead of special-casing small things like this in the demo target, how about using a nested child config which allows more? Ref. last section in docs: https://realm.github.io/SwiftLint/index.html

BjornRuud (approved):
Nice cleanup! 👍 When moving to Warp we can probably also fix the deprecation warnings about colors.

BjornRuud:
In order to use the same config everywhere, should we consider using a remote config? SwiftLint supports that. 🤔

Demo/SnapshotTests/BroadcastMessageTests.swift
osanoj: I just special cased it for quick fixing now. I think we could quite easily avoid most of the force unwraps in Demo as well, so I hope we will get less of them in the future by having it on 🙂

osanoj:
> In order to use the same config everywhere, should we consider using a remote config? SwiftLint supports that. 🤔

That could be interesting to look into in the future 👍

PR #1280, opened 2023-10-31
Fix testflight build script

osanoj:
# Why?
The fastlane build script for TestFlight distribution didn't work.

# What?
There was a @testable import that caused problems, for now I just made everything that class needed public instead.

Also turned off bitcode since there was an issue with that after fixing the import.

# Version Change
No changes to FinniversKit api


fastlane/Fastfile
```diff
@@ -99,7 +99,7 @@ def build_archive_for_testflight(workspace, scheme)
   build_app(
     scheme: scheme,
     workspace: workspace,
-    include_bitcode: true
+    include_bitcode: false
```
BjornRuud, line 102: Bitcode is deprecated and should always be off now. 👍

FinniversKit/Sources/SwiftUI Components/Button/SwiftUIFloatingButton.swift
```diff
@@ -21,8 +21,10 @@ struct FloatingButtonStyle: ButtonStyle {
     }
 }
 
-struct SwiftUIFloatingButton_Previews: PreviewProvider {
-    static var previews: some View {
+public struct SwiftUIFloatingButton_Previews: PreviewProvider {
```
BjornRuud, line 24: Any other way around this? Public previews feels off. 🤔

BjornRuud (approved):
Remember to fix the public previews at some point. 😄

AtishPapa:
Do you know why @testable didn't work? 🤔 I've been using it in some demo project, but never had any issues with it.

FinniversKit/Sources/SwiftUI Components/Button/SwiftUIFloatingButton.swift
osanoj: There hopefully is, but I just wanted the script to work now. 

I guess either replicate those in the demo project makes the most sense, either that or we will still have demo/preview things available publicly in FinniversKit 🤔

osanoj:
> Do you know why @testable didn't work? 🤔 I've been using it in some demo project, but never had any issues with it.

When you make a release build of Demo it will not be building a @testable version of dependencies, which makes sense, but also makes it so that you don't notice it until making a release 😅

FinniversKit/Sources/SwiftUI Components/Button/SwiftUIFloatingButton.swift
osanoj: I copied them instead of making public, it is better that way I think, do you mind having another look? 😄

BjornRuud (approved):
I'm not a big fan of using preview structs for demos in general, but it works. 👍

PR #1284, opened 2023-11-08
Tori onboarding alert panel for tori users

teddyfahi:
# Why?

Tori onboarding info panel for tori users to inform their RE and mobility favorites are no longer there. 

# What?

A Panel is displayed for all Tori users, all the time, until the temporary view is removed. The panel is displayed for each favorites list.

# Version Change
Breaking. Change is decided by FavoritesConfiguration target bool.

# UI Changes
Skjermbilde 2023-11-06 kl  15 41 52


ag12 (approved):
LGTM

FinniversKit/Sources/Fullscreen/FavoriteAdsList/Subviews/FavoriteAdsListTableHeader.swift
```diff
@@ -76,6 +99,13 @@ class FavoriteAdsListTableHeader: UIView {
         return label
     }()
 
+    private lazy var infoPanel: Panel = {
+            let panel = Panel(style: .tips)
+            panel.backgroundColor = .aqua50
+            panel.translatesAutoresizingMaskIntoConstraints = false
+            return panel
+        }()
+
```
sakhnenkoff, line 108: I think indent might be a bit off here 🤔
try Command + A then Ctrl + I

sakhnenkoff (approved):

FinniversKit/Sources/Fullscreen/FavoriteAdsList/FavoriteAdsListViewModel.swift
```diff
@@ -13,6 +13,8 @@ public struct FavoriteAdsListViewModel {
     public let emptyListViewTitle: String
     public let emptyListViewBody: String
     public let emptyListViewImage: UIImage
+    public var panelTextToriOnboarding: String
+    public var showPanelForToriOnboarding: Bool
```
agusti, line 17: could it be better to just call it
* ` panelTextOnboarding`
* `showPanelForOnboarding`
?
This probably will be used for other brands in the future

teddyfahi: I'll change the var names 👍 Most likely we will have similar onboarding for other brands yes, but this one is very Tori-specific so most likely they will look different or live other places in the app.

agusti (changes requested):
If I understood the PR properly, wouldn't make sense to have `panelTextOnboarding` as optional and then we can even skip the bool property for `showPanelForOnboarding`.
Then if `panelTextOnboarding` is nil, we can just don't show the onboarding panel. Or is there a time where we have text for onboarding but we don't show the panel?

FinniversKit/Sources/Fullscreen/FavoriteAdsList/Subviews/FavoriteAdsListTableHeader.swift
```diff
@@ -53,16 +53,39 @@ class FavoriteAdsListTableHeader: UIView {
...
+        }
+    }
+
+    var showPanel: Bool = false {
+        didSet {
+            infoPanel.isHidden = !showPanel
+        }
+    }
+
+    var panelText: String = "" {
```
agusti, line 71: Maybe optional with the change I mentioned in my request?

FinniversKit/Sources/Fullscreen/FavoriteAdsList/Subviews/FavoriteAdsListTableHeader.swift
```diff
@@ -53,16 +53,39 @@ class FavoriteAdsListTableHeader: UIView {
...
+        }
+    }
+
     // MARK: - Private properties
 
     private lazy var tapRecognizer = UITapGestureRecognizer(target: self, action: #selector(handleViewTap))
 
     private lazy var contentStackView: UIStackView = {
-        let stackView = UIStackView(arrangedSubviews: [titleLabel, subtitleView, searchBar, sortingContainerView])
+        let stackView = UIStackView(arrangedSubviews: [titleLabel, subtitleView, infoPanel, searchBar, sortingContainerView])
```
agusti, line 83: If there is no onboarding, shouldn't it be better to not have the infoPanel at all? it seems we are putting an empty view in case of no onboarding.

teddyfahi: I might be confused here, but since we are getting the `panelTextOnboarding` regardless of which target we are running, we need `showPanelForOnboarding` also to avoid displaying for `finn`? So even though we don't have a case where we want to have text without the panel it self, we do have the text (since it it generated for all the targets simultaneously). And therefore need to know whether or not to show it🤔

agusti (approved):
I checked the PR that implements this in the FINN app and it seems due to the way the localizable strings are done (autogenerated) would be not straightforward to have an optional string for this, so having the boolean makes sense.

PR #1285, opened 2023-11-09
Add extra spacing in SwiftUI CallToAction button

ninarg:
# Why?

Our SwiftUI action button does not have the correct size according to our design system. It's supposed to have more paddings inside, same as our UIKit action button has. 

# What?

Updated to vertical spacing 13, which is the correct spacing from Warp.

I see this button is already used in transaction journeys, where it seems like they have compensated for too little spacing by adding more spacing to the label inside. So we'll need to update those and probably remove the extra spacing.

# Version Change

Minor.

# UI Changes

| Before | After |
| --- | --- |
| Screenshot 2023-11-09 at 08 34 30 | Screenshot 2023-11-09 at 08 32 00 |


bstien (approved):

FinniversKit/Sources/Base Components/ButtonStyle.swift
```diff
@@ -80,11 +80,13 @@ public struct CallToAction: ButtonStyle {
     @Binding var isEnabled: Bool
     private let background: Color
     private let font: Font
+    private let verticalPadding: CGFloat
     private let fullWidth: Bool
 
     public init(size: Button.Size = .normal, background: Color = .btnPrimary, fullWidth: Bool = true, isEnabled: Binding<Bool>? = nil) {
         self.background = background
         self.fullWidth = fullWidth
+        self.verticalPadding = size == .normal ? 13 : .spacingS
```
AtishPapa, line 89: Could you move 13 out to a constant or something? 🙂

ninarg: Not sure what I would call it, 13 is not an "official" kind of spacing, like spacingM, spacingS etc 🤔 
But could move it out to one called `normalButtonSpacing` or something like that?

AtishPapa: Hehe, maybe it'll be an official spacing in Warp then 😄 Maybe just use a local constant then, just to make sure, this magic number is defined somewhere.

ninarg: yeah, I think it will be 🤞 Will add a local one, thanks 👍

ninarg:
@bstien could you give this another review? 😄 had to do some changes after another PR merged to master enabled injection of custom paddings - now we use a new default padding.

bstien (approved):

ag12:
merge this?

ninarg:
> merge this?

We need to update buttons in TJT to merge this. @AtishPapa do you have any time to look into this in the near future? 😄

AtishPapa:
> > merge this?
> 
> We need to update buttons in TJT to merge this. @AtishPapa do you have any time to look into this in the near future? 😄

On it 🙂

pbodsk:
> On it 🙂

@ninarg @AtishPapa 

I now have https://github.com/finn-no/FinniversKit/pull/1306 standing by. Should we coordinate and make a shared release with _this_ PR and _my_ PR?

PR #1288, opened 2023-11-13
Add full width parameter for default style buttons

Breiby:
# Why?

Missing option that we already have for CTA style. 

# What?

Add a parameter for turning off full width on default style buttons.

# Version Change

Minor

# UI Changes

Screenshot 2023-11-13 at 12 32 02


sakhnenkoff (approved):
🚀

PR #1292, opened 2023-11-17
Nmp branding adjustments

ChristianTherkelsen:
# Why?

Enable branding for NMP apps

# What?

A new colour is added here to enable the correct colours for the NMP apps and keep the same for FINN legendary. After a workshop preparing apps for internal verification its clear that we needed more granularity in the colours to meet the design we would like. 

In this specific case its for the tabBar where we want to (keep) use the primaryColor for FINN legendary but the secondaryColor (watermelon) for Tori. This is not a colour pattern we have seen before and is an odd one out. This was the quickest solution without impacting any other places we apply colour schemes

# Version Change

Breaking change as its a new addition to the protocol

# UI Changes

No visible changes in FinniversKit


bstien (approved):
🎨👨‍🎨

PR #1293, opened 2023-11-21
Optional map tiles in AddressMapView

ninarg:
# Why?

We need an option to show `AddressMapView` without map tiles, as we don't have tiles for Finland.

# What?

Add an option to hide map tile selection button in init.

# Version Change

Breaking.

# UI Changes

| Before | After |
| --- | --- |
| Simulator Screenshot - iPhone 14 Pro - 2023-11-21 at 10 56 50 | Simulator Screenshot - iPhone 14 Pro - 2023-11-21 at 10 57 23 |


ag12 (approved):
⚡

robinsalehjan (approved):

PR #1299, opened 2023-12-11
Use updated DemoKit.

BjornRuud:
# Why?

DemoKit has been updated to use the official swift-snapshot package instead of the fork.

# What?

Updated DemoKit.
One snapshot test had to be rerecorded although the diff didn't show anything obviously different. 🤷 

# Version Change

Minor.


BjornRuud:
We still need to use the swift-snapshot fork.

PR #1302, opened 2023-12-15
Smallfix/set 401 color dark mode package label

KristerMoen:
# Why?
Wrong background color on package label background and text in dark mode.

# What?
**This PR is related to**: [smallfix/SET-401-color-dark-mode-package-label](https://github.schibsted.io/finn/ios-app/tree/smallfix/SET-401-color-dark-mode-package-label)

Was initially working on [Investigate package label color bug iOS ](https://nmp-jira.atlassian.net/browse/SET-401) and then i stumbled upon the same issue in Tori.
Had a small talk with @thorbjorn-throndsen-bonvik and i got info about Warp colors and that it's not really finished yet. So we took the decision on choosing colors similar to the `BadgeView` with the `warning`-style.
 
- Add new `textAlert` color
- Remove private extension for warningTextColor and replace with ColorProvider type instead.

# Changelog
Add color for textAlert, dark mode for bgAlert removed private color extension


BjornRuud (approved):
Nice, always good to get rid of custom colors. 👍 Has this been cleared with UX? Although we should in theory get all colors from the design system we do have a few that are on purpose.

KristerMoen:
@BjornRuud Yes, UX is onboard.
The design system didn't give me any suggestions about this issue.

How do i rerun the `ci/circleci: build-and-test` and get the snapshot tests green✅ ?

BjornRuud:
> How do i rerun the `ci/circleci: build-and-test` and get the snapshot tests green✅ ?

You can tap the details link and do it there. Do the tests run locally with the same target device and iOS version that the CI build uses?

KristerMoen:
@BjornRuud Is it ok to push all new snapshots or should i just upload for the snapshot failed ?
I think it was 120+ snapshot tests that failed.

BjornRuud:
> @BjornRuud Is it ok to push all new snapshots or should i just upload for the snapshot failed ? I think it was 120+ snapshot tests that failed.

If they fail because of intended changes then the snapshots need to be updated. But double-check that not something else is triggering this, like wrong device/iOS used for snapshot.

KristerMoen:
@BjornRuud Yes, you are correct it failed because of wrong device type. 
Reverted back to [master](https://github.com/finn-no/FinniversKit/tree/master) snapshot and now hopefully the CI tests will pass.

PR #1303, opened 2023-12-18
Refactore disabled logic

AndreyMomot:
# Why?

Button is still clickable even if location not available

# What?

- simplify logic
- disable button if needed

# Version Change

Major

# UI Changes

| Before | After |
| --- | --- |
| щдв| new |

### DIFF
--- a/FinniversKit/Sources/Components/MapAddressButton/MapAddressButton.swift
+++ b/FinniversKit/Sources/Components/MapAddressButton/MapAddressButton.swift
@@ -59,8 +59,10 @@ public class MapAddressButton: UIView {
         button.setImage(icon, for: .normal)
     }
 
-    public func setTitleColor(_ titleColor: UIColor) {
+    public func setButton(isEnabled: Bool) {
+        let titleColor: UIColor = isEnabled ? .textAction : .textDisabled
         buttonStyle = buttonStyle.overrideStyle(textColor: titleColor)
+        button.isUserInteractionEnabled = isEnabled
 
         // Updating the style sets certain properties on the button back to their default values, including `titleEdgeInsets`.
         // Getting the current value here so we can set it back after changing the style.


FinniversKit/Sources/Components/MapAddressButton/MapAddressButton.swift
```diff
@@ -59,8 +59,10 @@ public class MapAddressButton: UIView {
         button.setImage(icon, for: .normal)
     }
 
-    public func setTitleColor(_ titleColor: UIColor) {
+    public func setButton(isEnabled: Bool) {
+        let titleColor: UIColor = isEnabled ? .textAction : .textDisabled
```
bstien, line 63: I don't know why the properties on `Button.Style` is not public yet, but maybe we could make them public and rather here use `buttonStyle.titleColor` instead of `.textAction` so we don't unintentionally override the style we set above.

AndreyMomot: Make sense
`Button.Style `is inside `public extension` so no need in more `public` 
Updated, does it looks better now?
p.s. Tested both with active & inactive = looks the same

bstien: The struct `Button.Style` is public, but the properties are not – they are only internal.

Yes, now the colors will be in sync if we update the button style 🙂 👌

AndreyMomot: So you basically saying it is smth new? :)
Screenshot 2023-12-19 at 09 40 28

bstien: No, this has been like this for a long time. I'm talking about the properties, not the structs themselves.

Screenshot 2023-12-19 at 09 44 01

bstien: But it's not that important now – you're able to access them since what you're adding is within FinniversKit.

AndreyMomot: 👌

PR #1305, opened 2023-12-28
Favorites: Add support for messages

bstien:
# Why?
The favorite API has added a new capability for sending messages to clients when the contents of a favorite folder is requested. Currently we only have support for presenting a single message, aka. the Tori onboarding, within a favorite folder, but now we might need to present multiple ones.

# What?
- Refactor layout of `tableHeaderView`.
  - Layout was not consistent when changing the content of the header after initial presentation.
- Add method for configuring info messages.
- Remove old properties for Tori onboarding.
- Update demo and snapshots reference images.

# Version Change
Breaking. Removing a few public properties.

# UI Changes
| Before | After |
| --- | --- |
| Screenshot 2023-12-28 at 14 05 34 | Screenshot 2023-12-28 at 13 36 05 |


thorbjxrn (approved):
Looks good

PR #1306, opened 2024-01-08
Feature/swiftui infobox view

pbodsk:
# Why?

We...sorry _I_ needed a Infobox View matching the one in Warp so I made that. I also made a SwiftUI version of our existing `InfoboxView` if anyone needs that

# What?

## `NMPInfoboxView`
First item on the agenda is a new SwiftUI view called `NMPInfoboxView`. This matches what we need for the Warp component.

(And yes I'm willing to negotiate on the name 😄 I wanted something to reflect that this is not the "old" InfoboxView and NMP was the best I could come up with (come on...what do you want from me...I'm not creative 😄))

Here's how it looks first of all

Screenshot 2024-01-08 at 11 54 42

### `NMPInfoboxView.ViewModel`
To populate the Infobox you give it a `NMPInfoboxView.ViewModel` that holds the various attributes needed for the Infobox. To handle the "optional interaction components with a title and an action callback" scenario I introduced a `Interaction` `struct` that can be used for optional links, primary and secondary button interactions

### `NMPInfoboxView.InformationType`
The most interesting or exotic component in the ViewModel is probably the `NMPInfoboxView.InformationType` that allows you to define which Infobox you would like to use. 

You have options for the values defined in Warp, and if you use one of the default values colors and icons will be set up for you automagically 🪄 

You also have the option to use a custom Infobox type if you (or more likely a designer 😄 ) desires so.

## `InfoboxSwiftUIView`
Since I was here anyway and to improve my SwiftUI skills (and since I promised on Slack :wink:) I thought I'd take a stab at making a SwiftUI version of our existing `InfoboxView`

Here's how that ended up

Screenshot 2024-01-08 at 12 01 19

Nothing too crazy here. I was able to reuse the existing `ViewModel` and `Style` but I had to add a couple of (optional )callbacks for the primary and secondary button actions

```
public var onPrimaryButtonTapped: (() -> Void)?
public var onSecondaryButtonTapped: (() -> Void)?
```

### ViewModifier for button styles
The most interesting problem here was probably how to convert values like this `var primaryButtonStyle: Button.Style` into something tangible for SwiftUI

I (with my good partner in crime ChatGPT) ended up solving it like so. I'd like to hear your thoughts on that. To me it is quite "SwiftUI'y" but at the same time I don't like that I'm polluting the global namespace/"viewmodifier land" with this...but it looks good  😎 👌 

```
struct PrimaryButtonStyleModifier: ViewModifier {
    let style: InfoboxView.Style
    func body(content: Content) -> some View {
        switch style {
        case .small:
            content.buttonStyle(DefaultStyle())
        case .normal:
            content.buttonStyle(CallToAction())
        case .warning:
            content.buttonStyle(FlatStyle())
        }
    }
}

extension View {
    func primaryButtonStyle(for style: InfoboxView.Style) -> some View {
        self.modifier(PrimaryButtonStyleModifier(style: style))
    }
}

struct SecondaryButtonStyleModifier: ViewModifier {
    let style: InfoboxView.Style
    func body(content: Content) -> some View {
        switch style {
        case .small:
            content.buttonStyle(FlatStyle())
        case .normal:
            content.buttonStyle(InlineFlatStyle())
        default:
            content.buttonStyle(DefaultStyle())
        }
    }
}

extension View {
    func secondaryButtonStyle(for style: InfoboxView.Style) -> some View {
        self.modifier(SecondaryButtonStyleModifier(style: style))
    }
}
```

## Everything else
I've made demo screens for the new components so you have those in the mix as well 😄 

# Version Change

This is probably a minor version. A new component but no breaking changes.

# UI Changes

See above 👆 

# Parting words
Since this is my first PR to FinniversKit I hope you'll tear it to shreds, let me know what I've missed and what should be changed...don't hold back, rather you than our customers 😄 

Also, @bstien and @ninarg I've added you two as reviewers, mostly because I don't know who else to add. If you have suggestions on who should review, just let me know 😄


FinniversKit/Sources/Components/Infobox/InfoboxSwiftUIView.swift
```diff
@@ -0,0 +1,113 @@
...
+
+public struct InfoboxSwiftUIView: View {
+    public let style: InfoboxView.Style
+    public var viewModel: InfoboxViewModel?
+    public var onPrimaryButtonTapped: (() -> Void)?
+    public var onSecondaryButtonTapped: (() -> Void)?
+
+    public init(
+        style: InfoboxView.Style,
+        viewModel: InfoboxViewModel? = nil,
```
BjornRuud, line 15: Is not having a view model really an option?

FinniversKit/Sources/Components/Infobox/InfoboxSwiftUIView.swift
```diff
@@ -0,0 +1,113 @@
...
+        onPrimaryButtonTapped: (() -> Void)? = nil,
```
BjornRuud, line 16: Tip: If you want to follow the view modifier pattern you can express these as modifier functions instead. Something like:

```
func onSecondaryButton(_ handler: () -> Void) -> Self {
    var infobox = self
    infobox.onSecondaryButtonTapped = handler
    return infobox
}
```

Not really necessary here, but a nice trick if you end up with a huge init parameter list.

FinniversKit/Sources/Components/NMPInfobox/NMPInfoboxView+InformationType.swift
```diff
@@ -0,0 +1,80 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
```
BjornRuud, line 2: Top template comments should be deleted.

FinniversKit/Sources/Components/NMPInfobox/NMPInfoboxView.swift
```diff
@@ -0,0 +1,201 @@
...
+            .bodyStrong
+        }
+
+        var detailStyle: Label.Style {
+            .body
+        }
+    }
+}
+
+#warning("Og så skal der laves en SwiftUI version som rent faktisk bruger den der Infobox ting")
```
BjornRuud, line 53: Indeed! 😄 For these kinds of comments which are about the intent of the code more than the code itself it might be better to use `// TODO:` and let SwiftLint flag it as a warning.

FinniversKit/Sources/Components/NMPInfobox/NMPInfoboxView.swift
```diff
@@ -0,0 +1,201 @@
...
+public struct NMPInfoboxView: View {
+    public let viewModel: ViewModel
+
+    public init(viewModel: ViewModel) {
+        self.viewModel = viewModel
+    }
+
+    public var body: some View {
+        HStack(alignment: .top, spacing: 8) {
```
BjornRuud, line 62: For spacing, prefer the Fabric (and I assume Warp) spacing constants defined in FinniversKit like `.spacingS`.

FinniversKit/Sources/Components/NMPInfobox/NMPInfoboxView.swift
```diff
@@ -0,0 +1,201 @@
...
+                    .buttonStyle(InlineFlatStyle())
+                }
+
+                HStack(spacing: 8) {
+                    if let primaryButtonInteraction = viewModel.primaryButtonInteraction {
+                        SwiftUI.Button(
+                            action: { primaryButtonInteraction.onTapped() },
+                            label: {
+                                Text(primaryButtonInteraction.title)
+
```
BjornRuud, line 88: Extra space.

FinniversKit/Sources/Components/NMPInfobox/NMPInfoboxView.swift
```diff
@@ -0,0 +1,201 @@
...
+            RoundedRectangle(cornerRadius: 4)
+                .stroke(viewModel.informationType.borderColor, lineWidth: 1.0)
+        )
+        .overlay( // sidebar
+            Rectangle()
+                .fill(viewModel.informationType.sideboxColor)
+                .frame(width: 4),
+            alignment: .leading
+        )
+        .frame(maxWidth: /*@START_MENU_TOKEN@*/.infinity/*@END_MENU_TOKEN@*/)
```
BjornRuud, line 118: Placeholder markers shouldn't be there.

FinniversKit/Sources/Components/NMPInfobox/NMPInfoboxView.swift
```diff
@@ -0,0 +1,201 @@
...
+                    }
+
+                    Spacer()
+                }
+            }
+
+        }
+        .padding(.spacingM)
+        .overlay( // border
+            RoundedRectangle(cornerRadius: 4)
```
BjornRuud, line 109: Not sure if the design system has predefined radius constants? 🤔

BjornRuud:
> If you have suggestions on who should review, just let me know 😄

Nice work! What I usually do is add `ios-reviewers` as reviewer to get two random reviewers and then I add any specific people that should review or otherwise be involved.

FinniversKit/Sources/Components/NMPInfobox/NMPInfoboxView.swift
pbodsk: Friday afternoon Peter telling Monday morning Peter what to do 😄 

Gone now ✅

FinniversKit/Sources/Components/Infobox/InfoboxSwiftUIView.swift
pbodsk: Yeah I was surprised too but it seems so, here's the original in `InfoboxView`

```
public var model: InfoboxViewModel? {
        didSet {

        }
}
```

🤔 but I don't know if that is a side-effect or workaround for some compiler related maybe?

FinniversKit/Sources/Components/Infobox/InfoboxSwiftUIView.swift
pbodsk: OK _that_ was a nice trick! Thanks 🙌

FinniversKit/Sources/Components/NMPInfobox/NMPInfoboxView.swift
pbodsk: Ah! Nice catch ✅

FinniversKit/Sources/Components/NMPInfobox/NMPInfoboxView.swift
pbodsk: 🤔 interesting. Must've been because I just used the default value, it looked OK in Xcode

FinniversKit/Sources/Components/NMPInfobox/NMPInfoboxView.swift
pbodsk: It doesn't look so. I can use `.spacingXS`, that matches 4

FinniversKit/Sources/DNA/Color/ColorProvider.swift
```diff
@@ -94,6 +106,50 @@ public struct DefaultColorProvider: ColorProvider {
         .red100
     }
 
+    public var bgInformationAlert: UIColor {
```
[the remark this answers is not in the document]
pbodsk: You're right

FinniversKit/Sources/Components/NMPInfobox/NMPInfoboxView+InformationType.swift
pbodsk: Sorry I don't follow...In all files?

FinniversKit/Sources/Components/Infobox/InfoboxSwiftUIView.swift
BjornRuud: My suggestion is to require the view model and see what problems crop up. 😄

FinniversKit/Sources/Components/NMPInfobox/NMPInfoboxView+InformationType.swift
BjornRuud: Yes, we have no need for these autogenerated comments. Or wait, this is public Github, might be different here? @osanoj

FinniversKit/Sources/Components/NMPInfobox/NMPInfoboxView.swift
BjornRuud: Yes that's correct, the default is always a placeholder with hidden markup so you have to input something. As you can see it will work fine but the code gets messy. 😄

FinniversKit/Sources/Components/NMPInfobox/NMPInfoboxView.swift
BjornRuud: Not sure that's a good idea. If for some reason the spacing gets changed the corner radius will change also, which is not correct. We need a new set of constants for radius if we need it. For now just leave it to what the design says it should be. You can ask the design team if the design system has or should have corner radius definitions.

FinniversKit/Sources/Components/Infobox/InfoboxSwiftUIView.swift
```diff
@@ -0,0 +1,122 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import SwiftUI
+
+public struct InfoboxSwiftUIView: View {
+    public let style: InfoboxView.Style
+    public var viewModel: InfoboxViewModel?
+    var onPrimaryButtonTapped: (() -> Void)?
```
BjornRuud, line 10: Since this view uses a view model I like to put view behaviour and "business" logic inside the view model. You can then easily reference the handler like `Button(action: viewModel.onPrimaryButtonTapped)` since the signature is the same. The view model can also be easily tested. The style can go in the view model too.

BjornRuud: This is just a suggestion though, the "what goes where" debate is endless. 😄

FinniversKit/Sources/Components/NMPInfobox/NMPInfoboxView.swift
```diff
@@ -0,0 +1,216 @@
...
+                    SwiftUI.Button(
+                        action: { linkInteraction.onTapped() },
+                        label: {
+                            Text(linkInteraction.title).underline()
+                        }
+                    )
+                    .buttonStyle(InlineFlatStyle())
+                }
+
+                HStack(spacing: 8) {
```
BjornRuud, line 98: Can use spacing constant.

FinniversKit/Sources/Components/Infobox/InfoboxSwiftUIView.swift
pbodsk: So far none...and it makes the code prettier so let's change it to that ✅

FinniversKit/Sources/Components/NMPInfobox/NMPInfoboxView+InformationType.swift
pbodsk: Yeah all the other files has the comments so that's why I kept them. I mean, I'll happily delete them, as long as we are consistent 😄

FinniversKit/Sources/Components/NMPInfobox/NMPInfoboxView.swift
pbodsk: dammit...I thought I'd found them all 😅

FinniversKit/Sources/Components/NMPInfobox/NMPInfoboxView.swift
pbodsk: Fair point, I can live with that 🤝

FinniversKit/Sources/Components/Infobox/InfoboxSwiftUIView.swift
pbodsk: I don't disagree 😄  

However...

I think the reason I haven't done so already is that I've reused the `InfoboxViewModel` that already exists and is used with the existing `UIKIt` `InfoboxView`. Here the `InfoboxView` has:

- a `InfoboxViewModel`
- a `style` (`InfoboxView.Style`)
- a `delegate` (`InfoboxViewDelegate`)

And we can discuss if that is a pretty solution or not...it is what it is 🤷 

After reading your suggestion I tried adding `var onPrimaryButtonTapped: (() -> Void)? { get set }` to the `InfoboxViewModel` but that is a protocol so that just opened a new can of worms and would require changes to the existing implementations of `InfoboxViewModel` so....I slowly backed away 😄

BjornRuud (approved):
Nice work, and great PR description!

FinniversKit/Sources/Components/NMPInfobox/NMPInfoboxView+InformationType.swift
osanoj: Yes, we usually keep them on public GitHub repos. 

And remove them on internal repos, just to make sure to keep everyone one their toes 😜

FinniversKit/Sources/DNA/Color/ColorProvider.swift
```diff
@@ -54,168 +66,110 @@ public struct DefaultColorProvider: ColorProvider {
 
     public static let shared = DefaultColorProvider()
 
-    public var bgPrimary: UIColor {
-        .dynamicColor(defaultColor: .white, darkModeColor: .darkMilk)
-    }
+    public var bgPrimary: UIColor { .dynamicColor(defaultColor: .white, darkModeColor: .darkMilk) }
```
ninarg, line 69: I think this format makes the colors less readable, they really stand out when placed below, here you have to read to the right 🤔 and all the colors are not aligned in the same position
Not sure if it's just me though?

pbodsk: No I see your point.

What about single value vars like this one for instance:

`public var accentPrimaryBlue: UIColor { .blue600 }`

would you prefer to have that expressed like this too

```
public var accentPrimaryBlue: UIColor { 
    .blue600 
}
```

For consistency's sake?

I've grown to like the "new parameter = new line" pattern so if it was me I would write it like so:

```
public var textCritical: UIColor {
    .dynamicColor(
        defaultColor: .red600,
        darkModeColor: .red400
    )
}
```

But that leads to a lot of lines, which may not be what we're after in a "set and forget" struct like this one.

TL;DR; I can live with both 😄

FinniversKit/Sources/DNA/Color/ColorProvider.swift
```diff
@@ -12,9 +12,21 @@ public protocol ColorProvider {
...
+    var bgInformationInfo: UIColor { get }
+    var bgInformationSuccess: UIColor { get }
     var bgInfoHeader: UIColor { get }
     var bgSuccess: UIColor { get }
     var bgCritical: UIColor { get }
+    var borderAlert: UIColor { get }
+    var borderCritical: UIColor { get }
+    var borderInfo: UIColor { get }
+    var borderSuccess: UIColor { get }
+    var sidebarAlert: UIColor { get }
```
ninarg, line 26: What is sidebar referring to? There are quite a few new semantic colors here, we seldom add new ones. But I guess the colors you are using in this PR are not really used anywhere else yet - so they're not available in other semantic colors?

FinniversKit/Sources/DNA/Color/ColorProvider.swift
ninarg: I would prefer 
```
public var accentPrimaryBlue: UIColor { 
    .blue600 
}
```
so your eyes can follow the same "path", no matter if it's dynamic or not. Tbh I prefer to keep everything like it was 😄 it will also make life easier for the ones of us who currently work on colors as well (I am switching out colors in Tori, and Mohsen is making a new design system, so both of us will get a bunch of conflicts if we change the format of all colors 🙈).

BjornRuud: I agree for the dynamic colors. I don't have any strong opinions here so long as it's consistent within the file. The style we follow in general is function body on separate line and allow very short and simple bodies on same line, so depends on how we define short and simple. 😄

ninarg:
This is a breaking change, you add new required colors to the public API, and you'll need to inject these colors for Tori. If I'm not mistaken, all these info boxes should have other colors in Tori. So you'll need to have an approved PR in the main app before merging this one.

FinniversKit/Sources/DNA/Color/ColorProvider.swift
pbodsk: Yeah it slowly span out of control for me as I worked my way through te design. I've added this picture that hopefully says more than words 😄 

Screenshot 2024-01-09 at 09 55 23

So you have 3 required colors for the component...and you have 4 variants of the component... :sweat_smile:

I'm happy to change the name but I'm struggling to come up with something intelligent. 

I asked ChatGPT who came up with these suggestions

```
InfoBar: Since the dark blue area includes an informational icon (the "i"), "InfoBar" might be a descriptive and succinct term.
Highlight Strip: This term could work because the area stands out and serves to highlight or draw attention to the importance of the information.
Accent Panel: The term "panel" is often used in design to refer to distinct sections of a user interface, and "accent" suggests that it has a visual emphasis different from the main content area.
Status Bar: If the dark blue area is intended to indicate some kind of status or importance, "Status Bar" could be an appropriate term.
Lead-in Stripe: This term could be used if the purpose of the area is to lead the eye to the important message or if it acts as an introductory element.
Tagline: If the element is meant to tag the content with some sort of categorization or importance, "Tagline" could be suitable.
Alert Rail: If the element serves to alert the user to important information, combining "alert" with "rail" (a design term often used for side elements) might fit.
Notification Edge: This term implies that the area serves as a notification or alert to the user about something important.
Preface Sidebar: A preface generally introduces something, and this term could be used if the dark blue area serves to introduce or preface the content to follow.
Priority Bar: If the content within this dark blue area is of higher priority or importance, this name would convey that significance.
```

Accent could be a candidate 🤔

FinniversKit/Sources/DNA/Color/ColorProvider.swift
BjornRuud: Let's do separate lines then, so it's consistent with the style we have in general. Separate lines when there are multiple parameters is also fine by me.

FinniversKit/Sources/DNA/Color/ColorProvider.swift
BjornRuud: How are these colors referenced in the design system? Maybe we can get names from there?

pbodsk:
> So you'll need to have an approved PR in the main app before merging this one.

Perfect, I got the the details on Slack and will get going once I'm through the other PR feedback

FinniversKit/Sources/DNA/Color/ColorProvider.swift
ninarg: Yeah, I think we can follow the the design system. Semantic colors should be reusable in other contexts, the sidebar naming here is probably only relevant for this exact view. I checked the case with yellow-600, that one seems to be called `color-background-warning` in the design system, so we could reuse something like `bgWarning`.

FinniversKit/Sources/DNA/Color/ColorProvider.swift
osanoj: I would agree with Nina here, keep it as it is now, as small diff as possible, since we have a quite important color change being worked on for Tori. (Or wait until after that is done)

But in general we probably want to keep the changes here as few as possible because it is a temporary solution until we can go over to warp. So even making things more readable and whatever can be seen as a bit unnecessary at the time 😅

FinniversKit/Sources/DNA/Color/ColorProvider.swift
BjornRuud: Also, any colors unique to a view do not belong in the color provider. Just use a private extension along with the view, but it should be very rare that a semantic color you can use doesn't exist.

FinniversKit/Sources/DNA/Color/ColorProvider.swift
pbodsk: Right...Gotcha, I will fall back to separate lines.

FinniversKit/Sources/DNA/Color/ColorProvider.swift
pbodsk: After getting introduced to the "NMP Design Tokens taxonomy v2.0" document I made some changes so that we now have colors named:

- border<informationType>Subtle: for the border around the edges of the box
- border<InformationType>: for the left sidebar.

where informationType can be:

- critical
- information
- success
- warning

It seems these values must support light and dark mode and must be different in default and Tori configurations. That speaks to the colors being resent in the Color provider right?

pbodsk: Sheeeesh! I forgot about the background colors. They got an update as well and now there is support for light and dark mode

Screenshot 2024-01-09 at 13 25 40

FinniversKit/Sources/DNA/Color/ColorProvider.swift
```diff
@@ -129,13 +189,13 @@ public struct DefaultColorProvider: ColorProvider {
...
     }
-    
+
     public var textAlert: UIColor {
         .lightNuttyBrown
     }
 
     public var textCritical: UIColor {
-        .dynamicColor(defaultColor: .red600, darkModeColor: .red400)
+        .dynamicColor( defaultColor: .red600, darkModeColor: .red400 )
```
ninarg, line 198: can remove spaces here for consisteny 😄

BjornRuud, line 198: No space before/after parenthesis.

FinniversKit/Sources/DNA/Color/ColorProvider.swift
ninarg: ooo, nice 😍

ninarg (approved):
Reviewed color changes, looks good! Great work 🔥

FinniversKit/Sources/DNA/Color/ColorProvider.swift
BjornRuud: > It seems these values must support light and dark mode and must be different in default and Tori configurations. That speaks to the colors being resent in the Color provider right?

👍

FinniversKit/Sources/DNA/Color/ColorProvider.swift
pbodsk: I don't know what you're talking about, there's no space there 😉 

(this would be way more convincing if you couldn't see all my changes 🤷 )

pbodsk: Haha...both @BjornRuud and you found this 🔍  

fixed ✅

PR #1307, opened 2024-01-10
upgrade snapshot tests dependency

adriansergheev:
# Why?
https://github.com/finn-no/DemoKit-iOS/pull/4
https://github.com/finn-no/DemoKit-iOS/pull/5

we don't need to use a fork anymore for swift-snapshot-testing since [this pr](https://github.com/pointfreeco/swift-snapshot-testing/pull/628) has been merged

# What?
- even though the fork was dropped, this is not a drop-in replacement. unfortunately, Circle CI which is the current environment seems to be a [special case ](https://github.com/pointfreeco/swift-snapshot-testing/issues/419)
- there's still 4 failing screenshot tests when run on CI env (CircleCI) but not on `Mx` macs, and for those the precision has been dropped. 

# Version Change
- minor change, should affect only internal snapshot tests


adriansergheev:
https://github.com/pointfreeco/swift-snapshot-testing/issues/419

bstien (approved):

PR #1308, opened 2024-01-10
Fix: Recommendation images loading issues

bstien:
# Why?
For some time we've had the issue that images for ad recommendations were not presented, or would present in the wrong cell.

The issue seems to be happen because of a couple of reasons:
- We're starting loading of images in `collectionView(_:willDisplay:forItemAt:)`, instead of `collectionView(_:cellForItemAt:)`.
- Ongoing image requests would be cancelled when a new cell is configured, i.e. when scrolling or the collection view figuring out the height needed for a cell. `UICollectionView` cell dequeuing happens in Apple's black box, and the image requests are async, which could cause cancellation of image requests we actually expect to finish.

# What?
- `RemoteImageView`: Verify that the image request callback is called for the current image.
- `AdRecommendationsGridView`: Start loading images as soon as cell is dequeued.
- Remove calls to `cancelLoading()` from cells.
  - Images requested will be cached, and since the cells are dequeued and change content frequently we should not cancel ongoing requests.

# Version Change
Patch. No API changes.

# UI Changes
| Before | After |
| --- | --- |
| Simulator Screen Recording - iPhone 15 Pro - 2024-01-10 at 11 57 14 | Simulator Screen Recording - iPhone 15 Pro - 2024-01-10 at 10 27 53 |


osanoj (approved):
🎉

BjornRuud (approved):
Nice debugging! 🐛

PR #1309, opened 2024-01-10
Prepare for new Tori colors and update FINN dark mode blue color

ninarg:
# Why?

Some of our semantic colors don't make as much sense anymore with the transition to Warp.

Warp has a button which is called primary/CTA button. This button is watermelon in Tori, while all other branded colors are blueberry. In order to introduce this change into the app, we need to change so our `btnPrimary` (which is used on CTA buttons) to watermelon.

It seems like we have had a habit of using `btnPrimary` color in many of the cases where something is supposed to be blue in FINN, on all various of components. `btnPrimary` should only be used in the cases where there’s one **primary** action on the screen, e.g. CTA buttons. We have other semantic colors which are more suitable in other cases.

We also have an alternative to this button, `btnAction`, which looks exactly the same in FINN, except for a differing dark mode color. It has not been totally clear when this one should be used as opposed to `btnPrimary`. 

As explained, there's some confusion going on with our semantic colors, and I intend to clear that up in this PR. 

# What?

## ONE dark mode blue in FINN

FINN has had two versions of our dark mode blue. In the beginning of the color tranisition I tried not to mess that up and continue using the right one in the right contexts. After a discussion with UX, we found that both our blues are actually wrong according to the new design system, where there's a single blue color. So to make things easier for ourselves, I made sure to update FINN so we use this new dark mode blue everywhere.

## Deprecate `btnAction`

There's not really a thing called button action in the design system, and Android does not have this semantic color. In many cases it can be replaced with `textAction`, as it's simply used on a text link. So I decided to mark this one as deprecated, so we should migrate to other colors that make more sense based on the context. If we did not do this, developers could easily have picked the wrong color, not knowing `btnPrimary` and `btnAction` look equal in FINN, but completely different in Tori.

## Migrate `btnPrimary` to other semantic colors

Also, we make sure not to use `btnPrimary` in cases where the button is _not_ the primary/CTA button on the screen. Once this has been done, we can safely convert `btnPrimary` to watermelon in Tori.

## Migration of `btnAction` and `btnPrimary` usages

This PR migrates usages of `btnAction` and `btnPrimary`.
Both of these could be converted to any of these alternatives, based on the context they're used in, without any risk in FINN (they are the same in FINN):
- `textAction`: Used on text, e.g. our well known blue (flat) links without any background
- `nmpBrandControlSelected`: Used in a context where something is selected/activated, like toggles and range sliders
- `nmpBrandDecoration`: Used in other cases, for instance for tinting icons etc.  - where the color is just some decor

# Version Change

Breaking change, snapshots outside of this repo which uses the changed blue color in FINN will break.

# UI Changes

Due to the dark mode color change, almost every single dark mode snapshot had to be updated. As you can see in the screenshot, they look quite similar - but the blue is slightly changed.

A few light mode snapshots were also updated, I have no idea why, they have changes which I can't see with my own eyes, so I suspect a very slight difference.

Screenshot 2024-01-10 at 15 03 00


FinniversKit/Sources/Components/Switch/SwitchView.swift
```diff
@@ -39,7 +39,7 @@ public class SwitchView: UIView {
 
     private lazy var uiSwitch: UISwitch = {
         let uiSwitch = UISwitch(withAutoLayout: true)
-        uiSwitch.onTintColor = .btnPrimary //DARK
+        uiSwitch.onTintColor = .nmpBrandControlSelected //DARK
```
AndreyMomot, line 42: Can we remove unneeded comment here?

AndreyMomot (approved):
Great work! 
No typos was found during review, everything else hard to follow :)

FinniversKit/Sources/DNA/Color/Color+SwiftUI.swift
```diff
@@ -63,14 +65,16 @@ extension Color {
...
-    public static var borderDefault: Color       { Color(UIColor.borderDefault) }
+    public static var accentPrimaryBlue: Color          { Color(UIColor.accentPrimaryBlue)}
+    public static var accentSecondaryBlue: Color        { Color(UIColor.accentSecondaryBlue) }
+    public static var accentPea: Color                  { Color(UIColor.accentPea) }
+    public static var accentToothpaste: Color           { Color(UIColor.accentToothpaste) }
+    public static var tableViewSeparator: Color         { Color(UIColor.tableViewSeparator) }
+    public static var imageBorder: Color                { Color(UIColor.imageBorder) }
+    public static var decorationSubtle: Color           { Color(UIColor.decorationSubtle) }
+    public static var borderDefault: Color              { Color(UIColor.borderDefault) }
+    public static var nmpBrandControlSelected: Color    { Color(UIColor.nmpBrandControlSelected) }
```
BjornRuud, line 76: Do we need the `nmpBrand` prefix?

BjornRuud (approved):
👍

FinniversKit/Sources/DNA/Color/Color+SwiftUI.swift
ninarg: I don't think we do, I asked myself the same question when I saw it 🤔 Renaming it is out of scope for this PR and will make it a breaking change though. But perhaps I can add another public var without the prefix, so we can start using that one in upcoming PRs - instead of increasing usage of nmpBrand prefix (I will use controlSelected color in many more PRs to come).

FinniversKit/Sources/Components/Switch/SwitchView.swift
ninarg: haha sure, not sure what this comment means 🤷 I'll delete it.

FinniversKit/Sources/DNA/Color/Color+SwiftUI.swift
ninarg: I meant renaming the UIColor, that one is already used other places in the app.

mohsen-schibsted: We don't have a 1to1 mapping in Warp for these either, we are trying to map out all the usage of colors in FinniverseKit here https://docs.google.com/spreadsheets/d/1Q-Tr_dwJVfxgh3527IFjXQKqtnPPM42QeIr-M6q-zu8/edit#gid=2070969738

And on this draft PR https://github.com/finn-no/FinniversKit/pull/1290/files#diff-659017700820fe4dcb3aaffab44230d2e872c22784cb38cd47264e6b89d94328R12 you can see what we wil probably use instead of these colors later on

ninarg: Okay, I see _all_ these colors will be deprecated once we introduce Warp. So to avoid doing any extra work that will be overwritten soon anyways, I'll just keep them as they are right now. Thanks for the input, @mohsen-schibsted 😄

PR #1310, opened 2024-01-16
actually force remove the callout from the view

teddyfahi:
# Why?
The CalloutView is not actually removed from the view on hide, only the alpha value is set to zero. This is creating problems for callouts when they are opened via notifications.

# What?
After the fade-out animation is complete, the `removeFromSuperview()` method is called as a completion. This ensures that the view is not only hidden but also removed from the view hierarchy.


ag12 (approved):
⚡

PR #1313, opened 2024-01-17
Sponsored ribbon color updates

ninarg:
# Why?

We currently use accentToothpaste color for sponsored ribbons. This semantic color is watermelon in Tori, but sponsored ribbons shouldn't be watermelon. 

# What?

We can't just change accentToothpaste to be something else in Tori, since it's used many other places as well. So to fix this I added a new semantic color which is a part of Warp. It's called `bgSponsored`. 

I updated FINN to use Warp colors for `bgSponsored`, so now the ribbon is dark mode compatible. 

# Version Change

Breaking change.

# UI Changes

| Before | After |
| --- | --- |
| Screenshot 2024-01-17 at 13 16 40 | Screenshot 2024-01-17 at 13 16 49|


ChristianTherkelsen (approved):
Looks good! 🤓 👌 ✅

PR #1315, opened 2024-01-19
Bump simulator to iPhone 15 Pro, iOS 17 and Xcode to 15.2.0

ninarg:
# Why?

We should be up to date on latest iOS versions.

# What?

- CircleCI runs Xcode 15.2.0 with iPhone 15 Pro (iOS 17.2) simulator.

When bumping to a new major iOS version, there's always minor differences in snapshots, so they all have to be rerecorded. 
I checked the diff in them, there's mostly tiny spacing and text rendering differences.

There's a big diff on iPad for `AdRecommendationsGridView`, see screenshot below. But this should not block us from merging this. We have been building with Xcode 15 in production for a long time, so there shouldn't be any issues. When running the demo app on iPad, the feature looks fine for me.

# Version Change

Won't bump version.

# UI Changes

Snapshot diff for recommendations on iPad.

Screenshot 2024-01-19 at 11 12 31


osanoj (approved):
The Circle CI config looks good, I have not checked all of the 852 snapshot changes yet though, but they are probably fine... 😅

ninarg:
> We can probably add a jira task for looking into fixing the `AdRecommendationsGridView` snapshot at a later point

I had a quick look, running the Demo app on iPad iOS 16 and iOS 17, there's no difference between the OS versions. 
And both of them look equal to the new updated snapshot. 
So I think the new snapshot is an improvement from what we had before.

Screenshot 2024-01-19 at 12 27 35

So running on iPad, it looks like the image to the right ☝️

PR #1320, opened 2024-01-31
move unique resources into AdInsertion

thorbjxrn:
# Why?

These files seem to only be used in the AdInsertion and thus need not be in the shared resources. And then AdInsertion will be refactored further

# What?

Moving custom AdInsertion components. Button is changed to open from public

Mirror PR on https://github.schibsted.io/finn/ad-insertion-ios/pull/510

# Version Change

breaking!

# UI Changes

some UI components are moved


bstien (approved):
Nice😄🗑️ You should also remove the old snapshots related to the deleted demos, btw 🙂

PR #1321, opened 2024-01-31
Checkbox and radio buttons without animation

ninarg:
# Why?

FINN has branded and animated checkboxes and radio buttons. We used to animate using 10-20 images per animation.
The problem is all these images are branded, so they have a FINN color. We don't want to bring it into NMP, cause exporting all these animated images are time consuming, we don't want to do it for every brand due to UX resources etc. 
We don't get much benefit from it, your finger is often covering the animation. 🤷 

# What?

We have solved it by simplifying the radio button and checkbox. Now it will be a simple view where we toggle between 2 images. 

I had to rename the file names and also the functions, they are not animated anymore, so removed the `Animated` prefix. This makes it a breaking change. I will fix the renaming in other repos. Will open PRs once this one has been approved, to avoid doing extra work in case of changes required during review. 

Had to rerecord a bunch of snapshots which contain these views, there are minor differences which I can't see, probably in the icon.

# Version Change

Breaking.

# UI Changes

## Checkbox 

| Before | After |
| --- | --- |
| Screen Recording 2024-01-31 at 13 45 18 | Screen Recording 2024-01-31 at 13 40 26 |

## Radio button

| Before | After |
| --- | --- |
| Screen Recording 2024-01-31 at 13 47 39 | Screen Recording 2024-01-31 at 13 46 23 |


BjornRuud:
If we want the animations these aren't too difficult to replicate in code.

ninarg:
> If we want the animations these aren't too difficult to replicate in code.

Yeah, we talked about it, but decided to keep the scope down atm, since we have other things with higher priority.

AtishPapa:
Yes, as @BjornRuud mentioned there are `SwiftUICheckBox` and `SwiftUIRadioButton` but using those might involve some interfacing with SwiftUI. It might worth a try though. 🙂

AndreyMomot (approved):
Legendary way of doing this is pretty rectilinear :)
It can be done in code better
And possible trying to wrap SwiftUI's solution is an easy way
However can be merged if needed as it is with creating a low priority task for future

ninarg:
Thanks for the input! 
The original plan was to merge this as is due to priorities, but make a task for our backlog, so it can be picked up again whenever it makes sense for our team. Will include the suggestions in the task. 👍

ninarg:
Changed my mind regarding a task in the backlog. The design system which we'll get soon does have checkboxes and radio buttons (of course). ✨ So improving this one is not worth the effort. It will be deprecated very soon anyways.

ag12:
Do you know when you will update this for TJT @BjornRuud ?

BjornRuud:
> Do you know when you will update this for TJT @BjornRuud ?

@AtishPapa will have to answer that, I'm on TJM now. 😄

ag12:
Great

PR #1322, opened 2024-01-31
Remove "FINN" shapes from EmptyView

ninarg:
# Why?

We have a view which contains FINN branded shapes. We don't want to bring this into other brands. We have decided to remove the shapes, as this view is deprecated and no longer available in the design system.

# What?

- Remove code related to the shapes
- Rerecord snapshots

# Version Change

Minor. Made it non-breaking by deprecating current init, instead of removing it.

# UI Changes

| Before | After |
| --- | --- |
| Simulator Screenshot - iPhone 15 Pro - 2024-01-31 at 15 23 52 | Simulator Screenshot - iPhone 15 Pro - 2024-01-31 at 15 24 50 |


bstien (approved):
# 💃 🔥🗑️🔥 🕺

PR #1323, opened 2024-02-05
Color adjustments: Loading image color & callouts

ninarg:
# Why?

We need to make some extra color adjustments to accommodate for other brands. 

# What?

- Remove FINN branded loading colors, use plain `bgTertiary` instead
- Add new semantic colors for callout background and border, so they can be configured for each brand

# Version Change

Breaking.

# UI Changes

See newly recorded snapshots

Before - After
Screenshot 2024-02-05 at 10 53 11


ag12 (approved):
⚡

teddyfahi (approved):
🎨

PR #1324, opened 2024-02-06
frontpage-recommendations-layout-fix

adriansergheev:
# Why?
- fix a layout bug in recommendations

# What?
- `ExternalAdRecommendationCell` image would render taller than the cell it is enclosed in which causes a layout bug. I replaced that logic so it works akin to `JobAdRecommendationCell` i.e with a height multiplier (1).
- moved one `ExternalAdRecommendationCell` up in the demo view so it shows up in the snapshot tests (it did not before)

# Version Change
- minor 

# UI Changes
| Before | After |
| --- | --- |
| invalid | valid |


ag12 (approved):
👍

MarenZR (approved):
💯

PR #1327, opened 2024-02-21
Add/place holder image no image

ag12:
# Why?
This particular placeholder image is used in many places. We want to make sure that we only have one version of it. 
Until Warp is in place. 

# What?
Make a public placeholder image class with the default placeholder image for ads with no images. 
Remove unused image assets 

# Version Change
BREAKING 🧠 

# UI Changes
None.


ag12:
1. Not sure where to add the test. making a view seemed too much for this not sure but open to suggestions. 
2. This can also be added to `DefaultImageProvider` open to suggestions there as well. 

cc: @ninarg

ninarg:
> 1. Not sure where to add the test. making a view seemed too much for this not sure but open to suggestions.
> 2. This can also be added to `DefaultImageProvider` open to suggestions there as well.
> 
> cc: @ninarg

hmm, maybe `DefaultImageProvider` can be a good place to add it 🤔  Then it would be super easy to customise it for Tori, or any other brand later, if needed. And that's kinda a common place for all public images too, so I like that idea, it's more visible and easier to find, when it's not hidden alone like this.
Just beware, then this becomes a breaking change, since Tori has to inject an image in `ToriImageProvider` (it can inject `DefaultImageProvider.shared.noImage`, so no need to make it branded for now, and no duplicate image either)

ag12:
> > 1. Not sure where to add the test. making a view seemed too much for this not sure but open to suggestions.
> > 2. This can also be added to `DefaultImageProvider` open to suggestions there as well.
> > 
> > cc: @ninarg
> 
> hmm, maybe `DefaultImageProvider` can be a good place to add it 🤔 Then it would be super easy to customise it for Tori, or any other brand later, if needed. And that's kinda a common place for all public images too, so I like that idea, it's more visible and easier to find, when it's not hidden alone like this. Just beware, then this becomes a breaking change, since Tori has to inject an image in `ToriImageProvider` (it can inject `DefaultImageProvider.shared.noImage`, so no need to make it branded for now, and no duplicate image either)

Cool. Yes 
Something like this than?
```
    public var placeholderAdImage: UIImage {
        DefaultImageProvider.shared.placeholderAdImage // Default placeholder for ad image
    }
```

ag12:
addressed PR comments. Just wondering... do we have tests for imageprovider?

ninarg:
> addressed PR comments. Just wondering... do we have tests for imageprovider?

I don't think we do, so it's fine not to have any tests for this either, until we add it for all images!

ninarg (approved):
nice 🌟

ninarg:
Maybe it could be called just `placeholder`? It's obviously an image, none of the other images are suffixed with image 😄

ag12:
> placeholder

sure 😹

osanoj:
> Maybe it could be called just `placeholder`? It's obviously an image, none of the other images are suffixed with image 😄

But isn't it for ads only actually? So `placeholderAd`? 😛

ninarg:
> > Maybe it could be called just `placeholder`? It's obviously an image, none of the other images are suffixed with image 😄
> 
> But isn't it for ads only actually? So `placeholderAd`? 😛

is it? not sure 😂 then it makes sense 🤔 
but naming is HARD

PR #1329, opened 2024-02-23
replace forc_unwrapp with default fonts

AndreyMomot:
# Why?

For some reason Charcoal Demo 15.2.2 had a crash when open filters.
The issue was about not initialised Fonts. It happened ones, but it could be related to Tori testing session issue we had. 

# What?

Replace force unwrapped custom Fonts with default system similar options. 
Personally I don't see any drawbacks of this

# Version Change

Patch

# UI Changes

No changes

### DIFF
--- a/FinniversKit/Sources/DNA/Font/FontProvider.swift
+++ b/FinniversKit/Sources/DNA/Font/FontProvider.swift
@@ -38,7 +38,7 @@ public struct DefaultFontProvider: FontProvider {
     public var title1: UIFont {
         registerCustomFonts()
 
-        let font = UIFont(name: FontType.medium.rawValue, size: 34.0)! // swiftlint:disable:this force_unwrapping
+        let font = UIFont(name: FontType.medium.rawValue, size: 34.0) ?? UIFont.systemFont(ofSize: 34.0, weight: .medium)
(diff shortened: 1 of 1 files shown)


BjornRuud:
This will set system fonts when the code enters the same state that caused the crash, which will then continue and use the wrong font. Designers will not be happy. 😄

AndreyMomot:
As a temp fix to have a wrong font looks better than to have a crash. But this is just my opinion and I'm open to discussion

BjornRuud:
> As a temp fix to have a wrong font looks better than to have a crash. But this is just my opinion and I'm open to discussion

I agree, as long as we remember to follow up and figure out how to fix the crash or avoid it. 👍

AndreyMomot:
Yet this is the only drawback of this solution - smbd has to follow up the issue otherwise it became much less noticeable :)

Breiby:
If it only happened once, maybe it's best to leave it as is so that when we really do break fonts it'll be obvious?

AndreyMomot:
Decided to close this due to we want have more clear control of crashes and will add tracking to FinniversKit

AndreyMomot:
Decided to close this due to we want have more clear control of crashes and will add tracking to FinniversKit

PR #1330, opened 2024-03-01
TMP-2347 Transaction entry view moved

BjornRuud:
# Why?

The Mobility transaction entry view is not a common component and belongs in the MotorTransaction module.

# What?

Deleted transaction entry view and related files. They have been moved to the MotorTransaction module.

# Version Change

Breaking, removal of views.


robinsalehjan (approved):
Nice!

PR #1331, opened 2024-03-04
Remove version info footer from settings

ninarg:
# Why?

We no longer need to display app version info at the bottom of settings page. It's going to displayed elsewhere.

# Merged into this PR
- https://github.com/finn-no/FinniversKit/pull/1333

# What?

Removed version info from settings page.

# Version Change

Breaking.

# UI Changes

Screenshot 2024-03-04 at 11 01 23


BjornRuud (approved):
Can we also move the SettingsView cell to the settings module? Settings doesn't seem like it belongs in FinniversKit.

ninarg:
> Can we also move the SettingsView cell to the settings module? Settings doesn't seem like it belongs in FinniversKit.

Agree, there's many things in FinniversKit that don't belong here. But it's out of scope of my current work.

PR #1332, opened 2024-03-05
Selection list update

AtishPapa:
# Why?

I needed to add the accessibility description for each item in the selection list.

# What?

- Add an extra `accessibilityDescription` parameter to `SwiftUISelectionListItemModel`.
- Also added an extension to `View` for setting the foreground color without affecting SwiftUI animations.

# Version Change

Minor: Add accessibility description to SwiftUISelectionListItemModel.


FinniversKit/Sources/ViewModifiers/AdaptiveForegroundColorModifier.swift
```diff
@@ -0,0 +1,33 @@
...
+
+    @Environment(\.colorScheme) private var colorScheme
+
+    public func body(content: Content) -> some View {
+        content.foregroundColor(resolvedColor)
+    }
+
+    private var resolvedColor: Color {
+        switch colorScheme {
+        case .light:
```
BjornRuud, line 17: This case is unnecessary since the default is `.light`.

FinniversKit/Sources/ViewModifiers/AdaptiveForegroundColorModifier.swift
```diff
@@ -0,0 +1,33 @@
+import SwiftUI
+
+// Aplying a .foregroundColor modifier with DNA colors, e.g. .foregroundColor makes the animations glitchy.
+// When there's a possibility that the view's position can change, it's advised to apply the updated
+// .foregroundColor(dynamicColor:) modifier to avoid animation issues.
+public struct AdaptiveForegroundColorModifier: ViewModifier {
```
BjornRuud, line 6: Amazing that this is necessary, but here we are. 😄

BjornRuud (approved):

PR #1334, opened 2024-03-06
Fixes to bannerTransparencyModal in NMP 

andreas-jonasson-ios:
# Why?

We were seeing the incorrect logo inside the bannerTransparencyModal - also discovered we were redirecting to the wrong URLs. This PR aims to fix that. 

# What?

Added a: 
- correctly sized toriLogo to FinniverseKit 
- a switch to keep track of the current brand / target 
- updated redirect URLs inside the modal. 

# Version Change

Breaking change


FinniversKit/Sources/DNA/Images/ImageProvider.swift
```diff
@@ -121,4 +121,8 @@ public struct DefaultImageProvider: ImageProvider {
     public var brandMapDirections: UIImage {
         UIImage(named: .mapDirections)
     }
+    
+    public var toriLogo: UIImage {
```
BjornRuud, line 125: I don't think adding this here will be correct. This is the default ImageProvider which serves Finn branded images. The Tori logo should be served by `brandLogo` on the ImageProvider implementation for Tori (conveniently named `ToriImageProvider` in the app repo).

FinniversKit/Sources/Components/Consent/BannerTransparency/BannerTransparencyViewModel.swift
```diff
@@ -28,6 +30,13 @@ extension BannerTransparencyViewModel {
     }
 }
 
+// MARK: - LogoType Enum
+
+public enum LogoType {
```
BjornRuud, line 35: The FinniversKit components should in general not care about what specific type of images is being used. Could this be a more generic ID of some kind that is configured on setup?

FinniversKit/Sources/Components/Consent/BannerTransparency/BannerTransparencyViewModel.swift
```diff
@@ -28,6 +30,13 @@ extension BannerTransparencyViewModel {
...
+    case finn
+    case tori
```
pbodsk, line 37: It is a matter of taste but I would add a `logoImage` var to the enum and then move this

```
            switch model.logoType {
             case .finn:
                 logoImageView.image = UIImage(named: .finnLogo)
             case .tori:
                 logoImageView.image = UIImage(named: .toriLogo)
             }
```

inside of that. 

So something like: 

```
var logoImage: UIImage {
    switch self {
         case .finn:
             UIImage(named: .finnLogo)
         case .tori:
             UIImage(named: .toriLogo)
    }
}
```

And then you could use that like so in your `BannerTransparencyView`

`logoImageView.image = model.logoType.logoImage`

(OK lots if `image`s in there, maybe clear that up 😄 )

But again, your taste may vary 😄

pbodsk (approved):
Just a quick suggestion that you can just ignore 🤷

FinniversKit/Sources/Components/Consent/BannerTransparency/BannerTransparencyView.swift
```diff
@@ -46,6 +46,13 @@ public final class BannerTransparencyView: UIView {
                 return
             }
 
+            switch model.logoType {
+            case .finn:
+                logoImageView.image = UIImage(named: .finnLogo)
```
BjornRuud, line 51: The image should be configurable, so this view should be generic when it comes to logo type and configured for a brand where it is used.

FinniversKit/Sources/DNA/Images/ImageProvider.swift
andreas-jonasson-ios: Thank you for your input! Has been resolved now 🚀

FinniversKit/Sources/Components/Consent/BannerTransparency/BannerTransparencyViewModel.swift
andreas-jonasson-ios: Removed this part of the PR - thank you again for your input Björn :D

FinniversKit/Sources/Components/Consent/BannerTransparency/BannerTransparencyViewModel.swift
andreas-jonasson-ios: Simplified it based on Björns suggestions so have removed this now 😅  But thank you for your input!

BjornRuud (approved):
👍

PR #1336, opened 2024-03-19
frontpage-marketgrid-fix-label-iPad

adriansergheev:
# Why?
even though it should work as it is, [it was reported ](https://drive.google.com/file/d/1shBEIG20KzGB1XCv2Kwmt9W3ql-x3L_Z/view)that the labels for the marketgrid items get truncated on iPadOS for some reason

# What?
- increase the size of the cell for iPadOS to avoid the truncation.

# Version Change
- minor

# UI Changes
[see diff](https://github.com/finn-no/FinniversKit/pull/1336/files)


FinniversKit/Sources/Recycling/GridViews/Markets/GridView/MarketsGridView.swift
```diff
@@ -32,9 +32,14 @@ public class MarketsGridView: UIView, MarketsView {
 
     public var isMarketGridCellLabelTwoLined: Bool = false
     private var itemSize: CGSize {
-        isMarketGridCellLabelTwoLined
-        ? CGSize(width: 96, height: 88)
-        : CGSize(width: 92, height: 72)
+        switch (isMarketGridCellLabelTwoLined, UIDevice.isIPad()) {
+        case (true, true):
+            return CGSize(width: 128, height: 92)
```
BjornRuud, line 37: Why is the size fixed? It will be a problem for text scaling.

BjornRuud (approved):

FinniversKit/Sources/Recycling/GridViews/Markets/GridView/MarketsGridView.swift
adriansergheev: that's true, we just discussed this with the designer!
we agreed that it is a problem and agreed to solve it, but split this fix with the scaling solution
we're about to re-design the frontpage *sometime soon* and thought it's a great idea to have it done then

PR #1344, opened 2024-05-06
github-actions

adriansergheev:
# Why?
- we're moving from CircleCI towards Github Actions.

# What?
- github actions workflow
- drop CircleCI configuration files, disable webhook
- bump to Xcode 15.3

# Version Change

- none


FinniversKit.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -18,7 +17,16 @@
         "revision" : "b44b563e6ddc5a613a107fb57626047a7f1d4e1a",
         "version" : "1.9.666"
       }
+    },
```
adriansergheev, line 20: reverted package.resolved to the previous commit, since this build failed:
https://github.com/finn-no/FinniversKit/commit/7beb04072878004cf1913a56cbcf8c4bc43f810d

osanoj (approved):
Good job, let's try this out 👍

FinniversKit.xcworkspace/xcshareddata/swiftpm/Package.resolved
bstien: Since we now use Xcode 15.3 we should be fine with the current setup, and I guess it'll be updated by either of us during the next PR 🙂

PR #1345, opened 2024-05-06
upgrade snapshot tests dependency (take 2)

adriansergheev:
https://github.com/finn-no/DemoKit-iOS/pull/8

https://github.com/finn-no/FinniversKit/pull/1319
https://github.com/finn-no/FinniversKit/pull/1307


bstien:
I think we can/should change the runner to use [`macos-14-arm64`](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md) instead :)

bstien (approved):

PR #1350, opened 2024-06-05
show the button

thorbjxrn:
# Why?

We now want to show the text and button for viewings even after the date has passed.

# What?

reverting parts of [1340
](https://github.com/finn-no/FinniversKit/pull/1340/)
# Version Change

patch

# UI Changes

| Before | After |
| --- | --- |
|  |  |


teddyfahi (approved):
Har ikke helt fulgt tråden, og hva det landet på, men LGTM

PR #1352, opened 2024-06-14
Replace RefreshControl with native

AndreyMomot:
# Why?

Going away from FINN's style of RefreshConrol and UIActivityIndicatorView
[JIRA](https://nmp-jira.atlassian.net/browse/NMPAPP-1095)

# What?

Replace custom RefreshConrol with native
Instead of replacing all entities of LoadingIndicatorView for ios-app and SDKs I'm just making from it an extension of UIActivityIndicatorView in this PR

# Version Change

Patch

# UI Changes

| Before | After |
| --- | --- |
 | Simulator Screenshot - iPhone 15 Pro Max - 2024-06-14 at 15 00 31 |Simulator Screenshot - iPhone 15 Pro Max - 2024-06-14 at 14 36 35|
|Simulator Screenshot - iPhone 15 Pro Max - 2024-06-14 at 15 01 15|Simulator Screenshot - iPhone 15 Pro Max - 2024-06-14 at 14 36 43|

LOADER


osanoj:
> Just a question: should we remove `RefreshConrol` by itself now or just let it be?

I would say to remove it if we will not use it, might be good to keep it until you replaced all usages in all repos though - unless you already have?

osanoj (approved):

PR #1356, opened 2024-06-24
Get rid of code related to years old migration of loading indicators

ninarg:
# Why?

We found this var which is a remainder from a migration like 5 years ago.

# What?

Get rid of `shouldUseOldIndicator`. We can default to `.nmpBrandDecoration` which is just another shade of blue than we used to have.

# Version Change

Breaking.

# UI Changes

There's actually no UI change, due to the loading view always being updated before display, while we only changed the initial/default value.


ChristianTherkelsen (approved):
LGTM :shipit:

SaskaRad (approved):
Nice!

PR #1358, opened 2024-07-01
Make UIColor extensions internal

dinsen:
# Why?

I'm facing this nice error: `Ambiguous use of 'init(hex:)'` which is available both in `FinniversKit` and in `Warp`. 
So let's start removing the use of this extensions outside `FinniversKit`.

# What?

Make `init(hex:) and init(r:g:b:)` internal.

# Version Change

Major change, since it could break codebases 🙈 

# UI Changes

No UI changes

### DIFF
--- a/FinniversKit/Sources/DNA/Color/UIColor+FinniversKit.swift
+++ b/FinniversKit/Sources/DNA/Color/UIColor+FinniversKit.swift
@@ -487,7 +487,7 @@ public extension UIColor {
     /// Base initializer, it creates an instance of `UIColor` using an HEX string.
     ///
     /// - Parameter hex: The base HEX string to create the color.
-    convenience init(hex: String) {
+    internal convenience init(hex: String) {
         let noHashString = hex.replacingOccurrences(of: "#", with: "")
         let scanner = Scanner(string: noHashString)
         scanner.charactersToBeSkipped = CharacterSet.symbols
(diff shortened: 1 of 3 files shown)


FinniversKit/Sources/DNA/Color/UIColor+FinniversKit.swift
```diff
@@ -480,14 +480,14 @@ public extension UIColor {
     ///   - g: green (0-255)
     ///   - b: blue (0-255)
     ///   - a: alpla (0-1)
-    convenience init(r: CGFloat, g: CGFloat, b: CGFloat, a: CGFloat = 1.0) { // swiftlint:disable:this identifier_name
+    internal convenience init(r: CGFloat, g: CGFloat, b: CGFloat, a: CGFloat = 1.0) { // swiftlint:disable:this identifier_name
```
adriansergheev, line 483: having it empty is equal to having it internal

adriansergheev (approved):

FinniversKit/Sources/DNA/Color/UIColor+FinniversKit.swift
adriansergheev: you could also specify the module you want and then call the corresponding method i think

adriansergheev:
you could also specify the module you want and then call the corresponding method i think

FinniversKit/Sources/DNA/Color/UIColor+FinniversKit.swift
dinsen: I actually tried specifying the module, but the compiler was not happy 😞

PR #1359, opened 2024-07-03
Use warp colors

dinsen:
# Why?

# What?

I know it is lots of file changes, but it is mostly find and replace of `spacing` and `colors` from the old system to WARP. I have deprecated the old colors and added extensions for colors. My suggestion is removing them later on and directly use WARP. That will come later on.

# Version Change

This is breaking change, lots of colors has changed, specially in dark mode. After we merge this we are gonna move to the other repos and then release a new version with the update colors so designers can test and approve before the main release to the users.

# UI Changes

There are lots of UI changes, mainly color changes. You can see them in the snapshots.


pbodsk (approved):
Yeah...I mean, I haven't looked meticulously at _every_ single file...so sue me 🤷 

But, I guess the important thing is. You updated the warp package and renamed all occurrences accordingly to that...that should do for a first pass, then we can come back later if we discover nitty gritty details later on in some corner scenarios.

Good work on this :clap: Not a fun task but important to get done to make our work easier in the future

PR #1360, opened 2024-07-17
Fix Warp design review

milad-schibsted:
# Why?

# What?

After a design review with our design team at WARP we changed some colors.

# Version Change

Continuation of this [PR](https://github.com/finn-no/FinniversKit/pull/1359)

# UI Changes

There are lots of UI changes, mainly color changes. You can see them in the snapshots.


FinniversKit/Sources/Components/Feedback/FeedbackView.swift
```diff
@@ -359,8 +359,8 @@ private class ButtonView: UIView {
 private extension Button.Style {
     static func negativeButton() -> Button.Style {
         Button.Style.default.overrideStyle(
-            borderWidth: 1.0,
-            borderColor: .backgroundPrimary
+            borderWidth: 2.0,
```
ag12, line 362: should the borderWiidth change from 1 to 2?

ag12 (approved):
Good job. Just a minor comment regarding border width

FinniversKit/Sources/Components/Feedback/FeedbackView.swift
mohsen-schibsted: yes that was intentional.

PR #1368, opened 2024-08-06
Use component tokens for button

mohsen-schibsted:
# Why?

# What?

Fixing some wrong `colors`.

# Changelog

# Tests

No

# UI Changes

There are lots of UI changes, mainly color changes.


dinsen (approved):
🚀

PR #1369, opened 2024-08-13
recommendations-iPad-fix

adriansergheev:
# Why?
- there's a crash on iPads when launching the app from background
```
Fatal Exception: NSInternalInconsistencyException
UICollectionView internal bug: Attempting to create a cell with invalid attributes. Collection View: <UICollectionView: 0x108dedc00; frame = (0 0; 1376 888); clipsToBounds = YES; gestureRecognizers = <NSArray: 0x301442940>; backgroundColor = UIExtendedGrayColorSpace 0 0; layer = <CALayer: 0x301a57660>; contentOffset: {0, 30996}; contentSize: {1376, 33730.5}; adjustedContentInset: {0, 0, 0, 0}; layout: <FinniversKit.AdRecommendationsGridViewLayout: 0x11a2f92c0>; dataSource: <FinniversKit.AdRecommendationsGridView: 0x11a28ec00; frame = (0 0; 1376 888); layer = <CALayer: 0x301a57420>>>, Attributes: <UICollectionViewLayoutAttributes: 0x13b4fcc00; index path: (0-0); element kind: (UICollectionElementKindSectionHeader); frame = (0 0; 1376 252)>, Index Path: <NSIndexPath: 0x8c246f538daf20e4> {length = 2, path = 0 - 0}
```

# What?
- improve the lookup in the layout logic for `AdRecommendationsGridViewLayout`

# Version Change
patch

# UI Changes
- none

### DIFF
--- a/FinniversKit/Sources/Recycling/GridViews/AdRecommendations/GridView/AdRecommendationsGridView.swift
+++ b/FinniversKit/Sources/Recycling/GridViews/AdRecommendations/GridView/AdRecommendationsGridView.swift
@@ -51,15 +51,20 @@ public class AdRecommendationsGridView: UIView {
 
     private let imageCache = ImageMemoryCache()
 
-    private lazy var collectionViewLayout: AdRecommendationsGridViewLayout = {
-        let layout = AdRecommendationsGridViewLayout()
-        layout.delegate = self
-        return layout
-    }()
+//    private lazy var collectionViewLayout: AdRecommendationsGridViewLayout = {
+//        let layout = AdRecommendationsGridViewLayout()
+//        layout.delegate = self
+//        return layout
+//    }()
 
     // Have the collection view be private so nobody messes with it.
     public private(set) lazy var collectionView: UICollectionView = {
+        let collectionViewLayout = AdRecommendationsGridViewLayout()
+        collectionViewLayout.delegate = self
         let collectionView = UICollectionView(frame: .zero, collectionViewLayout: collectionViewLayout)
+//        let layout = UICollectionViewFlowLayout()
+//        layout.itemSize = CGSize(width: 100, height: 100)
+//        let collectionView = UICollectionView(frame: .zero, collectionViewLayout: layout)
         collectionView.translatesAutoresizingMaskIntoConstraints = false
         collectionView.delegate = self
         collectionView.dataSource = self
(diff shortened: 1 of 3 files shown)


FinniversKit/Sources/Recycling/GridViews/AdRecommendations/GridView/AdRecommendationsGridView.swift
```diff
@@ -51,15 +47,14 @@ public class AdRecommendationsGridView: UIView {
 
     private let imageCache = ImageMemoryCache()
 
-    private lazy var collectionViewLayout: AdRecommendationsGridViewLayout = {
```
adriansergheev, line 54: since collectionView is a mutable var, it's best to re-create the layout

FinniversKit/Sources/Recycling/GridViews/AdRecommendations/GridView/AdRecommendationsGridViewLayout.swift
```diff
@@ -122,7 +118,13 @@ class AdRecommendationsGridViewLayout: UICollectionViewLayout {
     }
 
     override func layoutAttributesForItem(at indexPath: IndexPath) -> UICollectionViewLayoutAttributes? {
-        return itemAttributes[indexPath.row]
+        guard itemAttributes.indices.contains(indexPath.row)
+        else { return nil }
+        if itemAttributes[indexPath.row].representedElementCategory == .cell {
```
adriansergheev, line 123: `layoutAttributesForItem ` retrieves layout info at the indexPath for cells only
we create above a `UICollectionViewLayoutAttributes` for the header which is stored in the same array

for some reason, on iPads when launching the app from background this method is called with indexPath.row == 0 which would cause a crash, since at index 0 in the array we have a layout attribute for a header and not for a cell
on iPhones, that does not happen 🤷‍♂️

FinniversKit/Sources/Recycling/GridViews/AdRecommendations/GridView/AdRecommendationsGridViewLayout.swift
```diff
@@ -122,7 +118,13 @@ class AdRecommendationsGridViewLayout: UICollectionViewLayout {
...
+            return itemAttributes[indexPath.row]
+        } else {
+            return nil
+        }
```
osanoj, line 127:
```suggestion
        if let itemAttribute = itemAttributes[safe: indexPath.row], itemAttribute.representedElementCategory == .cell {
            return itemAttribute
        } else {
            return nil
        }
```

ninarg (approved):
Great work 🙌

PR #1370, opened 2024-08-14
Remove color provider

mohsen-schibsted:
# Why?

# What?

Removing Finniverskit ColorProvider (and some more cleaning up, some colors are kept because we use them for a decoder mapper, will remove them next time since I need to change the decoder also) so everywhere we are forced to use WARP colors.

# Changelog

# Tests

No

# UI Changes

There are lots of UI changes, mainly color changes.


FinniversKit.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -33,8 +33,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/warp-ds/warp-ios.git",
       "state" : {
-        "revision" : "ae6d80f9fd3477ebf487cc060306697b557cc532",
-        "version" : "0.0.22"
+        "branch" : "remove-warp-config",
```
dinsen, line 36: Remember to change to latest version when Warp is released

dinsen (approved):
🚀

AndreyMomot (approved):
Great stuff!

PR #1372, opened 2024-08-22
Use span mapper for HTMLText in SwiftUI similar to UIKit

AtishPapa:
# Why?

The class `HTMLText` had a hard coded colour specific to TJT and since we changed that colour recently it was impossible to get it to the view.

# What?

A new span mapper parameter is added to the class enabling configuring the specific section colours from the "outside"

# Version Change

Breaking: Make it possible to configure span colours in HTMLText.

### DIFF
--- a/Demo/Sources/SwiftUI/HTMLTextDemoViewController.swift
+++ b/Demo/Sources/SwiftUI/HTMLTextDemoViewController.swift
@@ -3,6 +3,7 @@ import Foundation
 import SwiftUI
 import UIKit
 import DemoKit
+import Warp
 
 final class HTMLTextDemoViewController: UIViewController {
     let hostingController: UIHostingController<HTMLText>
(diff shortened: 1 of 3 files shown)


Demo/Sources/SwiftUI/HTMLTextDemoViewController.swift
```diff
@@ -3,6 +3,7 @@ import Foundation
 import SwiftUI
 import UIKit
```
shilena91, line 4: probably we don't need to import UIKit when we already have SwiftUI.

Demo/Sources/SwiftUI/HTMLTextDemoViewController.swift
```diff
@@ -50,17 +51,34 @@ extension HTMLTextDemoViewController: TweakableDemo {
     }
 
     func configure(forTweakAt index: Int) {
+        let spanMapper: HTMLStringSwiftUIStyleTranslator.SpanMapper = { attributes, currentStyle in
+            for attribute in attributes {
+                guard attribute.name == "style",
+                      attribute.value == "color:tjt-price-highlight" else {
```
shilena91, line 57: would be nice not to hardcode this here and in HTMLText as well

shilena91 (approved):
LGTM, just some nitpickings 👍

Demo/Sources/SwiftUI/HTMLTextDemoViewController.swift
AtishPapa: For the demo I think it's OK, since the purpose is to test the span mapping. I could call it differently, I just didn't want to change the test HTML strings. 🙂

PR #1375, opened 2024-08-28
feat: Adds support for anchor tags in HTMLText

pbodsk:
# Why?

Because I needed it 🤷 

OK then...in an error screen we need to present a localized text with a link to our help center, dependent on country. When you click that link you are taken to the correct help center in a browser.

The easiest would be if we could use HTML for this and then use an anchor tag to control where to go. That way we don't need to have two separate keys in our localization and do all kinds of combining, tap gestures and so on.

Sadly our `HTMLText` SwiftUI view does not support HTML attributes...till now 😅 

# What?

## Reading Order 👀 
I managed to extend the existing `HTMLStringSwiftUIStyleTranslator` class with a new capability to also read and store attributes so that would be a good place to start.

You will notice that the internal `StyledText` has been extended with a new array of `Attribute` elements like so:

```swift
public let attributes: [Attribute]?
```

And `Attribute` is defined like so:

```swift
extension HTMLStringSwiftUIStyleTranslator {
    public enum Attribute: Equatable {
        /*
         For now we just have a URL attribute here (which is read from the href attribute of an anchor tag)
         but this could be extended if so needed
         */
        case url(value: URL)
    }
}
```

So...at the moment we only support URL's but that can be extended if so needed later on.

### OK so...
We are now able to send potential `Attribute`s as part of the parsing, how do we do that? 

Still in `HTMLStringSwiftUIStyleTranslator` take a look at `translate(tokens: )` where there are the following changes to the `switch`:
- the `.tagStart` arm has been extended to now also handle `a` tags. All we do here is store a URL for the current href value on a new `attributeStack` which is somewhat similar to the existing `styleStack` (and yet different...basically it is a simple Stack datastructure used to hold the current Attribute)
- the `.text` arm has been extended to look for attributes, and if any are present, they are added as part of the `StyledText` element we create. Note that we don't pop the attributes here...we just `peek` them
- in the `.tagEnd` arm we clean up any lingering attributes...we do that here because here we have a name for them, which makes it easier when querying

Here's how that looks in code:

```swift
public func translate(tokens: [HTMLToken]) throws -> [StyledText] {
    var styledText: [StyledText] = []
    for token in tokens {
        switch token {
        case .tagStart(let name, let attributes, _):
            switch name.lowercased() {
            case "br":
                styledText.append(StyledText(text: "\n", style: styleStack.currentStyle))
                continue
            case "a":
                if let urlString = attributes.first(where: { $0.name.lowercased() == "href"})?.value,
                   let url = URL(string: urlString) {
                    attributeStack.pushAttribute(.url(value: url), elementName: name)
                }
            default:
                break
            }
            let styleMapper = self.styleMapper ?? defaultStyleMapper
            if let style = styleMapper(name, attributes) {
                styleStack.pushStyle(style, elementName: name)
            }
        case .tagEnd(let name):
            attributeStack.popAttribute(elementName: name)
            styleStack.popStyle(elementName: name)
        case .text(let text):
            if let lastAttribute = attributeStack.peek() {
                styledText.append(.init(text: text, style: styleStack.currentStyle, attributes: [lastAttribute]))
            } else {
                styledText.append(.init(text: text, style: styleStack.currentStyle, attributes: nil))
            }
        default:
            break
        }
    }
    return styledText
}
```

If you are curious about the `attributeStack`, you can find it in the `HTMLAttributeStack` file.

### OK so...
Now we have parsed the HTML and built a structure for each token. Next we want to actually use that to build our `Text` items.

That takes place in `HTMLText` where we now check if we have a URL or not. If we do not have a URL it is business as usual, and if we have a URL we can generate a `Text` element with that URL, using a new extension to `StringInterpolation`

First, here's how the `reduce` from styledTexts into a combined `Text` looks.

```swift
return styledTexts.reduce(
    into: Text("").applyStyle(translator.styleStack.defaultStyle)
) { textView, styledText in
    let newText: Text
    if let url = styledText.url {
        newText = Text("\(styledText.text, link: url)")
    } else {
        newText = Text(styledText.text)
    }
    textView = textView + newText
        .applyStyle(styledText.style)
}
```

And then the linchpin that ties all this together in `LocalizedStringKey+Extensions`!

```swift
mutating func appendInterpolation(_ linkTitle: String, link url: URL) {
    var linkString = AttributedString(linkTitle)
    linkString.link = url
    self.appendInterpolation(linkString)
}
```

With this extension we are able to actually create `Text` elements from `URL`s...and with that we can stay in a "pure Text" environment...thereby allowing us to use the + operator overload for `Text` 😅  

### Other Elements
- There are a couple of new unit tests in `HTMLStringSwiftUIStyleTranslatorTests.swift`
 
# Version Change

This is a minor change...we have a new option that doesn't break the existing usage.

# UI Changes

| Code | Output |
| --- | --- |
|Screenshot 2024-08-28 at 14 58 15 | Simulator Screen Recording - iPhone 15 - 2024-08-28 at 14 59 09 |

### DIFF
--- a/FinniversKit/FinniversKit.xcodeproj/project.pbxproj
+++ b/FinniversKit/FinniversKit.xcodeproj/project.pbxproj
@@ -306,6 +306,8 @@
 		B26B582E2B4BD695000E0245 /* NMPInfoboxView.swift in Sources */ = {isa = PBXBuildFile; fileRef = B26B582C2B4BD695000E0245 /* NMPInfoboxView.swift */; };
 		B26B582F2B4BD695000E0245 /* NMPInfoboxView+InformationType.swift in Sources */ = {isa = PBXBuildFile; fileRef = B26B582D2B4BD695000E0245 /* NMPInfoboxView+InformationType.swift */; };
 		B26B58312B4BDD15000E0245 /* InfoboxSwiftUIView.swift in Sources */ = {isa = PBXBuildFile; fileRef = B26B58302B4BDD15000E0245 /* InfoboxSwiftUIView.swift */; };
+		B2829E2D2C7F054F0032BDFA /* HTMLAttributeStack.swift in Sources */ = {isa = PBXBuildFile; fileRef = B2829E2C2C7F054F0032BDFA /* HTMLAttributeStack.swift */; };
+		B2829E2F2C7F05D20032BDFA /* LocalizedStringKey+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = B2829E2E2C7F05D20032BDFA /* LocalizedStringKey+Extensions.swift */; };
 		B6801C1C2982ECCA004A0F4F /* SettingsHeaderView.swift in Sources */ = {isa = PBXBuildFile; fileRef = B6801C1B2982ECCA004A0F4F /* SettingsHeaderView.swift */; };
 		CBD2F08D20615990002AA385 /* TextField+State.swift in Sources */ = {isa = PBXBuildFile; fileRef = CBD2F08C20615990002AA385 /* TextField+State.swift */; };
 		CF05B7E023325968003A7F1C /* FavoriteFolderActionButton.swift in Sources */ = {isa = PBXBuildFile; fileRef = CF05B7DF23325968003A7F1C /* FavoriteFolderActionButton.swift */; };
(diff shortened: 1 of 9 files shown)


agusti (approved):
LGTM 🎉

pbodsk:
> LGTM 🎉

Long Gone Talk Monday 😉 

Yes I know...Looks Good To Me...but I always think Long Gone Talk Monday 😄

PR #1377, opened 2024-09-04
Remove font provider

mohsen-schibsted:
# Why?

We are going to start using WARP Typography instead of old fonts, to make it easier to move forward with other brands.

# What?

Removing Finniverskit FontProvider so everywhere we are forced to use WARP Typography.

# Version Change

Since it is changing the way we set the fonts it should be major

# UI Changes

Changes in some fonts


FinniversKit/Sources/DNA/Font/UIFont+FinniversKit.swift
```diff
@@ -3,118 +3,89 @@
...
 
+    /// Font Medium/Bold with a size of 16 scaled for UIFontTextStyle.callout
+    ///
     /// ## Usage:
-    /// - If a T2 have more sublevels, you can use T3.
-    class var title3: UIFont {
-        Config.fontProvider.title3
+    /// - This have the same size as the body text, but is always bolded (Medium) to differenciate them.
+    public static var bodyStrong: UIFont {
+        Warp.Typography.title4.uiFont
```
ag12, line 43: just checking ...        ` Config.fontProvider.title3 ==  Warp.Typography.title4.uiFont`

ag12 (approved):
legit. Added a comment.

FinniversKit/Sources/DNA/Font/UIFont+FinniversKit.swift
mohsen-schibsted: Probably it shows it like that because we removed `title3Strong` and we only have `title3`, and `bodyStrong` is `title4`. So the result is okey :D the change log is a bit strange.

PR #1378, opened 2024-09-12
brazed-braze

adriansergheev:
# Why?
- move braze ui to the internal repo

# What?
-  braze implementation was removed from finniverskit and moved to the internal repo

# Version Change

major, braze is dropped


agusti (approved):
LGTM 🎉

PR #1383, opened 2024-10-16
Fix: Bump to Xcode 16 / iOS 18

bstien:
# Why?
We're ready to step into the future and migrate to Xcode 16.

# What?
- ~Bump platform in `Package.swift` to iOS 16.~
	- **Reverted**. Several other modules still have iOS 15.
- Set deployment target in Demo project to iOS 16.
- Update GHA to use Xcode 16 and iPhone 16 Pro (iOS 18.0).
- Retake snapshots.
  - Good luck finding any differences 🍀

# Version Change
None – there are no changes to the module itself.

# UI Changes
None.


Demo/Demo.xcodeproj/project.pbxproj
```diff
@@ -2064,7 +2064,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 15.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 16.6;
```
adriansergheev, line 2067: shouldn't it be 16.0?

bstien: That's the value it sets when selecting `iOS 16` from the dropdown. This is only for the Demo, so I won't bother fixing it manually.

Screenshot 2024-10-16 at 14 38 48

PR #1384, opened 2024-10-16
Remove extra layer of logic for fonts

mohsen-schibsted:
# Why?

Simplify font structure.

# What?

There is an extra layer in our Typography structure. We will try to remove that on this branch.

# Version Change

Major

# UI Changes

There should not be any UI changes

It is interesting that we get some UI changes on the `snapshots` but in reality when you run and take a `screenshots` from `master` and `this branch` they are NO changes.

And since the `screenshots` look like the `new snapshot`, as you can see below, I would say something was wrong with the old one.

New screenshot

Simulator Screenshot - iPhone 15 Pro - 2024-10-16 at 18 19 12

Old and New snapshots

Screenshot 2024-10-16 at 18 19 30


Demo/SnapshotTests/__Snapshots__/SwiftUIViewTests/ToastSwiftUIDemoView_Previews.dark_iPhone.png
Breiby, line 1: Text should probably not be truncated here

mohsen-schibsted: In reality it is not truncated. This is the screenshot of the demoView. Interesting that it shows it like that on Snapshots 🤔

Simulator Screenshot - iPhone 15 Pro - 2024-10-18 at 11 41 45

Breiby: Strange! Have you checked if the same discrepancies appear if you re-record snapshots without your changes?

mohsen-schibsted: My guess is the way live view renders the screen is a bit different that how snapshots does. I was able to fix it by adding `.fixedSize(horizontal: false, vertical: true)` to it. Fixes the snapshot and does not change the live view.

Breiby (approved):

Breiby:
I think it'd be nice to figure out what causes the different behaviour in the snapshots, but seems like everything works as expected in the main app so ✅

PR #1387, opened 2024-11-12
fix voice over for finnish

AndreyMomot:
# Why?

We wanted to remove emojies symbols for norwegian but removed accidentally some finnish symbols (`ä ` particularly) 

# What?

Remove broken method 

# Version Change

Minor


ag12 (approved):
🎸

PR #1389, opened 2024-11-19
Fix collapsed native ads 

andreas-jonasson-ios-main:
# Why?

We were seeing collapsed native ads in certain verticals. This simple code change aims to remedy this ! 

image

image


ag12 (approved):
⚡

charlieeddus (approved):

PR #1390, opened 2024-11-26
Update Warp

mohsen-schibsted:
# Why?

Did some breaking clean up on Warp, so some initializers are not available anymore.

# What?

Update `Warp`

# Version Change

Minor

# UI Changes

Should not cause any UI changes in the latest build, but since in 0.0.40 we updated some colors probably some Snapshots will break. Will update those.


ag12 (approved):
👍

PR #1392, opened 2024-11-29
Fix loan calculator background

mohsen-schibsted:
# Why?

After the changes in Warp Token values, `LoanCalculatorView` broke in `dark mode`. `Background` for the view is `backgroundSubtle` and `TrackColor` for `StepSlider` is `backgroundDisabledSubtle` which both are `gray750` in `dark mode`. See the images below:

Simulator Screenshot - iPhone 16 Pro - 2024-11-29 at 09 19 14
Simulator Screenshot - iPhone 16 Pro - 2024-11-29 at 09 19 30

# What?

So previously it looked like `Warp.Box` in `Neutral` type, so we decided to change it to look like `Bordered` type instead.

Screenshot 2024-11-29 at 10 30 41

# Version Change

Patch

# UI Changes

Now it looks like this:

Simulator Screenshot - iPhone 16 Pro - 2024-11-29 at 09 41 42
Simulator Screenshot - iPhone 16 Pro - 2024-11-29 at 09 41 57


thorbjxrn (approved):
looks good! and designer gives 👍

PR #1395, opened 2025-01-24
Remove the old Message form since it's been moved to Messaging SDK

charlieeddus:
# Why?

We [moved the Message form](https://github.schibsted.io/finn/messaging-ios-sdk/pull/486) to Messaging SDK and want to remove it from here.

# What?

- Remove Message form

# Version Change

Minor


ag12 (approved):
kult

PR #1396, opened 2025-01-28
Accessibility improvements to `Broadcast` component

robinsalehjan:
# Why?

- `Broadcast` component is missing several accessibility properties to make it easy to interact with.

# What?

- Create accessibility groups, traits and labels for the `Broadcast` and `BroadcastItem`
  - Accessibility agent will read out the labels
  - The close button is accessible

# Version Change

- Minor

# UI Changes

| Before | After |
| --- | --- |
| https://github.com/user-attachments/assets/71e54714-15e7-43cd-b2bd-3c2485f1a042 | https://github.com/user-attachments/assets/0fe1090c-7dba-4f79-9716-c4ba3ba51c0a |


AndreyMomot (approved):
👍

FinniversKit/Sources/Components/Broadcast/Broadcast.swift
```diff
@@ -171,12 +172,20 @@ private extension Broadcast {
     }
 
     func insert(_ messages: Set<BroadcastMessage>) {
-        for message in messages {
+        var accessibilityLabelForContainer = ""
+        
+        for (index, message) in messages.enumerated() {
             let item = BroadcastItem(message: message)
+            let accessibleLabelForItem = "Alert \(index + 1), \(item.message.text)"
```
osanoj, line 179: Shouldn't `"Alert"` be localized?

FinniversKit/Sources/Components/Broadcast/BroadcastItem.swift
```diff
@@ -54,6 +55,9 @@ class BroadcastItem: UIView {
         let button = UIButton(frame: .zero)
         button.setImage(UIImage(named: .remove), for: .normal)
         button.tintColor = .icon
+        button.isAccessibilityElement = true
+        button.accessibilityTraits = .button
+        button.accessibilityLabel = "Close broadcast"
```
osanoj, line 60: This should also be localized I guess

FinniversKit/Sources/Components/Broadcast/Broadcast.swift
robinsalehjan: Yes if we want to keep it it will have to be provided from the caller 👍

PR #1399, opened 2025-02-12
Add tooltip to KeyValueGridView

marianaSchibsted:
# Why?

A tooltip was requested on the Mobility Item Page to provide brief explanatory text for a specific specification.

# What?

- Added new optional `infoTooltip` property in `KeyValuePair`.

Tooltip implementation in `KeyValueGridView`:
- Displays an info icon next to the specification if `infoTooltip` is present.
- Shows a tooltip with `infoTooltip` content on click.
- Dismiss the tooltip on info icon click, on click on tooltip it self or on scroll.

# Version Change

Minor

# UI Changes

| iPhone | iPhone | iPhone |
| --- |  --- | --- |
| simulator_screenshot_59468D9B-A029-453E-8C2F-CF6979C139B4 | simulator_screenshot_6868887B-E0C5-494B-A564-2AA44F4D0F1B |  simulator_screenshot_A158E185-767C-488C-A2EF-896E60BACD14 | 

| iPad | iPad | 
| --- |  --- |
| simulator_screenshot_28BA6BE3-BCCB-4BE3-840A-74691B79FDC1 | simulator_screenshot_E79613B6-2142-4752-BDB3-D195F16FAA57 |  

The video to show how the tooltip disappears:

RocketSim_Recording_iPhone_14_Pro_Max_6 7_2025-03-21_14 23 00

**Accessibility**:
The tooltip is testet with Voice Over. 

https://github.com/user-attachments/assets/316bb1c9-cfcf-465a-9f70-05a7be39c4c8

### DIFF
--- a/FinniversKit/Sources/Components/KeyValueGridView/KeyValueGridView.swift
+++ b/FinniversKit/Sources/Components/KeyValueGridView/KeyValueGridView.swift
@@ -411,6 +411,6 @@ private class PaddableLabel: Label {
     }
 }
 
-extension Notification.Name {
+public extension Notification.Name {
     static let immediateDismissTooltip = Notification.Name("ImmediateDismissTooltip")
 }


FinniversKit/Sources/Components/KeyValueGridView/KeyValueGridView.swift
```diff
@@ -80,11 +80,41 @@ public class KeyValueGridView: UIView {
...
+        )
 
         let titleLabel = Label(style: titleStyle, numberOfLines: 2, withAutoLayout: true)
         titleLabel.lineBreakMode = .byWordWrapping
 
+        let titleContainer = UIStackView()
+        titleContainer.axis = .horizontal
+        titleContainer.alignment = .center
+        titleContainer.distribution = .fill
+        titleContainer.spacing = Warp.Spacing.spacing100
```
bstien, line 98: Nitpick, but we could use the initializer for this.

```swift
let titleContainer = UIStackView(
    axis: .horizontal,
    spacing: Warp.Spacing.spacing100,
    alignment: .center,
    distribution: .fill,
    withAutoLayout: true
)
```

FinniversKit/Sources/Components/KeyValueGridView/KeyValueGridView.swift
```diff
@@ -80,11 +80,41 @@ public class KeyValueGridView: UIView {
...
+            let infoButton = UIButton(type: .custom)
+            infoButton.setImage(Warp.Icon.info.uiImage, for: .normal)
+            infoButton.translatesAutoresizingMaskIntoConstraints = false
+            NSLayoutConstraint.activate([
+                infoButton.widthAnchor.constraint(equalToConstant: Warp.Spacing.spacing200),
+                infoButton.heightAnchor.constraint(equalToConstant: Warp.Spacing.spacing200)
+            ])
+            infoButton.addAction(UIAction(handler: { [weak self] _ in
+                self?.showTooltip(infoText, from: infoButton)
+            }), for: .touchUpInside)
```
bstien, line 113: This will create a strong reference to `infoButton` from the closure. We should either weakify it, or use `action.sender` instead.

```swift
infoButton.addAction(UIAction(handler: { [weak self] action in
    guard let view = action.sender as? UIView else { return }
    self?.showTooltip(infoText, from: view)
}), for: .touchUpInside)
```

bstien:
How does this look in the app if you show it on the itempage, and then scroll the screen? Does the tooltip scroll alongside, or is it "stuck" in its initial position?

FinniversKit/Sources/Extensions/UIKit/UIApplicationExtensions.swift
```diff
@@ -0,0 +1,10 @@
+import UIKit
+
+extension UIApplication {
+    public var firstWindow: UIWindow? {
+        if let scene = UIApplication.shared.connectedScenes.first as? UIWindowScene {
+            return scene.windows.filter { $0.isKeyWindow }.first
```
bstien, line 6:
```swift
scene.windows.first { $0.isKeyWindow }
```

FinniversKit/Sources/Components/KeyValueGridView/KeyValueGridView.swift
```diff
@@ -80,11 +80,41 @@ public class KeyValueGridView: UIView {
...
+        titleContainer.distribution = .fill
+        titleContainer.spacing = Warp.Spacing.spacing100
+
+        titleLabel.text = pair.title
+
+        titleContainer.addArrangedSubview(titleLabel)
+        if let infoText = pair.infoTooltip, !infoText.isEmpty {
+            let infoButton = UIButton(type: .custom)
+            infoButton.setImage(Warp.Icon.info.uiImage, for: .normal)
+            infoButton.translatesAutoresizingMaskIntoConstraints = false
```
bstien, line 106: There's no label for VoiceOver users for this button. How should the button and the presentation work for them?

marianaSchibsted:
> How does this look in the app if you show it on the itempage, and then scroll the screen? Does the tooltip scroll alongside, or is it "stuck" in its initial position?

It looks like this. And I am not able to scroll until I close the tooltip (by clicking on any part of the screen

https://github.com/user-attachments/assets/4f9dc908-8d17-435a-9c63-2de4f7de124e

FinniversKit/Sources/Components/KeyValueGridView/KeyValueGridView.swift
```diff
@@ -80,11 +80,45 @@ public class KeyValueGridView: UIView {
...
+            infoButton.accessibilityLabel = pair.infoTooltipAccessibilityLabel
+            NSLayoutConstraint.activate([
```
marianaSchibsted, line 110: @bstien Like this?

FinniversKit/Sources/Components/KeyValueGridView/KeyValueGridView.swift
```diff
@@ -119,6 +164,230 @@ public class KeyValueGridView: UIView {
...
+
+    private func findScrollView() -> UIScrollView? {
+        var view: UIView? = self
+        while view != nil {
+            if let scrollView = view as? UIScrollView {
+                return scrollView
+            }
+            view = view?.superview
+        }
+        return nil
```
bstien, line 358: Not a needed change at all, but we could use [`sequence(first:next:)`](https://developer.apple.com/documentation/swift/sequence(first:next:)) here instead to avoid the control variables.

```swift
for view in sequence(first: self, next: { $0.superview }) {
    if let scrollView = view as? UIScrollView {
        return scrollView
    }
}
return nil
```

FinniversKit/Sources/Components/KeyValueGridView/KeyValueGridView.swift
```diff
@@ -143,3 +412,7 @@ private class PaddableLabel: Label {
         super.drawText(in: rect.inset(by: textPadding))
     }
 }
+
+public extension Notification.Name {
+    static let ImmediateDismissTooltip = Notification.Name("ImmediateDismissTooltip")
+}
```
bstien, line 418: I don't think we need to have this public, unless you're using it in the app. And we should change the name be consistent in our use of lower camel case

```swift
extension Notification.Name {
    static let immediateDismissTooltip = Notification.Name("ImmediateDismissTooltip")
}
```

bstien (approved):
This looks AWESOME!!😄🙌 Great job adding this!
The nitpick comments can be ignored, if needed 🙂

FinniversKit/Sources/Components/KeyValueGridView/KeyValueGridView.swift
marianaSchibsted: It's used in the app

PR #1400, opened 2025-03-11
Change broadcast view icon to align with design

alishariatbahadori:
# Why?

The broadcast icon on iOS did not match the intended design. To ensure consistency, we are updating it to the Warp warning icon.

# What?
- Replaced the leading and dismiss icons with Warp icons.
- Adjusted view anchors to fix the different size of the Warp icon.

# UI Changes

| Before | After |
| --- | --- |
| simulator_screenshot_F634E339-43F8-45CB-9537-F2E0B7B65ABA | Simulator Screenshot - iPhone 16 - 2025-03-11 at 14 46 58 |


charlieeddus (approved):
Good stuff! 👏

charlieeddus (approved):
Nice updates! I think the new snapshots look better!

PR #1402, opened 2025-03-26
frontpage-accessibility-audit

adriansergheev:
# Why?
- this PR improves accessibility for frontpage (and other consumers of Recommendations views)

# What?
- add accessibility labels for each cell. Now instead of navigating via headers, users of voiceover can navigate between recommendation cells. This removes the need for multiple interactions for voice over navigation (`5.1`)
- add `UIAccessibilityCustomAction` for marking a recommendation as favorite and remove the button from accessibility perspective.  (`7.1`)
- now the image is also seen by voiceover and a description for it can be generated in select languages (none nordic unfortunately 😕) (`5.1`)
- increase tappable area for favorite button 

- <s>- fix font sizing for `FinniversKit.Label` (this affects the whole app! i.e where `Label` is used and if scaling is enabled) (`8.2`)]</s>

identifiers for issues i.e `x.x` are available in the internal sheet

# Version Change
- minor

# UI Changes
- <s>see snapshot diff. there's lots of subtle changes in the text size, I assume now the correct size is applied. (before it was fixed).
It is easier to see the changes in something like [kaleidoscope](https://kaleidoscope.app). </s>

ex:
image


FinniversKit/Sources/Recycling/GridViews/AdRecommendations/Cell/StandardAdRecommendationCell.swift
```diff
@@ -1,10 +1,27 @@
-//
-//  Copyright © FINN.no AS, Inc. All rights reserved.
-//
-
 import UIKit
 import Warp
 
+fileprivate func accessibilityMultiplier() -> CGFloat {
```
adriansergheev, line 4: this seems to be useful across the changes I made (see companion PR in main repo)
should it be exported by FinniversKit?

AndreyMomot (approved):
LGTM

PR #1408, opened 2025-05-20
Feature: DynamicStackView

bstien:
# Why?
We need to (finally) start supporting dynamic font sizes, and to lessen the burden on this migration I'm adding a helper view that we should (in a lot of cases) use to replace usages of `UIStackView` with a horizontal axis.

By specifying a value in `breakAtContentSize` you define at which levels we should use a **horizontal** axis and when we should use a **vertical** axis:
- If the current content size category is **less than** `breakAtContentSize`: we use a horizontal axis.
- If the current content size category is **equal to or greater than** `breakAtContentSize`: we use a vertical axis.

The view accepts a few parameters we often use to configure a `UIStackView`: spacing, alignment and distribution. More can be added when necessary. These parameters will either accept a value that'll be used for both axes, or individual values for either axis.

I've also added a delegate that will be notified when the axis changes, so we can act on these changes. Examples are adding/removing subviews, changing content of views or changing properties (like `Label.numberOfLines`).

To get started I've tested this on `FavoriteAdView`, which is used to present a list of favorited ads within a folder.

# What?
- Add `DynamicStackView`.
- Bump Warp to `0.0.61`.
- Bump DemoKit to `0.4.3`.
- Changes in `FavoriteAdView`:
	- Use `Label` instead of `UILabel`.
	- Replace horizontal stackview with `DynamicStackView`.
- Increase size of `textField` in `FavoriteAdCommentInputView`.
	- It's placed within a scroll view, but the initial low height made the textfield hard to see/reach.

# Version Change
Minor.

# UI Changes

| | Before | After |
| --- | --- | --- |
| Default size, aka. `large` | Screenshot 2025-05-20 at 13 05 04 | Screenshot 2025-05-20 at 13 01 17 |
| `.accessibilityExtraExtraExtraLarge` | Screenshot 2025-05-20 at 13 05 27 | Screenshot 2025-05-20 at 13 00 53 |


osanoj (approved):
Nice!

PR #1409, opened 2025-05-20
FavoriteAdsList: Improve dynamic content support

bstien:
# Why?
There were a few simple steps missing in order to improve support for dynamic font sizes in our `FavoriteAdsListView`.

# What?
- Replace usages of `UILabel` with `Label`.
- Fix some constraints for sorting view.
- Limit the searchBar to `.extraExtraLarge`.
	- Letting it grow "infinitely" doesn't look good, and Apple seems to limit it in their apps.

# Version Change
Patch.

# UI Changes
As seen in the screencap below the table view header doesn't calculate the _exact_ size it needs when changing the content size category when viewing this view, but it calculates it correct when you enter the view. Fighting against table view headers is not something anyone enjoys, so I will strongly argue that this a good enough start for now😄

| Before | After |
| --- | --- |
| Simulator Screen Recording - iPhone 16 Pro - 2025-05-20 at 16 20 50 | Simulator Screen Recording - iPhone 16 Pro - 2025-05-20 at 16 15 25 |


AndreyMomot (approved):
🚀

bstien:
I believe "all" snapshots are failing because I've updated Warp for the workspace. I'll move the Warp update to another PR so it's easier to spot the differences in snapshots.

PR #1413, opened 2025-05-22
Button: Improve dynamic content support

bstien:
# Why?
`Button`s would not recalculate its size properly on changes to dynamic fonts, thanks to a few things:
- We calculate size when we give it a title for a state – not the state that's currently being presented.
- We don't recalculate size when state changes.
- We don't recalculate size when `preferredContentSizeCategory` changes.
- We use `UIFont` directly, instead of `Warp.Typography`.

# What?
- Bump DemoKit to [`0.4.5`](https://github.com/finn-no/DemoKit-iOS/releases/tag/0.4.5).
- Use `Warp.Typography` for the style, instead of `UIFont`.
- Recalculate `intrinsicContentSize` when state, title or content size category changes.

# Version Change
Patch.

# UI Changes
None directly. However, thanks to the inconsistent size calculation we've had there could be some smaller changes.


bstien:
The retaken snapshot images are a product of calling `layoutIfNeeded` after each tweak in the DemoKit PR linked above.

PR #1414, opened 2025-05-22
Fix/settings view dynamic font

AndreyMomot:
# Why?

No Dynamic Font support for SettingsView

# What?

Add support for Dynamic Fonts for SettingsView

# Version Change

Patch

# UI Changes

| Before | After |
| --- | --- |
| notif_old| notif_new |
|sett_old|privacy_new|


AndreyMomot:
@bstien thanks for valuable feedback. All updated

PR #1416, opened 2025-06-13
Fix compile error on Xcode 26

nssina:
# Why?

There was a compile error on Xcode 26. This PR aims to fix that.

# What?

In one file we where importing `SwiftUICore` which we needed to replace it with `SwiftUI`

# Version Change

Patch

# UI Changes

N/A


ag12 (approved):
GO!

robinsalehjan (approved):
🥇

PR #1458, opened 2026-08-07
✨ FavoriteAdsList + Folders: full iOS 26 redesign (Warp swipe / Alert / Badge / list hooks)

teddyfahi:
# Version

  Minor bump **157.2.4 → 157.3.0**.
  
  # What changes

  Backs the iOS 26 Liquid Glass Favorites redesign in NMP iOS (host PR: [finn-no/ios-app#9930](https://github.com/finn-no/ios-app/pull/9930) pattern).

  **`FavoriteAdView` cell**
  - Note renders a slim custom element (Warp FileText icon + Caption text on surfaceSunken background), aligned with the ad's title text column.
- commentHostingController sets sizingOptions = .intrinsicContentSize (iOS 16+) so the note wraps correctly on cell reuse.
  - Status pill now renders `Warp.Badge` instead of `RibbonView`. `RibbonViewModel` API preserved — mapped to `Warp.BadgeVariant` internally.
  - FavoriteAdViewModel.commentAlertTitle is now the note's accessibility prefix ("<title>: <text>"), not a rendered heading.

  **`FavoriteFoldersListView`**
 - Trailing swipe: [delete, rename, share] (visual left→right: Share · Rename · Delete) with circular Warp-token discs (bin/edit/share) via UIImage.warpSwipeActionDisc. Default folders (isDefault=true) expose Share only. All actions use .normal style; full swipe never fires destructive.
- editingStyleForRowAt returns .none — trailing-swipe config is the single source of truth for swipes.
- canEditRowAt returns true for every folder row outside edit mode so default folders still receive their share-only swipe.
- estimatedRowHeight set to 80pt to accommodate the 40pt swipe disc.
  - New delegate methods `didShareItemAtIndex:sender:` / `didRenameItemAtIndex:` (default no-op).
  - View model gains optional `renameActionTitle` / `shareActionTitle` / `deleteActionTitle` — omit any to hide that swipe button.
  - New `updateSearchEmptyText(for:)` for hosts that drive search from a docked bar.
  - Public `isSearchBarHidden` / `isAddButtonHidden` for hosts that provide their own search + add-folder controls.

  **`FavoriteAdsListView`**
  - Trailing swipe: [delete, share, comment] (visual left→right: Add note · Share · Delete) with circular Warp-token discs via UIImage.warpSwipeActionDisc.
  - New share action + `didSelectShareItemAt:sender:` delegate (default no-op) + `shareAdActionTitle` on the view model.
  - Public hooks for the iOS 26 host layout: `isSearchBarPermanentlyHidden`, `isSortingViewPermanentlyHidden`, `isMoreButtonPermanentlyHidden`, `pinnedSectionZeroHeaderView`.
  - Warp-styled header, subtitle, section headers, sorting-detail label; empty-state uses Warp tokens; fixes stacked empty view when filtering yields zero rows.
  
  **`FavoriteAdsListEmptyView`**
- Content pins to centerYAnchor so the empty state sits centered vertically.

### `UIImage.warpSwipeActionDisc`
New helper: composes a 40pt filled disc + centred `iconInverted`-tinted icon,
scaled with Dynamic Type (clamped at 80pt). Consumers pair with
`action.backgroundColor = .clear` on iOS 26 (`Warp.UIToken.background` on
older) so the surrounding chip blends into the row.

Duplicates `MyAdsListViewController+SwipeActions.makeCircularSwipeIcon` from
finn/ios-app PR #10376 — planned to move it into warp.

  # Testing
  
Snapshot baselines re-recorded for FavoriteAdCellDemoView (iPad light + dark). Verified in the FINN NMP app on iPhone 17 Pro / iOS 26.5.


teddyfahi:
Converted to draft due to additional changes we want in.

AndreyMomot:
Imho snapshotTests should not be re-triggered here...

FinniversKit/Sources/Fullscreen/FavoriteAdsList/FavoriteAdsListView.swift
```diff
@@ -385,6 +448,13 @@ extension FavoriteAdsListView: UITableViewDelegate {
         return headerView
     }
 
+    public func tableView(_ tableView: UITableView, heightForHeaderInSection section: Int) -> CGFloat {
+        if section == 0, pinnedSectionZeroHeaderView != nil {
+            return UITableView.automaticDimension
+        }
+        return UITableView.automaticDimension
```
AndreyMomot: `heightForHeaderInSection` returns `UITableView.automaticDimension` in both branches, so the `if section == 0` guard is dead code. Either the two cases were meant to return different heights, or this method can be removed - the table view's default is already `automaticDimension`.

AndreyMomot:
`FavoriteAdCommentInputDemoView` snapshots changed (all 4 variants: dark/light × iPad/iPhone) but nothing in this diff touches `FavoriteAdCommentInputView` or its demo. The description only mentions re-recording `FavoriteAdCellDemoView`. These look like they were swept in during a full snapshot re-record run on a different simulator. Worth reverting these 4 PNGs and re-recording only what this PR actually changed.

AndreyMomot:
The `dark_iPhone_editmodenoneselected` snapshot (and some others) is broken - cells render completely blank. Only the section headers ("August", "Juli") are visible; no ad content, no checkboxes, no images.

The root cause is the `UIHostingController` embedding in `FavoriteAdView`. Both `statusBadgeHostingController.view` and `commentHostingController.view` are added as subviews without `addChild`/`didMove(toParent:)`. In the snapshot test environment (headless, no real VC hierarchy), SwiftUI-hosted views do not layout or render when the hosting controller has no parent VC. This likely also breaks the AutoLayout of the surrounding UIKit subviews, producing a fully blank cell.

Options to fix:
1. Route a parent VC into `FavoriteAdView` so it can call `addChild`/`didMove` on both hosting controllers (cleanest but requires cell/view API changes).
2. Use `UIHostingConfiguration` (iOS 16+) for the badge and note if the cell itself is a `UITableViewCell` subclass.
3. Replace with native UIKit Warp equivalents - `Warp.BadgeView` if it exists, or a simple label-based badge - to avoid UIHostingController entirely for these small components.

FinniversKit/Sources/Extensions/UIKit/UIImageExtensions.swift
```diff
@@ -3,9 +3,32 @@
...
 
 extension UIImage {
     convenience init?(frameworkImageNamed name: String) {
         self.init(named: name, in: .finniversKit, compatibleWith: nil)
     }
+
+    static func warpSwipeActionDisc(icon: UIImage, fill: UIColor) -> UIImage {
+        let baseDiameter: CGFloat = 40
+        let baseIconSize: CGFloat = 20
+        let maxDiameter: CGFloat = 80
```
AndreyMomot: `maxDiameter = 80` is hardcoded to match `FavoriteFoldersListView.estimatedRowHeight = 80`, but that row height is set as a fixed value (`tableView.rowHeight = estimatedRowHeight`), not `automaticDimension`. At large accessibility sizes the disc hits 80pt while the row is locked at 80pt — zero vertical padding, and cell text clips because the row can't grow with Dynamic Type. The row height should be `automaticDimension` with 80 as the estimated height only, and `maxDiameter` should either be removed or derived from the actual row height at runtime rather than a magic constant.

teddyfahi:
> Imho snapshotTests should not be re-triggered here...

Agree. I’ll restore the unrelated ones.

FinniversKit/Sources/Fullscreen/FavoriteAdsList/FavoriteAdsListView.swift
teddyfahi: Removed it and kept the default automatic header sizing.

teddyfahi:
> `FavoriteAdCommentInputDemoView` snapshots changed (all 4 variants: dark/light × iPad/iPhone) but nothing in this diff touches `FavoriteAdCommentInputView` or its demo. The description only mentions re-recording `FavoriteAdCellDemoView`. These look like they were swept in during a full snapshot re-record run on a different simulator. Worth reverting these 4 PNGs and re-recording only what this PR actually changed.

Hm, but these snapshots include the updated sold badge, so I would think they are needed.

AndreyMomot:
> Hm, but these snapshots include the updated sold badge, so I would think they are needed.

This one was an AI catched, not reviewed by me manually. So might be wrong 👍

teddyfahi:
> The `dark_iPhone_editmodenoneselected` snapshot (and some others) is broken - cells render completely blank. Only the section headers ("August", "Juli") are visible; no ad content, no checkboxes, no images.
> 
> The root cause is the `UIHostingController` embedding in `FavoriteAdView`. Both `statusBadgeHostingController.view` and `commentHostingController.view` are added as subviews without `addChild`/`didMove(toParent:)`. In the snapshot test environment (headless, no real VC hierarchy), SwiftUI-hosted views do not layout or render when the hosting controller has no parent VC. This likely also breaks the AutoLayout of the surrounding UIKit subviews, producing a fully blank cell.
> 
> Options to fix:
> 
> 1. Route a parent VC into `FavoriteAdView` so it can call `addChild`/`didMove` on both hosting controllers (cleanest but requires cell/view API changes).
> 2. Use `UIHostingConfiguration` (iOS 16+) for the badge and note if the cell itself is a `UITableViewCell` subclass.
> 3. Replace with native UIKit Warp equivalents - `Warp.BadgeView` if it exists, or a simple label-based badge - to avoid UIHostingController entirely for these small components

Good catch. I've replaced the uncontained UIHostingControllers with UIHostingConfiguration in favorite ad cells.

FinniversKit/Sources/Extensions/UIKit/UIImageExtensions.swift
teddyfahi: Yup. Rows now use 80 only as the estimate, and swipe-action circles are capped at 48pt because they where cropped top and bottom when I enlarged the text size.

FinniversKit/Sources/Fullscreen/FavoriteAdsList/FavoriteAdsListViewModel.swift
```diff
@@ -7,6 +7,7 @@ public struct FavoriteAdsListViewModel {
     public let headerShareButtonTitle: String
     public let addCommentActionTitle: String
     public let editCommentActionTitle: String
+    public let shareAdActionTitle: String
```
AndreyMomot, line 10: What is this?

AndreyMomot (approved):

FinniversKit/Sources/Recycling/ListViews/FavoriteFolders/FavoriteFoldersListView.swift
```diff
@@ -322,7 +364,7 @@ extension FavoriteFoldersListView: UITableViewDataSource {
 
         switch section {
         case .addButton:
-            return isSearchActive || tableView.isEditing ? 0 : 1
+            return isSearchActive || tableView.isEditing || isAddButtonHidden ? 0 : 1
```
AndreyMomot, line 396: When `isAddButtonHidden` is true, section 0 already has zero rows, but `setEditing` still deletes or inserts `[0, 0]` for the one-folder/zero-folder transitions. This produces inconsistent table updates and can crash. Could the batch-update condition account for `isAddButtonHidden`, or use `reloadData()` while the add row is permanently hidden?

FinniversKit/Sources/Fullscreen/FavoriteAdsList/FavoriteAdsListViewModel.swift
teddyfahi: we have a new “Share ad” swipe action, so it's the title

AndreyMomot: Gotcha 👍

FinniversKit/Sources/Recycling/ListViews/FavoriteFolders/FavoriteFoldersListView.swift
teddyfahi: 👍 fixed by skipping batch updates while the add button is hidden.

